PR #2, opened 2015-02-25
Simplify API

Ryman:
This introduces overhead in non-optimised builds but should optimise out in `-O` builds. 

It's preferable over just specifying strings by `b"foo"` as that syntax doesn't allow for non-ASCII.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+target/*
+Cargo.lock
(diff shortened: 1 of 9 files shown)


Geal:
Sorry, I had not seen your PR before replacing "".as_bytes() by b"" everywhere, but I think the rest of the functionality is useful. Could you separate the AsBytes trait implementations in a new PR?

Ryman:
@Geal Rebased onto your changes.

Geal:
Thank you for your help!

PR #6, opened 2015-02-26
Missed the let

thehydroimpulse:
/cc @Geal


Geal:
Oh, that's right. Can you make it mutable as well?

Geal:
Fixed in 38b1d2fdae31f9b4b81236abf90aba542b02f35a, please confirm.

thehydroimpulse:
@Geal Awesome, thanks

PR #22, opened 2015-03-31
fix unhelpful warnings

jag426:


Geal:
Thank you for your help! Those warnings were annoying indeed.

PR #24, opened 2015-04-04
Remove core feature dependency

skade:
Remove dependency from core by not using Repr,
but direct pointer access instead.


Geal:
Thanks!

PR #25, opened 2015-04-04
Replace usage of push_all and resize with extend

skade:
This allows removal of the collections feature gate.

Builds and tests on beta-1.0.0.

Please note that extend is slower then `push_all` currently: https://github.com/rust-lang/rust/issues/24028#issuecomment-89455384, but the error message says that extend will probably be the future way to go.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -34,6 +34,7 @@ use std::fs::File;
 use std::path::Path;
 use std::io;
 use std::io::{Read,Seek,SeekFrom};
+use std::iter::repeat;
 
 /// Holds the data producer's current state
 ///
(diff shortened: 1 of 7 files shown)


src/producer.rs
```diff
@@ -78,9 +79,10 @@ impl FileProducer {
 
 impl Producer for FileProducer {
   fn produce(&mut self) -> ProducerState<&[u8]> {
+    let len = self.v.len();
     //let mut v = Vec::with_capacity(self.size);
     //self.v.clear();
-    self.v.resize(self.size, 0);
+    self.v.extend(repeat(0).take(self.size - len));
```
skade: I'm making the assumption here that `resize` only every grows the vector. I might be mistaken there - I only glanced at the surroundings.

Geal: yes, it can only grow the vector

Geal:
I think it can work, but it may be a good idea to wait a bit for extend to get faster, or push_all to be reimplemented, what do you think?

skade:
Hm, I tried finding any substance to that claim of extend/iter being slower, and I didn't find any.

I ran your mp4 parser on bigbuck.mp4 and the beta version with this implementation is in the same range, even a bit faster in debug mode (??). Take it with a grain of salt, as I hacked around nom_benchmark a little, but given that `extend` is stable and ready now, I'd prefer that in this case.

Geal:
I think now it is a better idea to support beta. The performance hit will only affect producers and consumers, and there are other ways to optimize them.

Thank you for fixing this part of the code :)

PR #28, opened 2015-04-13
Update docs and remove debug message

filipegoncalves:
This patch serves two purposes:

* Remove a debugging `println!()` that was lying around in `multispace`/
* Update the docs in the `README` file. The section describing the `take_*` macros was outdated. Also, the old `fold` macros were still mentioned.


Geal:
Great! Thank you, keeping the documentation up to date is a never ending battle :)

PR #31, opened 2015-05-05
Added EOF parser.

filipegoncalves:
This patch adds a parser for EOF. It is useful if we want to build a parser that forces all of the input to be consumed.


Geal:
Thanks!

PR #43, opened 2015-06-01
make count! return a fixed-size array. fixes #42

### LINKED ISSUE #42: Make count! return a fixed-size array

Keruspe:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased][unreleased]
 
 ### Changed
+- `count!` now returns a fixed-size array
 
 ## 0.3.2 - 2015-05-31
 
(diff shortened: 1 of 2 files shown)


Keruspe:
I'm now 100% happy with $i being used for initialization but the rust compiler is quite annoying for this one.

We're sure to either rewrite the whole array or not to return it and it doesn't accepts us to initialize it with something with the same lifetime.

Geal:
What happens with types other than u8?
Example: using a child parser that returns an enum

What about types with a size unknown at compile time?

Keruspe:
For the types different from u8, it works just fine, see https://github.com/Keruspe/tar-parser.rs/blob/72a9d073785a773c9706142c7480ff15e8d055b3/src/parser.rs#L185

Do you have an example of a type that doesn't have a known size at compile time?

PR #44, opened 2015-06-07
chain: allow declaring mutable fields 

Keruspe:


Geal:
Thanks!

PR #47, opened 2015-06-09
Fix infinite macro recursion in the cond! macro

andrew-d:


Geal:
Thank you! I had not seen that one

PR #48, opened 2015-06-09
Don't use "pub mod" and reexports at the same time.

cmr:
This makes rustdoc inline everything that is re-exported into the crate root,
making it easier to see what can be used.


Geal:
Great, thanks! I'll update the online docs

Issue #14, opened 2015-02-26
Example parsers


Geal:
We currently have a few example parsers. In order to test the project and make it useful, other formats should be implemented. Here is a list, if anyone wants to try it:

- **text file formats**:
  - [x] INI
  - [ ] YAML
  - [ ] TOML
- **video and image formats**:
  - [x] MP4 (partial implementation)
  - [ ] OGG
  - [ ] MPEG TS
  - [ ] AVI
  - [ ] PNG
  - [ ] JPEG
  - [ ] GIF
- **executable formats**:
  - [ ] Portable executables (PE)
  - [ ] ELF
- **document formats**:
  - [ ] PDF
  - [ ] MS-CFB (compound format, used in doc, xls, ppt, cab, msi files)
  - [ ] TAR
  - [ ] GZ
  - [ ] ZIP
  - [ ] RAR
  - [ ] binary PLIST
- **network packet formats**:
  - [ ] IP
  - [ ] TCP
  - [ ] UDP
  - [ ] NTP
  - [ ] DNS
  - [ ] PCAP
- **crypto related**:
  - [ ] X.509 certificates
  - [ ] DER public and private keys
  - [ ] SSL/TLS packets
  - [ ] OpenPGP

thehydroimpulse:
I'm writing a Thrift library for Rust that'll use Nom for both their IDL and the network protocol, so that can be another example (although in a different repo).

Geal:
Nice idea, that will be useful! Please notify me when it is done, I will add a link in this list.

filipegoncalves:
This looks interesting. Is anyone actively working on any of these parsers? I'd like to work on a few of these.

Geal:
I have some code for a GIF one at https://github.com/Geal/gif.rs but it is hard to test, since the graphical tools in Piston change a lot.

You can pick any of them. Network packets may be the easiest, since they don't require a decompression phase.

I am using the gif example to see what kind of API can be built over nom. Most of the parsing example are done as one pass over the data, but often there is some logic on the side, and it is not easy to encode correctly.

elij:
I've started a fastq (http://en.wikipedia.org/wiki/FASTQ_format) parser https://github.com/elij/fastq.rs

Geal:
@elij this is a great idea! Was it easy to do?

elij:
[...]

PR #73, opened 2015-08-24
Implement seeking to the end of the content in a MemProducer

ngrewe:
Hi!

I'm currently building a parser for the old Apple/OpenStep binary property list format and that needs to seek to the end of the archive to find some bookkeeping information. To make it easier for testing, I've implemented support for SeekFrom::End in the MemProducer. 

Please have a look whether this makes sense to you. Thanks!

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -391,6 +391,7 @@ mod tests {
   use super::*;
   use internal::{Needed,IResult};
   use std::fmt::Debug;
+  use std::io::SeekFrom;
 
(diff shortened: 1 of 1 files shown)


Geal:
Looks good to me. Could you add unit tests? With one testing the underflow. To make sure I won't introduce regressions :)

ngrewe:
No problem :). I've added tests for seeking backwards within the buffer and beyond start and end

Geal:
great, thanks! By the way, may I ask which format are you parsing with nom? Please notify me on https://github.com/Geal/nom/issues/14 when it is done :)

PR #77, opened 2015-08-26
Allow consumers to seek after reaching eof

ngrewe:
Hi,

another suggestion for improvements with regards to seeking from the end of the input: Presently, once a consumer has reached the end of the input, the consumer's run() method will unconditionally end the parsing. But it is reasonable that the consumer wants to read some information from the end of the input (for example an offset able) and then seek backwards to parse some other portion of the input. 

This PR enables that usage: It ignores/resets the eof flag in run() if the consume() method returned ConsumerState::Seek, so that you can actually jump back from the very end of the input.

Cheers,

Niels

### DIFF
--- a/src/consumer.rs
+++ b/src/consumer.rs
@@ -128,6 +128,7 @@ pub trait Consumer {
           let _ = producer.seek(seek_from);
           //println!("seeking: {:?}", pos);
           should_seek = false;
+          eof = false;
           acc.clear();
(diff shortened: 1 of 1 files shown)


Geal:
Ok, so from what I understand, it keeps the current behaviour, but if you get to the end and return `Seek` it does not stop. That's a cool idea!

PR #85, opened 2015-09-07
Use `while let` in place of `loop` and `if let`.

daboross:
`while let` simplifies this greatly, and removes the need for an explicit ` else { break; } }` at the end of the loop statements.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1414,16 +1414,12 @@ macro_rules! many0(
     {
       let mut res   = Vec::new();
       let mut input = $i;
-      loop {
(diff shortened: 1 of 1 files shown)


Geal:
Thank you, I like it when code gets easier to read :)

daboross:
I found this while using clippy on a program which uses nom macros. Would other changes from running `clippy` on files in `tests/` be worthwhile?

Geal:
Probably. I used clippy for the 0.4.0 release, and it show a few lifetime elisions and issues with iterator usage. Don't hesitate to report if you see other cases.

PR #87, opened 2015-09-11
Clippy changes

daboross:
These are changes recommended by clippy, found using the `cargo-clippy` tool on test files, as follows:
```
cargo clippy --test arithmetic
cargo clippy --test ini
cargo clippy --test issues
cargo clippy --test mp4
cargo clippy --test omnom
cargo clippy --test test1
```

This produces more changes to merge then just running clippy on the library, because the macros are used in the tests.

Each commit has a separate type of change. I don't really know the style of this repository that well, so feel free to use any individual commit without the other ones in this PR.

### DIFF
--- a/src/character.rs
+++ b/src/character.rs
@@ -8,7 +8,7 @@ use util::ErrorCode;
 macro_rules! one_of (
   ($i:expr, $inp: expr) => (
     {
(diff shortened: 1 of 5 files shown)


Geal:
Looks good, thanks!

PR #89, opened 2015-09-17
Add a bare-function form for dbg and dbg_dump

ahenry:
Not much to say about this, really.  I followed a pattern I saw elsewhere in the code to deal with this case.


Geal:
Thank you!

PR #91, opened 2015-09-27
Add doc and comment for count_fixed! output being Copy

bluss:


bluss:
Not sure if this is needed. Just looking at all `unsafe` blocks. The output type being `Copy` guards this code from having panic / exception safety issues (destructors would run on uninit data at panic otherwise).

bluss:
(Since submac is user supplied, we can insert a panic while the uninit array is alive.)

Geal:
I see, nice catch! If you have other ideas like this one, I'm all ears :)

bluss:
I'm just auditing the `unsafe` code in the crate, hope you don't mind. The other important issue is the `print_offsets` (and similar functions) one.

To clarify this issue, The `[T; N]` expression is what imposes the `T: Copy` bound. Which is fortunate, without that bound, the code wouldn't be panic safe. Putting in the comment is good to ensure this isn't lost in refactoring.

bluss:
I think you'll be in a great position with these fixed. Not only offering safe rust-based parsing with very high performance, but doing so using almost no `unsafe`-marked code.

Issue #83, opened 2015-09-07
Rename `filter!`


Geal:
filter is really ill defined right now

PR #93, opened 2015-09-30
Remove `filter!`

sourrust:
As mented in issue #83, there was a need to rename the `filter!` macro. So I look into it and found that both [`filter!`][filter] and [`take_while!`][take_while] are one in the same. Removing `filter!` out right seems like the logical thing to do and just use `take_while!` instead.

[filter]: https://github.com/Geal/nom/blob/master/src/bytes.rs#L149
[take_while]: https://github.com/Geal/nom/blob/master/src/bytes.rs#L173


Geal:
Great, thank you!

PR #97, opened 2015-10-07
Add inline attributes to various helpers

meh:
As discussed on IRC many moons ago.


Geal:
Thanks! To be honest, I had forgotten about that idea :/

PR #98, opened 2015-10-07
qualify call to std::str::from_utf8

ccmtaylor:
the call needs to be qualified in the macro expansion, or else callers will get errors if they don't have `use std::str::from_utf8` in their code.


Geal:
Thank you!

PR #103, opened 2015-10-13
Use unboxed closure in MapConsumer

soro:
Don't have a lot of experience with Rust and trying out a few things. Haven't checked if this really generates better code, but it should.
I like the design a lot. It has very low overhead and looks like it would just optimize into nice loops - most generic iteratee style streaming libraries would probably have to compromise.
I think the T parameter should also have a Clone bound. I tried to implement map and bind for ConsumerState and that ended up being required. Unfortunately I wasn't able to pacify the borrow checker, despite it being perfectly safe code. Would make the code a lot shorter.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -336,15 +336,17 @@ impl<'x> Producer<'x,&'x [u8],Move> for FileProducer {
 use std::marker::PhantomData;
 
 /// MapConsumer takes a function S -> T and applies it on a consumer producing values of type S
-pub struct MapConsumer<'a, C:'a ,R,S,T,E,M> {
+pub struct MapConsumer<'a, C:'a,R,S,T,E,M,F> {
   state:    ConsumerState<T,E,M>,
   consumer: &'a mut C,
-  f:        Box<Fn(S) -> T>,
-  input_type: PhantomData<R>
+  f:        F,
+  consumer_input_type: PhantomData<R>,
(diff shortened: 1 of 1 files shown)


Geal:
That's interesting! Don't feel bad about the borrow checker annoying you here. I had to fight a lot too to get the MapConsumer working.

Look on the bright side: Travis says it does not build on stable, but it passes on nightly and beta. So in less than 6 weeks, I'll be able to merge that PR :)

soro:
Oh, I guess you changed something this morning since it was a clean merge to the 1.0 branch last night. No worries, just wanted to see if it was possible to make the closure unboxed and suggest the change.
It is unfortunate that some things can be written using a match but not a higher order function, despite basically having the exact same semantics, especially after inlining.

soro:
Oh and yes, I pretty much just use the unstable compiler since there are so many features that are still gated that are kind of important. I feel that the 1.0 tag for rustc was a bit premature.

Geal:
Yes, I changed a few things, but nothing too serious, it will be easy to update your PR

Geal:
Hey, since Rust 1.4 was released yesterday, that pull request passes the test. Could you update to the latest commits of 1.0? I'll see if this could go into 1.1

soro:
Rebased on 1.0

Geal:
Merged in master, thanks!

PR #106, opened 2015-10-20
Added length_bytes! macro

mrordinaire:
Added `length_bytes!` which gets a number from the first parser, then extracts that many bytes from the remaining stream.


Geal:
Thank you!

PR #111, opened 2015-10-28
Allow expressions in field bindings in `chain!`

abbradar:
`ident` is very restricting in sense that it doesn't allow e.g. qualified names:

```
foo: module::submodule::my_parser
```

Unfortunately, we can't just use `expr` because of '~' operator nearby. We can workaround this by allowing `expr` inside brackets instead, so with this patch the following is legal:

```
foo: (module::submodule::my_parser)
```

This is actually my second day of getting acquainted with Rust, so please bear with me if the solution is unidiomatic or just plain invalid -- I'll be happy to fix. Thanks!

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -623,9 +623,14 @@ macro_rules! chain (
 #[doc(hidden)]
 #[macro_export]
 macro_rules! chaining_parser (
+  ($i:expr, $consumed:expr, ( $e:expr ) ~ $($rest:tt)*) => (
+    chaining_parser!($i, $consumed, call!($e) ~ $($rest)*);
+  );
+
   ($i:expr, $consumed:expr, $e:ident ~ $($rest:tt)*) => (
     chaining_parser!($i, $consumed, call!($e) ~ $($rest)*);
   );
+
   ($i:expr, $consumed:expr, $submac:ident!( $($args:tt)* ) ~ $($rest:tt)*) => (
     {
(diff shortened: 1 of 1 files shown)


abbradar:
One issue left is this
```
name: (module::parser),
```
doesn't work, whereas that
```
name: (module::parser) ~
```
does. Any ideas?

Geal:
going from `ident` to `expr` by adding parenthesis is a big syntax change, I'm not sure that's a good idea. Maybe a `path` could work? Does `path` in macro arguments allow for local variables?

abbradar:
```
/home/abbradar/cgate/deps/nom/src/macros.rs:537:48: 537:55 error: `$e:path` is followed by `~`, which is not allowed for `path` fragments
/home/abbradar/cgate/deps/nom/src/macros.rs:537   ($i:expr, $consumed:expr, mut $field:ident : $e:path ~ $($rest:tt)*) => (
```

abbradar:
Ping. What do you think? I use this successfully in several places already, but there's a general question of allowing `expr`s and also an issue left (see above).

joelself:
Why can't you just put `use module::submodule::my_parser;` in your file and then just call `my_parser`
 directly?

abbradar:
Well, I think it was for clarity but I don't use this code anymore. If this is considered not a good idea, feel free to close it.

Geal:
chain is not used anymore, closing this now

PR #112, opened 2015-11-01
Fix `FileProducer` passing an empty slice as input

sourrust:
I found a bug where `FileProducer` will always return `Input::Eof(Some(_))` even when the slice is empty.

Also, just to let you know, the travis test is reporting the build as passing even when the test `stream::tests::seeking_consumer` is failing because it can't find a file named `testfile.txt`.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -322,7 +322,15 @@ impl<'x> Producer<'x,&'x [u8],Move> for FileProducer {
       //println!("producer state: {:?}", self.state);
       match self.state {
         FileProducerState::Normal => consumer.handle(Input::Element(&self.v[self.start..self.end])),
-        FileProducerState::Eof    => consumer.handle(Input::Eof(Some(&self.v[self.start..self.end]))),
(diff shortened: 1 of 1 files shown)


Geal:
I see, thanks! SO now, your flac parser uses the 1.0 beta, right? What issues did you encounter? Your repository triggered a few fixes in the beta, around error types, and a refactoring of bit level parsing.

What causes the hangs you refer to in your commit messages? Adding a function to grow the producer's buffer should be easy enough.

sourrust:
Haven't encountered that many, but the ones I've noticed are from Producers. Beside the one that got fix with this merge, the buffer size not growing in `FileProducer` is the other. And that is what is causing things to hang, functions using a `FileProducer` inside of it end up being stuck in a loop of the handle waiting for more than the current size of the buffer and the slice being passed is always to same size.

Geal:
It makes sense. If you need anything else in the producers, please tell me. It looks like I won't release the 1.0 until next week, and I'd like to fix as many issues as possible in the meantime.

PR #114, opened 2015-11-04
Fixing typo.

breard-r:
Title is self-explanatory.

PR #116, opened 2015-11-07
More typo fixes

nelsonjchen:
Supercedes #114 .


Geal:
Thanks! I should really check more what I write. Unfortunately, rustc does not check syntax for english yet

PR #118, opened 2015-11-08
Proof of concept: change default error bound to Box<std::error::Error>

danburkert:
OK here is a proof of concept of what it would take to change the default error bound from `u32` to `Box<std::error::Error>`.  The crate compiles, but the tests are failing to compile because most of them implicitly depend on the default `Err` type being `Eq`, which it no longer is.  You can see a few places in `util` where I had to make explicit some of those trait bounds.

The biggest issue I see with this is that `std::error::Error` is not part of core.  However, neither is `Box`, and that is already being used as part of `Err`, so I'm not really sure core is supported anyway.

I'm happy to fix up the tests if you think this is worth pursuing.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -6,6 +6,7 @@ use util::ErrorKind;
 #[cfg(feature = "core")]
 use std::prelude::v1::*;
 use std::boxed::Box;
+use std::error;
 
 /*
 /// (Experimental) Closure used to hold the temporary state of resumable parsing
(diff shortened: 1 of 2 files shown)


Geal:
Things like `Box` can be added on demand with the `collections` feature, so that's not much of an issue (but good point, I have to update the example `no_std` project).

I am still uneasy about that change. The default type was there to avoid breaking compatibility for code that did not use custom error codes anyway, so `()` instead of `u32` would have been fine. I am not sure imposing the `std::error::Error` trait improves that state.

The approach of errors in nom is not in looking at a specific error code, because it misses a lot of context. Instead, the idea is to match on an error chain, which gives the most precise information on which part of the input reached which part of the parser. This also enables advanced patterns like pre-generating error chains on known bad inputs, to keep up with parser refactorings.

Geal:
maybe that's not necessary anymore with the recent changes to error management

PR #120, opened 2015-11-11
Fix FileProducer::resize

sourrust:
Was having trouble with `resize` on the new beta and found that the producer was sending over the wrong part of the slice. So this pull request fixes that and I also refactored `resize`.


Geal:
Looks good to me, thanks!

PR #121, opened 2015-11-16
Use Write::write in FileProducer::resize

bluss:
Use Write::write in FileProducer::resize

This way we don't need any `unsafe` marked code here.

Use `vec![0; s]` to zero fill efficiently, and `Write` methods to do the
copying.


bluss:
This is just a simple PR to try to use safe code where possible. I don't know a way to do it for the shift case further down the file (copy of possibly overlapping slices).

bluss:
`Vec::write` appends while `<&mut [u8]>::write` overwrites. Both are useful, but I went for the latter to use the efficient zero initialize using `vec![]`.

Geal:
Thanks!

PR #130, opened 2015-11-23
Fix seeking_consumer test to use a file that exists

luser:
This test was failing for me because it was trying to use a nonexistent file.


Geal:
Thanks!

PR #134, opened 2015-11-26
Started work on str module

Binero:
1 function finished


Binero:
I'd argue coverage is pretty good on this. Not sure what caused it to drop.

Issue #144, opened 2015-12-17
Need way to distinguish incomplete input from eof


jsgf:
Something like `chain!(tag!("1")? ~ eof, ...)` will return `Incomplete` if given an empty string, because there's no way to distinguish between giving it a partial input and a complete input which will match `eof`. If the input is represented with `&[u8]` I don't think there is a way to do it without some hack like making a magic marker sequence for EOF. 

I guess one way to do it is have a Eof trait which the input can implement which can indicate whether a short input is just incomplete or EOF, and have the `eof` query that rather than directly looking at the input length. Similarly, other functions would only ever return `Incomplete` if the input hasn't hit EOF, otherwise they return errors.

mbr:
possibly related:

```rust
#[macro_use]
extern crate nom;

use std::str;
use nom::digit;

named!(num_f64<f64>,
    map_res!(map_res!(
        recognize!(
            chain!(digit ~ chain!(tag!(".") ~ digit, ||{})?, ||{})
        ),
    str::from_utf8), str::parse)
);

fn main() {
    let input = b"123";
    println!("Result: {:?}", num_f64(input));
}
```

will output

```
Result: Incomplete(Size(4))
```

Changing the `input` to `b"123 "` (extra space) will cause the parser to match:

```
Result: Done([32], 123)
```

PR #145, opened 2015-12-17
Add AtEof trait to determine when input has finished.

jsgf:
This is an attempt to address issue #144.

With this trait, it becomes possible to distinguish between a partial input
and a complete input. If there's a partial input, then it makes sense to
return Incomplete, but if the input is finished then it's just a plain
Error. The need_more() function encapsulates this logic.

This is still a work in progress; it breaks a couple of unit tests, presumably because they're expecting Incomplete where now they're getting eof.

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -56,7 +56,7 @@ macro_rules! bits_impl (
         $crate::IResult::Incomplete($crate::Needed::Unknown) => $crate::IResult::Incomplete($crate::Needed::Unknown),
         $crate::IResult::Incomplete($crate::Needed::Size(i)) => {
           //println!("bits parser returned Needed::Size({})", i);
-          $crate::IResult::Incomplete($crate::Needed::Size(i / 8 + 1))
+          $crate::need_more($i, $crate::Needed::Size(i / 8 + 1))
         },
         $crate::IResult::Done((i, bit_index), o)             => {
           let byte_index = bit_index / 8 + if bit_index % 8 == 0 { 0 } else { 1 } ;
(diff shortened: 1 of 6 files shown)


Geal:
Hi!

This idea is interesting. Usually, I recommend that people use the `complete!` combinator, to transform a `IResult::Incomplete` in an `IResult::Error`.
The issue here is that the semantics of whether there is more input or not depends on runtime properties, not on the type. When I use buffers to get data from the network or a file, the following cases can happen:

- there is enough data in the buffer to parse
- there is not enough data in the buffer, but more can come
- there is no data in the buffer (because the previous parsing session consumed everything), but more can come
- we're at EOF (no more file data, cloed connection, etc), and there is a bit of data left
- we're at EOF and there's no more data in the buffer

The parser cannot decide if there is more data, and the buffer type cannot hold that information. That's why I encoded that in the stream's [`Input` type](https://github.com/Geal/nom/blob/master/src/stream.rs#L15-L21).

There may be a better solution for this, but I don't see it right now.

jsgf:
Hm, I'd managed to overlook `complete!`. For this specific use case I am expecting the entire input at once, and so I could solve my problems by scattering it around the place in my parser.

But, as you say, hard-coding that kind of behaviour is undesirable. If I wanted to process a streaming input, I would need to have a different variant of the parser which is mostly identical except for the use of `complete!`.

I haven't looked at `Input` much. It looks to me like you need to break the parser up into pieces, and then do your own state management for each piece to incrementally parse the input?

jan-hudec:
@Geal,
> That's why I encoded that in the stream's Input type.

I think the way it is done for streams is OK. What is missing is simple way to say “here is a `&[u8]` (or `&str`) and it is complete”.

Geal:
Hi!

So, I'm really sorry for the delay in merging this, but now it will finally appear, in the next major release, nom 4! :)

Geal:
yup, I'll add this

Geal:
so, implementing on `[u8]` and `str` does not really work, since `need_more` should require the trait on the input type itself, not a reference.

veprbl:
I've opened #657 to discuss if this can be improved or replaced with something else.

PR #151, opened 2015-12-30
Update documentation for regex macros that return a Vec

joelself:
re_matches(_static), re_capture(_static), and re_captures(_static) all return Vectors of their matches/captures. Also updated the names of static macros in the documentation.


Geal:
Thanks!

Issue #152, opened 2015-12-31
Issue with named parsers in switch! macro


zrneely:
Using named parsers in the switch! macro can fail in interesting ways:

     named!(take4, take!(4));
     named!(xyz, tag!("XYZ"));
     named!(abc, tag!("abc"));
     named!(sw,
         switch!(take4,
             b"abcd" => xyz |
             b"efgh" => abc
         )
     );

    <nom macros>:20:25: 20:26 error: no rules expected the token `!`
    <nom macros>:20 { switch ! ( $ i , call ! ( e ) , $ ( $ rest ) * ) } ) ;

Geal:
Hi!

It seems the issue is in the right hand side of pattern matches, with `xyz` and `abc`. Right now, `switch!` only accepts macros there. I'll see what I can do to support normal functions too.

Geal:
I changed a few things in f255e2e, but we still cannot use functions instead on macros. I added a warning in the documentation for this issue: [https://github.com/Geal/nom/blob/master/src/macros.rs#L1136-L1159](https://github.com/Geal/nom/blob/master/src/macros.rs#L1136-L1159).

The documentation will be up once I release nom 1.1.

I'm sorry I could not fix it. There may be a macro trick to do it here, but it will require some thought.

DemiMarie:
Could this be partially solved by providing an identity macro that just expands to its argument?

PR #153, opened 2015-12-31
Partial fix for using named parsers in switch! macro

zrneely:
This addresses part of the issue raised in #152, and adds a test case to ensure the problem is solved.


zrneely:
Note: please don't merge this pull request as is, it doesn't fix the issue (and the tests fail for me, locally, as expected).

PR #155, opened 2016-01-01
Improve type inference of (n)one_of(_bytes).

zentner-kyle:
I was playing with nom earlier today, and found that the following doesn't compile:
```
many1!(one_of_bytes!(b"0123456789ABCDEFabcdef"))
```
I tracked this down to some unusual behavior in the type inferencer, which I don't really understand. Basically, adding the `_` type parameter causes the inferencer to correctly determine the type, so that the above (and similar variants) compile.

You can see an extracted version of the same problem here: http://is.gd/fuSl3z

Since I don't really understand the fix, I'm not sure you should merge this. But I assume a pull request is more useful than a bug report.

### DIFF
--- a/src/character.rs
+++ b/src/character.rs
@@ -9,7 +9,7 @@ macro_rules! one_of (
   ($i:expr, $inp: expr) => (
     {
       if $i.is_empty() {
-        $crate::IResult::Incomplete($crate::Needed::Size(1))
+        $crate::IResult::Incomplete::<_, _>($crate::Needed::Size(1))
       } else {
         #[inline(always)]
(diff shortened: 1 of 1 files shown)


Geal:
Hi!

From what I see, it would work, but I have to test it a bit before merging. If it works, though, the result will be huge, because I can remove some ugly fixes I had to do before. As an example, I could factor [take!](https://github.com/Geal/nom/blob/master/src/bytes.rs#L513-L525) and [take_s!](https://github.com/Geal/nom/blob/master/src/str.rs#L53-L65) in the same macro.

So, your fix could be very useful ;)

Geal:
Thanks!

PR #156, opened 2016-01-04
Added Unicode support to the str module

joelself:
As I was going in to add `take_while_s` and `take_while_and_consume_s` macros to the `str` module, I noticed that `str` character offsets were being used to offset into arrays of bytes (e.g.):

```rust
// position() returns character offsets
match $input.chars().position(|c| !$submac!(c, $($args)*)) {
          Some(0) => $crate::IResult::Error($crate::Err::Position($crate::ErrorKind::TakeWhile1Str,$input)),
          Some(n) => {
            let res = $crate::IResult::Done(&$input[n..], &$input[..n]);
                                         // &str[..n] takes byte offsets
// ...snip
```
The `position` function returns a character offset into the `str`, but `&str[..offset]` and `&str[offset..]` take byte offsets.

Now, for ASCII characters this is not a problem because `sizeof(char) == sizeof(byte)`, but for non-ASCII characters the sizes aren't equal; Unicode characters can be from 1 to 4 bytes. The result is that if you run these macros with non-ASCII strings it returns parsed slices that are smaller than expected.

Example:

`$input = "ÒA🀝z5"`

byte index |` 0 `|` 1 `|` 2 `|` 3 `|` 4 `|` 5 `|` 6 `|` 7 `
--------------|---|---|---|---|---|---|---|---
character  |`Ò`|  |`A`|`🀝`|   |   |`z`|`5` 
**char index**|**`0`**|   |**`1`**|**`2`**|   |   |**`3`**|**`4`**

If you got the character position of `'z'`, 3, and then created the slices `&*$input[..3]` and `&*$input[3..]` you would end up with `"ÒA"` and `"🀝z5"` instead of the intended `"ÒA🀝"` and `"z5"`. Even worse if it is split on the `'5'` the strings wouldn't even end and begin on Unicode character boundaries and the resulting `str` slices would be totally messed up.

The good news is that I fixed all the `str` module macros  to use byte offsets that line up with character boundaries.  I used `str.chars().char_indices()` to get the byte offset of each character while keeping track of the number of characters when needed. Then i used the byte offset to create `str` slices. I had to do this for every `str` macro except `tag_s` which works fine as-is, then I  added the `take_while_s` and `take_while_and_consume_s` macros and added test functions for all macros that currently don't have any. Then I ran my [tomllib](https://github.com/joelself/toml_parser), which uses the `str` module and 'regex_macros' exclusively, against [this new hard TOML file I produced](https://github.com/joelself/toml/blob/master/tests/hard_example_unicode.toml) and it parsed and reproduced the file perfectly, so everything seems to work just fine.

There are two other minor things I changed:
* In macros.rs line 1366 `peek!($i, call!(f));` was missing a `$` before the `f` so I added one.
* `is_not_s` and `is_a_s` loop through the entire `$arr` str for each character in the input looking for a match/non-match. This is an `O(n^2)` algorithm. I put `$arr`'s characters into a `HashSet` so each comparison against `$arr`'s chars would be `O(1)` making the overall algorithm `O(n)`.

One thing about the `use std::collections::HashSet;` statement. I tried putting it everywhere, but no matter where I put it, the compiler would complain that `HashSet` isn't defined. The only way i could get it to work was to put the declaration directly inside the macros themselves which just doesn't seem right...but works. If there's a better way to do this then please change it and please tell me what I need to do to to get this right in the future.

Thanks.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1260,7 +1260,7 @@ macro_rules! peek(
     }
   );
   ($i:expr, $f:expr) => (
-    peek!($i, call!(f));
+    peek!($i, call!($f));
   );
 );
 
--- a/src/str.rs
+++ b/src/str.rs
@@ -54,10 +54,19 @@ macro_rules! take_s (
   ($i:expr, $count:expr) => (
     {
       let cnt = $count as usize;
-      let res: $crate::IResult<&str,&str> = if $i.len() < cnt {
+      let res: $crate::IResult<&str,&str> = if $i.chars().count() < cnt {
         $crate::IResult::Incomplete($crate::Needed::Size(cnt))
       } else {
-        $crate::IResult::Done(&$i[cnt..],&$i[0..cnt])
+        let mut offset = 0;
+        let mut count = 0;
+        for (o, _) in $i.char_indices() {
+          if count == cnt {
+            offset = o;
+            break;
+          }
+          count += 1;
+        }
+        $crate::IResult::Done(&$i[offset..], &$i[..offset])
       };
       res
     }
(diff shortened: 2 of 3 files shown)


Geal:
Hi!

That's good, thanks! I forgot to use `char_indices` in the macro, while I added them for other functions.

Could you break up this PR in smaller ones, rebased on master? The merge commit is a bit unsettling :)

joelself:
I'll do what I can. I made [three separate pull requests to the toml repository](https://github.com/toml-lang/toml/pull/378) and github combined them all into one. I'm going to try merging from three separate branches and see if that works.

joelself:
Ok I broke up the changes into 5 separate pull requests. The only problem I see is that in the macro code you use 2 space indentation and at first you used 4 space tab indentation in the tests, which is the style I copied. However, while I was writing those tests you checked in some new tests that use 2 space indentation. I'll left my tests at 4 space indentation and you can decide what to do with the file indentation wise.

Geal:
ah, yes, the person who created the file put 4 spaces even though the whole project used 2 spaces. I'll fix that

PR #161, opened 2016-01-06
Added take_until_and_consume_s macro,tests cases error enum vals

joelself:
Added take_until_and_consume_s macro tests cases for the macro and error enum vals for the return values of the macro.

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -247,6 +247,7 @@ macro_rules! take_till_s (
 macro_rules! take_until_and_consume_s (
   ($input:expr, $substr:expr) => (
     {
+      use ::std::vec::Vec;
       #[inline(always)]
       fn shift_window_and_cmp(window: & mut Vec<char>, c: char, substr_vec: &Vec<char>) -> bool {
         window.push(c);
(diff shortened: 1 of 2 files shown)


src/str.rs
```diff
@@ -241,6 +241,52 @@ macro_rules! take_till_s (
...
+        window.push(c);
+        if window.len() > substr_vec.len() {
+          window.remove(0);
+        }
+        window == substr_vec
+      }
+      let res: $crate::IResult<&str, &str> = if $substr.len() > $input.len() {
+        $crate::IResult::Incomplete($crate::Needed::Size($substr.len()))
+      } else {
+        let substr_vec: Vec<char> = $substr.chars().collect();
```
Geal: `Vec` should be imported with the absolute path `::std::vec::Vec`, in case someone has their own `Vec` type. It happened for `Option` recently.

joelself: I tried `use ::std::vec::Vec;` and `use ::std::vec::Vec as Vec;` and they both warn of unused imports. Should I just make all references to Vec say `::std::vec::Vec` instead of `Vec`?

joelself:
Updated the pull request with the new change.

Geal:
I think a conflict during compilation is still possible if there's already a `Vec` symbol in the context, so I put absolute paths everywhere in the merge.

Thanks for your help!

PR #162, opened 2016-01-06
Fixed take_till_s to work with non-ASCII characters added a test

joelself:
Fixed take_till_s macro to work with non-ASCII characters using char_indices and added a succeed test.


Geal:
Thanks!

PR #163, opened 2016-01-06
Fixed take_while_s and take_while1_s to work with non-ASCII chars

joelself:
Fixed take_while_s and take_while1_s to work with non-ASCII characters using `char_indices`. Added take_while_1_s succeed and fail tests and take_while_s succeed and succeed_some tests.


Geal:
Thanks!

PR #164, opened 2016-01-06
Fixed is_a_s and is_not_s to work with non-ASCII chars and added tests

joelself:
Fixed is_a_s and is_not_s to work with non-ASCII characters using `char_indices`. Also changed the character comparison to use a HashSet to improve performance from `O(n^2)` to `O(n)`. Added succeed and fail tests for both macros.

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -64,6 +64,7 @@ macro_rules! take_s (
   );
 );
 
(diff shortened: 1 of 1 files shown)


Geal:
Thanks!

PR #167, opened 2016-01-08
Add missing methods on IResult<I,O,E>

GuillaumeGomez:


src/internal.rs
```diff
@@ -46,25 +46,25 @@ pub enum IResult<I,O,E=u32> {
   Incomplete(Needed)
 }
 
-impl<I,O> IResult<I,O> {
+impl<I,O,E> IResult<I,O,E> {
   pub fn is_done(&self) -> bool {
-    match *self {
-      Done(_,_) => true,
-      _         => false
+    match self {
```
Geal: why change that? matching on `*self` is valid. It is even recommended by clippy

src/internal.rs
```diff
@@ -46,25 +46,25 @@ pub enum IResult<I,O,E=u32> {
...
     }
   }
 
   pub fn is_err(&self) -> bool {
-    match *self {
-      Error(_) => true,
-      _        => false
+    match self {
+      &Error(_) => true,
+      _         => false
```
Geal: same comment

src/internal.rs
```diff
@@ -46,25 +46,25 @@ pub enum IResult<I,O,E=u32> {
...
     }
   }
 
   pub fn is_incomplete(&self) -> bool {
-    match *self {
-      Incomplete(_) => true,
-      _             => false
+    match self {
+      &Incomplete(_) => true,
+      _              => false
```
Geal: same comment

src/internal.rs
GuillaumeGomez: Absolutely, mistake I made, my bad.

GuillaumeGomez:
Updated.

Geal:
Thanks!

PR #170, opened 2016-01-15
Add `hex_digit` parser for hexadecimal digits

jansegre:
Implementation of the corresponding functions/methods to `digit`, but for hexadecimal digits named `xdigit` in line with regex character class.

I wonder if that's the best name because regex character classes naming is not a norm, and also if other classes should be implemented.  I needed this parser and thought it might be useful to belong to `nom`.


Geal:
I think `hexdigit`or `hex_digit` would be more explicit. While regular expressions are often concise, they do not have a reputation of readability ;)

jansegre:
Yeah, `xdigit` is not a good idea, there are already parsers like `alphanumeric` which are not called `alnum`, so `hex_digit` makes more sense (`hexadecimal_digit` is too long).

I just saw that coverage decreased for me not adding tests. Should I add a `tests/hex_digit.rs`? All the other tests seem well by themselves. Also there is not unit tests standard, but I'm a little uneasy not adding tests.

Geal:
Yes, please add tests for this, in the test part of `src/nom.rs`

Geal:
could you update `hex_digit` like it was done for other functions in https://github.com/Geal/nom/pull/172/files ?
Returning `&input[0..0]` when all of the input was consumed causes issues with offset calculations

jansegre:
Fix from #172 applied. Tests added. I think it's looking good.

My only concern if it is okay to bump all other variants on `error_to_u32`, it only seems natural, but it's now really backwards compatible, though that conversion probably needn't be.

What do you think?

Geal:
I'd prefer that you add it at the end of the `error_to_u32` method. This is what I usually do when I add a new error variant. Once it is done, I'll merge this. Thanks for putting up with my requests :)

Geal:
Thanks!

PR #172, opened 2016-01-18
Fixes #171 - alpha, digit, ... return proper value on no match

### LINKED ISSUE #171: recognize! has odd behaviour at end of input
There seems to be an odd behaviour, when `recognize!` is used at the end of input:

```rust
#[macro_use]
extern crate nom;

use nom::digit;

named!( integer_literal,
        recognize!(digit)
);

#[cfg(test)]
mod test {

    use super::{integer_literal};
    use nom::{IResult};

    #[test]
    fn literal_int () {
        let int_literal   = &b"123"[..];
        let res = integer_literal(int_literal);
        assert_eq!(res, IResult::Done(&b""[..], &b"123"[..]));
    }

    #[test]
    fn literal_int_add () {
        let int_literal   = &b"123abc"[..];
        let res = integer_literal(int_literal);
        assert_eq!(res, IResult::Done(&b"abc"[..], &b"123"[..]));
    }

}
```

While the first test fails, the second one succeeds. This is due to the fact, that in the second test `integer_literal(int_literal)`  will return `Done([], [])` instead of `Done([], [49, 50, 51])`:

```
> cargo test
   Compiling sql_2 v0.1.0 (file:///var/tmp/tstorch/src/rust/sql_2)
     Running target/debug/sql_2-47240fc3d9a7ec51

running 2 tests
test parser::test::literal_int_add ... ok
test parser::test::literal_int ... FAILED

failures:

---- parser::test::literal_int stdout ----
	thread 'parser::test::literal_int' panicked at 'assertion failed: `(left == right)` (left: `Done([], [])`, right: `Done([], [49, 50, 51])`)', src/parser.rs:17


failures:
    parser::test::literal_int

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured
```

tstorch:
The functions `alpha`, `digit`, `alphanumeric`, `space` and
`multispace` returned `Done(&input[0..0], input)`. This was a
problem in `recognize`, because it uses the pointer offset of
`s` and `t` in the returned `Done(s, t)`. This offset would be
zero before and therefore `recognize` would return `Done([], [])`
as it was `Done(s, &t[..0])`.

Now `alpha`, etc. return `Done(&input[input_length..], input)`
and therefore `recognize` returns `Done(s, &t[..t.len()])`,
which equals `Done([], &t[..])`.

Someone should probably write a test for this.
Also this functions could be combined into a macro, since they are basically the same except one function and one error in each.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -769,6 +769,7 @@ mod tests {
   use internal::IResult::*;
   use internal::Err::*;
   use util::ErrorKind;
+  use nom::{alpha, digit, alphanumeric, space, multispace};
 
   #[test]
   fn is_a() {
@@ -810,7 +811,6 @@ mod tests {
     assert_eq!(a_or_b(f), Done(&b""[..], &b"fghi"[..]));
   }
 
-  use nom::alpha;
   #[test]
   fn escaping() {
     named!(esc, escaped!(call!(alpha), '\\', is_a_bytes!(&b"\"n\\"[..])));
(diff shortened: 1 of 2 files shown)


Geal:
From what I see, this would not break anything, but please add tests for this

tstorch:
Where should I put them, since they involve recognize and digit? (One test is in #171 )

Geal:
The tests from #171 in the same file as `recognize!` and the tests for `digit` and others in `src/nom.rs`

tstorch:
What test should i write for `digit` and the others? An offset test for the return value?

Geal:
yes, please

tstorch:
Done. I checked both tests on current master (5f8af32) and both fail. With the patch both succeed.

Geal:
Thank you for your help!

PR #174, opened 2016-01-20
Sep list revamp

### LINKED ISSUE #166: separated_list should allow the child parser to consume 0 input
The separator must always consume, but empty cases between separators should be allowed

tstorch:
Overhaul of separated_list macro. Now easier and better. Includes test wich would fail in 2aff5a3 with empty elements. Test succeeds with patch.

Fixes #166

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1741,45 +1741,30 @@ macro_rules! delimited2(
 macro_rules! separated_list(
   ($i:expr, $sep:ident!( $($args:tt)* ), $submac:ident!( $($args2:tt)* )) => (
     {
-      let mut res   = ::std::vec::Vec::new();
-      let mut input = $i;
+      let mut res      = ::std::vec::Vec::new();
+      let mut input    = $i;
+      let mut buff_sep = $i;
 
-      // get the first element
-      match $submac!(input, $($args2)*) {
-        $crate::IResult::Error(_)      => $crate::IResult::Done(input, ::std::vec::Vec::new()),
-        $crate::IResult::Incomplete(i) => $crate::IResult::Incomplete(i),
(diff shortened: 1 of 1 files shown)


tstorch:
I personally would remove the part

```rust
if i == input {
  return $crate::IResult::Error($crate::Err::Position($crate::ErrorKind::SeparatedList,input));
}
```

because, if someone calls this with element and separator validly being empty, he/she is an moron and deserves infinite loops.

But it was mentioned in #166 and I respect that :stuck_out_tongue_winking_eye:

tstorch:
I don't know why travis fails. Everything is fine, when I run `cargo test`.

Geal:
it looks good. I'm launching a new Travis build, it happens sometimes.

Now that I think of it, `separated_list!` and `separated_nonempty_list!` should accumulate size if they msut return `Incomplete`, as it is done in `many0!` and `chain!`. I'll have to add it soon.

Geal:
See, it passes the tests now :)
Thanks for your help!

tstorch:
I thought about the thing about the separator having to consume...

I think this is a bit restrictive. Shouldn't it be the way, that one loop trip must consume, i.e. (elem + sep).len() != 0 (<- awsome pseudo code). This way many0 would be a special case of separated list (if you would implement it this way, depends on the performance, but i'm not the right guy to talk about this) and it would be a little bit more generic.

For the code: You would just have to switch variables (if i'm not mistaken):

```rust
if i == input {
  return $crate::IResult::Error([stuff here]);
}
```

to

```rust
if i == buff_sep {
  return $crate::IResult::Error([stuff here]);
}
```

Thoughts?

PR #176, opened 2016-01-21
Sep list trip consume

tstorch:
Now you are able to have empty separators. This will only fail if

`elem.len() + sep.len() == 0`

in one loop trip.


tstorch:
I close this, bacause this is included in another pull request

PR #178, opened 2016-01-22
Sep many revamp

### LINKED ISSUE #177: many0!(...) is not equal to separated_list!(tag!(""), ...)
This took me some time to understand, but I now think why that is. To get a hint try to change in current master 704f508 in marcros.rs line 1865

``` rust
if ($i).input_len() == 0 {
  $crate::IResult::Done($i, ::std::vec::Vec::new())
}
```

with

``` rust
if ($i).input_len() == 0 {
  return $crate::IResult::Done($i, ::std::vec::Vec::new());
}
```

I thought this would change nothing, but it did. The fact that this is a macro means, that it is placed verbatim, where it's called. This is the reason, why this expression does not always lead the "calling function" to return this. In separated_list return is used instead of expression return.

@Geal Are you aware of this? Should it be this way? should I change separated_list to mimic this?

tstorch:
This streamlines `separated_list` and `many0` and modifies/adds th test for `separated_list`.
Now `many0!(...)` and `separated_list!(tag!(""), ...)` are interchangable.

Fixes #177 
Fixes part of #175


tstorch:
There is a small mistake, but I don't have time now. Will fix it on Monday. Created pull request so that you know what I'm up to...

tstorch:
Tests will succeed once #180 is accepted

tstorch:
Corrected and split up in #185 and #186

PR #180, opened 2016-01-25
Updated obsolete tag macro in macros test

tstorch:
ATTENTION: This will fail some test. This is due to the fact, that the test in macors.rs previously used an obsolete and very wrong tag macro

This is a hotfix! One needs to restructure the code (probably splitting of named! macro and others in macros.rs to other files and reordering uses in lib.rs) in order to eliminate the need to copy code inside the crate

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2493,7 +2493,7 @@ mod tests {
     assert_eq!(r2, Done(&b"WXYZ"[..], C{a: 1, b: None}));
 
     let r3 = f(&b"abcdX"[..]);
-    assert_eq!(r3, Incomplete(Needed::Size(8)));
+    assert_eq!(r3, Done(&b"X"[..], C{a: 1, b: None}));
   }
 
   use util::{error_to_list, add_error_pattern, print_error};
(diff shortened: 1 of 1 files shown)


tstorch:
Tests are fixed now

Geal:
Hi!
The nice thing about that `tag!` implementation is that it made it easy to test for `Incomplete` returns. Could you add other tests for this? Like, for the test at `macros.rs:2481`, test also for `"abcdef"`.

tstorch:
I'll have a look at it later this week (pretty busy days coming up).

The thing with this kind of stuff is, that it is very confusing to have two different implementations of a function/macro. At first I thought this was the usual tag macro being used. I more or less stumbled over this definition by accident.

Someone should really restructure this in order to use the real tag and take macro defined in bytes.rs in the test.

Geal:
Yes, having macros imports affected by ordering is really annoying :/

tstorch:
Git Tree f*ck up (is now more or less a shrubbery and no tree :stuck_out_tongue_winking_eye: ). Opened new PR #184 .

PR #182, opened 2016-01-26
Added take_until_s macro and test functions

joelself:
Added a take_until_s (to go along with the take_until_and_consume_s macro I added previously) and also added succeed, incomplete, and error test functions for it. Added TakeUntilStr to the Error enums in util.rs.

Also re-fixed take_s to use character offsets instead of byte offsets and re-added the take_s_succeed test. I thought I had this macro fixed and the test added in my previous merge request, but something must have happened.

Currently `cargo test` fails on an unrelated test (`separated_list`'s tag("") doesn't go over well because the first parse doesn't consume anything, so the input stays the same and that causes the macro to return an Error).


Geal:
Thanks!

PR #183, opened 2016-01-27
Added is_a_n_s and is_not_n_s macros to str module

joelself:
I added macros similar to is_a_s and is_not_s that take a count expression and only take exactly that many characters. An example where this is needed (where I currently use regex parsers) is for parsing DateTime strings. A date regex for example is `[0-9]{4}\-[0-9]{2}\-[0-9]{2}`. With is_a_s I can specify I want  some number of characters "01234567879" just like `[0-9]` but I can't say I only want exactly 4 for the year and 2 for the month and day.

With `is_a_n_s` you can specify what characters you want to parse and exactly how many you need. The same goes for `is_not_n_s`. You specify what characters you don't want and exactly how many of the other characters you want to parse.

Now I can write
```rust
named!(date_parse<&str, &str>,
  chain!(
    year:   is_a_n_s!(4, "0123456789")~
            tag_s!("-")               ~
    month:  is_a_n_s!(2, "0123456789")~
            tag_s!("-")               ~
    day:    is_a_n_s!(2, "0123456789"),
    ||{Date{year: year, month: month, day: day}}
  )
)
```
to parse a date without regexes.


Geal:
Is it necessary to add this when you can do `flat_map!(is_a_s!("0123456789"), take!(4))`?

joelself:
Wouldn't that allow something like this, if parsing a date with an expected 4 digit year:

`25000000-03-01`
`
`is_a_s!("0...9)` would take all of `225000000`, then `take!(4)` would just grab the first 4 of that. Yeah, I tried it, it grabs "2500" and the leftover is "-03-01". I need something to take only n and fail otherwise. And probably in the future (like the many_m_n macro) take between m - n.

Geal:
I think I would be happier about a `take_while_n!` and `take_till_n!`, it would be more general, and can provide the same features

joelself:
That makes sense. I'll work on that.

PR #184, opened 2016-01-28
Updated tag and test macro and test overhaul

tstorch:
Becaus I f*cked up my git tree a little bit, I need to open this as a new PR. This is basically what was ment to be in #180 

Now the tag and take macros in the test in `macros.rs` are those from `bytes.rs` (copied). To get a better test coverage the test in `macros.rs` got a treatment. I spared some, because they might fail after an overhaul of the according macros.

Someone needs to refactor the macros to ditch circular inclusions, resp. copying code. I will make a proposal for this in the near future via a PR.

Based upon this commit I will rewrite the separeted_list and many0 macros.

PR #185, opened 2016-01-29
Many0: overhaul, additional test, dedicated error

tstorch:
Rewrite of the `many0` macro (shorter and more comprehensible), now with own ErrorKind, and more tests for this macro.

One `assert_eq` will fail in current master due to wrong `tag` macro in the tests in `macros.rs`. Once #184 is merged this test will also succeed.


Geal:
Hi! I'm finally taking the time to review that pull request, sorry for the delay. The reason I jumped through hoops with `Vec` creation is that reallocations can be a bit costly. I'll take your version anyhow, since it's much cleaner, and see what I can improve from there.

src/util.rs
```diff
@@ -641,28 +642,29 @@ pub fn error_to_u32<E>(e: &ErrorKind<E>) -> u32 {
...
-    ErrorKind::LengthValueFn             => 22,
-    ErrorKind::Eof                       => 23,
-    ErrorKind::ExprOpt                   => 24,
-    ErrorKind::ExprRes                   => 25,
-    ErrorKind::CondReduce                => 26,
-    ErrorKind::Switch                    => 27,
-    ErrorKind::TagBits                   => 28,
-    ErrorKind::OneOf                     => 29,
-    ErrorKind::NoneOf                    => 30,
+    ErrorKind::Many0                     => 9,
```
Geal: error codes should not be reordered

tstorch:
No Problem. Nice to hear that it's in :smile: 

What should I do about the error codes?

Geal:
Nothing, I'm taking care of it while merging the PR :)

tstorch:
That's how I like to work :grin:

PR #188, opened 2016-01-29
Count: macro revamp

tstorch:
More compact code. Dropped some variables.


Geal:
Nice one! It is definitely cleaner!

PR #191, opened 2016-01-29
Addition cond_with_error

cholcombe973:
Hi Geal,

I don't know if you'll find this useful like I have but I figured I'd contribute back what I needed.  When I used cond! I got the following error:
```
<nom macros>:11:1: 11:26 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
<nom macros>:11 $ crate:: IResult:: Error (
                ^~~~~~~~~~~~~~~~~~~~~~~~~
<nom macros>:5:7: 5:42 note: in this expansion of count! (defined in <nom macros>)
<nom macros>:42:35: 42:70 note: in this expansion of cond! (defined in <nom macros>)
<nom macros>:49:19: 51:13 note: in this expansion of chaining_parser! (defined in <nom macros>)
```
I tried using cond_reduce! but I really want an Option returned.  My use case is the following:
https://play.rust-lang.org/?gist=b64e6a2d9e7071ce194a&version=stable
I'd like to optionally parse a field and have it return None otherwise.  Here is my working example:

https://gist.github.com/cholcombe973/e200c877dbf450e4df7d#file-versioned-rs-L35
In that example if my fn is passed a head_version >= 2 then I parse the field, otherwise I skip it without erroring out :)

This function probably needs a better name but I'll leave that up to you.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1413,6 +1413,59 @@ macro_rules! opt_res (
   );
 );
 
+/// `cond_with_error!(bool, I -> IResult<I,O>) => I -> IResult<I, Option<O>>`
+/// Conditional combinator
+///
+/// Wraps another parser and calls it if the
+/// condition is met. This combinator returns
+/// an Option of the return type of the child
+/// parser.
+///
+/// This is especially useful if a parser depends
+/// on the value return by a preceding parser in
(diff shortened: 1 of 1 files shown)


Geal:
Thanks

PR #192, opened 2016-02-01
Adds method-making and method-calling macros to nom

joelself:
This adds macros for creating parsing methods and calling parsing methods. This way you can make a parsing struct and save parsed data to struct member variables instead of global variables. The alternative to global variables would be to do post-processing of the AST after parsing, but it's much more efficient to do on the fly. So the to avoid global variables and post-processing I made these method macros. You can read more on why I made this here: https://github.com/joelself/nomplusplus/blob/methods/README.md

You create a method like this:
```rust
impl<'a> SomeStruct<'a> {
  //      --self's type---                        -self-
  method!(<SomeStruct<'a>, &str, &str>, take4, mut self, take_s!(4));
}
```
And you call a parsing method like this:
```rust
impl<'a> SomeStruct<'a> {
  method!(<SomeStruct<'a>, &str, &str>, take4, self, take_s!(4));
  //                                                            -struct.method-
  method!(<SomeStruct<'a>, &str, &str>, call_take4, self, call_m!(self.take4));
}
```
You can read more about it on the README above. I've already got a full Toml parser up and running using methods. Manipulating `self` in, for example, the post-processing sections of `chain!` and `alt!` can be tricky to do without copying because of the lifetimes of any borrows, but it's doable. Currently I have working code for detecting heterogeneous arrays (which are not allowed in Toml) and then inserting the array into an error list that is a member variable of the `Parser` struct. I'm working on inserting key/val pairs into a `HashMap` member, but it's considerably more tricky. I really should've used `Rc`, `RefCell`, and `Cell` everywhere instead of plain references, but I digress.

I'm currently working on modifying the parser combinators in `macros.rs` to let you call methods without an explicit invocation of `call_m!` just like you don't need an explicit call to `call!` when using functions: https://github.com/joelself/nomplusplus/blob/methods2.0/src/macros.rs#L233


Geal:
Hi, could you squash a bit this commit history? 39 commits is a bit much :)

Also, could you add a short sentence at the top of the documentation for the `methods.rs` file to say that it makes parsers as methods of other structs? The current documentation goes directly into the implementation explanation.

joelself:
Ok, I'm not totally sure what I'm doing with git, but I think I just took it down to 4 commits + a merge with current upstream master + change to module documentation. I did a lot of experimenting, and then towards the end, I threw out everything I did and started from scratch so I was able to squash something like the first 36 commits into one.

Geal:
what is the `str.rs` file you added at the top level? The pull request still contains 38 commits, with lots of test code. You can reduce them easily using the command `git rebase -i`. See http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html for more information.

Geal:
Also, can you do a `git pull --rebase` when you integrate the changes from master? It makes the history easier to follow

joelself:
I did do a git `git rebase -i HEAD~39`, then selected to squash the first 35 or so commits and then spent the next 2 hours merging stuff. If I try to rebase again it only shows the 4 non squashed commits (plus the 2 I added afterwards) then a bunch of your old commits. My git log also just shows the last 6 commits then your old commit. Even github shows the squashed commits: https://github.com/joelself/nomplusplus/commits/methods-merge.

I'm going to try to close this pull request and see if making a new one helps.

I'll do `git pull --rebase upstream/master` from now on.

joelself:
Closing to retry a pull request after a rebase.

PR #193, opened 2016-02-03
Add method making/consuming macros redux

joelself:
After some trial and error, I finally got the rebase to do what I want. The trick was `git rebase -i -s recursive theirs upstream/master` instead of rebasing against the same branch 39 commits ago. I also removed that stray str.rs file. Not sure where that came from.

This adds macros for creating parsing methods and calling parsing methods. This way you can make a parsing struct and save parsed data to struct member variables instead of global variables. The alternative to global variables would be to do post-processing of the AST after parsing, but it's much more efficient to do on the fly. So the to avoid global variables and post-processing I made these method macros. You can read more on why I made this here: https://github.com/joelself/nomplusplus/blob/methods/README.md

You create a method like this:
```rust
impl<'a> SomeStruct<'a> {
  //      --self's type---                        -self-
  method!(<SomeStruct<'a>, &str, &str>, take4, mut self, take_s!(4));
}
```
And you call a parsing method like this:
```rust
impl<'a> SomeStruct<'a> {
  method!(<SomeStruct<'a>, &str, &str>, take4, self, take_s!(4));
  //                                                            -struct.method-
  method!(<SomeStruct<'a>, &str, &str>, call_take4, self, call_m!(self.take4));
}
```
You can read more about it on the README above. I've already got a full Toml parser up and running using methods. Manipulating `self` in, for example, the post-processing sections of `chain!` and `alt!` can be tricky to do without copying because of the lifetimes of any borrows, but it's doable. Currently I have working code for detecting heterogeneous arrays (which are not allowed in Toml) and then inserting the array into an error list that is a member variable of the `Parser` struct. I'm working on inserting key/val pairs into a `HashMap` member, but it's considerably more tricky. I really should've used `Rc`, `RefCell`, and `Cell` everywhere instead of plain references, but I digress.

I'm currently working on modifying the parser combinators in `macros.rs` to let you call methods without an explicit invocation of `call_m!` just like you don't need an explicit call to `call!` when using functions: https://github.com/joelself/nomplusplus/blob/methods2.0/src/macros.rs#L233


Geal:
Hi!

I just assembled your commits in one and pushed it to master (commit cf2ee25570c1413ce182c2607b7ab661226fd7a4). It looks good so far.

I'm happy to see this merged at last! Next time, please use `git pull --rebase` instead of merging changes from master, and squash the test commits. AS you can see, the end result can be small and more manageable for

PR #200, opened 2016-02-03
delimited test overhaul

tstorch:
More complete delimited test

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2925,30 +2925,14 @@ mod tests {
     named!( tag_ghi, tag!("ghi") );
     named!( delimited_abc_def_ghi<&[u8], &[u8]>, delimited!(tag_abc, tag_def, tag_ghi) );
 
-    let done = &b"abcdefghijkl"[..];
-    // let parsed_pre = &b"abc"[..];
-    let parsed_main = &b"def"[..];
-    // let parsed_post = &b"ghi"[..];
-    let rest = &b"jkl"[..];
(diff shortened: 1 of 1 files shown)


Geal:
It looks like `delimited` does not count data correctly when it's incomplete.

For `ab` it should give `Incomplete(Needed::Size(3))` (correct), for `abcde` it should give `Incomplete(Needed::Size(6))` (correct), for `abcdefgh` it should give `Incomplete(Needed::Size(9))` (correct). Can you update the test to reflect that? I'll fix `delimited` to match those tests.

tstorch:
Done with a rebase

tstorch:
Your idea seems reasonable.

I'm not that proficient in macros, but could you use `tuple` to implement `preceded`, `terminated` and `delimited`, i.e. "`preceded!(a, b)` = `tuple!(a, b).1`" and "`terminated!(a, b)` = `tuple!(a, b).0`"? Or could you use `preceded` and `terminated` to implement `delimited` a la `delimited!(a, b, c)` = `preceded!(a, terminated!(b, c))` = `terminated!(preceded!(a, b), c)`?

Geal:
Yes, they could be done either way. It is easy to match on the result of `tuple` and select the needed output. I'll implement that when I get the time

Geal:
I'm currently updating code to use tuple everywhere, so I'm merging all those PRs

tstorch:
Nice!

PR #201, opened 2016-02-03
terminated test overhaul

tstorch:
More complete terminated test

This test will succeed once #184 is merged

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2912,10 +2912,28 @@ mod tests {
 
   #[test]
   fn terminated() {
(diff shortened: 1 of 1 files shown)


Geal:
same as `delimited`, `terminated` does not calculate offsets correctly. I think implementing combinators like `delimited` or `terminated` using `tuple` instead of their own code would solve everything easily.

Geal:
Can you update the tests to expect `Incomplete(Needed::Size(8))` for `abcde`?

tstorch:
Done with a rebase

Geal:
Thanks

PR #204, opened 2016-02-03
pair test overhaul

tstorch:
More complete pair test

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2836,10 +2836,26 @@ mod tests {
 
   #[test]
   fn pair() {
-    named!(p<&[u8],(&[u8], &[u8])>, pair!(tag!("abcd"), tag!("efgh")));
+    named!( tag_abc, tag!("abc") );
+    named!( tag_def, tag!("def") );
+    named!( pair_abc_def<&[u8],(&[u8], &[u8])>, pair!(tag_abc, tag_def) );
 
-    let r1 = p(&b"abcdefghijkl"[..]);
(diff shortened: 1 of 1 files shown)


Geal:
After reading a few of these, I'm not sure defining all the input and output slices before the asserts improves the readability. Having all the related data on on line or two close lines looks better

tstorch:
I see, that this might some time to get used to. The main reason I did this, was because I could write the test quite fast with this template. (Also if you know this template it's quite easy to read, because you just have to look at the variables)

I could change this, if you insist. Just give me a template version of some test.

Geal:
It is a lot easier to debug code if the slices are right where the assert is. With the following:

```rust
 assert_eq!(pair_abc_def(done), Done(rest, (parsed_1, parsed_2)));
```
I have to go look where `done`, `rest`, `parsed_1`, and `parsed_2` are, so I have to go back and forth between the assert and the slice definitions. With the following:

```rust
 assert_eq!(pair_abc_def(&b"abcdefghijkl"[..]), Done(&b"ghijkl"[..], (&b"abc"[..], &b"def"[..])));
```

If the test fails, I have everything on one line and I can understand directly what's going on, moreover when there's a progression in the tests with the surrounding asserts.

If you want to make it more readable, you could put the slice definitions just above the related assertion. But as a rule, I'd prefer that you follow the current style of the code when you add tests instead of rewriting them entirely.

Geal:
Thanks

PR #213, opened 2016-02-05
recognize! sometimes panics when it leaves no more input

Xion:
I've encountered an issue with `recognize!` panicking with:
```
---- recognize_leaving_empty_input stdout ----
	thread 'recognize_leaving_empty_input' panicked at 'arithmetic operation overflowed', src/util.rs:137
```

when applied to an input that is consumed wholly by it. I reproduced it with `is_a!`, which is where I encountered it, though it may also occur for other macros.

The included test case is a minimal one where the issue still existed. My fix went directly to `recognize!`, though it's possible the root cause may be in `HexDisplay::offset`, which `recognize!` uses, or perhaps `is_a!`.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -20,8 +20,12 @@ macro_rules! recognize (
       use $crate::HexDisplay;
       match $submac!($i, $($args)*) {
         $crate::IResult::Done(i,_)     => {
-          let index = ($i).offset(i);
-          $crate::IResult::Done(i, &($i)[..index])
+          if i.is_empty() {
+            $crate::IResult::Done(i, $i)
+          } else {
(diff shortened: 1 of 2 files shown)


Geal:
Hi, this is actually an error in `is_a!`. For a few parsers, when the whole input is consumed, there's a default result that can look like this: `IResult::Done(&b""[..], input)`. The issue with that is that this empty slice is stored elsewhere, and its address can be lower than the starting address of `input`, thus creating the overflow you saw.

I updated all those parsers to return something like `IResult::Done(&input[input.len()..], input)` which will work correctly with `offset()`.

It was fixed in 2927310.

PR #214, opened 2016-02-05
Fix take_while_s and take_while1_s doc entries.

lwandrebeck:


lwandrebeck:
Added a fix for fn hex_digit ErrorKing return type if input_lenght == 0

Geal:
Thanks!

PR #216, opened 2016-02-06
Some more fixes to str.rs functions documentation.

lwandrebeck:


Geal:
Why put back `take_while_s!(&str -> bool)`? It's not a `&str` as argument

Geal:
Thanks!

PR #218, opened 2016-02-09
Add IResult::map and IResult::map_err

sourrust:
I mostly want to use map_err to allow my flac library to have better custom errors, but I also add map as well to have a similar interface as `Result`.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -80,6 +80,9 @@ impl<I,O,E> IResult<I,O,E> {
     }
   }
 
+  /// Maps a `IResult<I, O, E>` to `IResult<I, O, N>` by appling a function
+  /// to a contained `Error` value, leaving `Done` and `Incomplete` value
+  /// untouched.
   #[inline]
   pub fn map_err<N, F>(self, f: F) -> IResult<I, O, N>
    where F: FnOnce(Err<I, E>) -> Err<I, N> {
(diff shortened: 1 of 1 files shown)


Geal:
Thanks!

tstorch:
@Geal This was on my TODO list for this week. I planed to rewrite a major part of `macros.rs` using this. Interested?

Also I thought about writing a combination of `IResult::map` and `Needed::map`, which as a method of `IResult` takes to functions `f` and `g` and returns `IResult::Done(_, f(o))` resp. `IResult::Incomplete(Needed::Size(g(i)))` (or `IResult::Incomplete(Needed::Unknown)` or `IResult::Error(_)`. Your thoughts?

Geal:
@tstorch in short, don't try that ;)

That was my first approach with nom, and the one favored by libraries like combine. You end up writing specializations for a lot of type combinations, and fight the compiler for lifetime issues. `map` and `map_err` here are more useful to manipulate the result of nom parsers elsewhere in the code.

The macros approach is tricky, but the principle is simple, and it relies on the compiler being smart enough to guess types and lifetimes in one function scope. If you want to try and rewrite stuff, there are already the `map!`, `map_opt!`, `flat_map!` macros you can use.

I'd advise against touching `chain!`, though. This parser is overly complex, I'd like to replace it with something simpler to use.

tstorch:
Got it! I'll have a look at it.

Is there a particular reason, why `IResult::map_incomplete` is missing?

sourrust:
@tstorch no real reason, I was just matching the `Result`API.

PR #219, opened 2016-02-10
Changed name of of `apply_rf` to `apply_m`

joelself:
The 'm' is for method. This was in my code at some point, but must have gotten lost in a merge.


Geal:
Thanks

PR #221, opened 2016-02-15
Added map (and is_known) function for enum Needed

tstorch:
Because @sourrust snagged #218, which was on my TODO list for this week I implemented map for Needed, which might come in handy (esp. in the `chaining_parser` macro)


Geal:
Thanks

PR #222, opened 2016-02-15
Added IResult:: map_inc (map incomplete)

tstorch:
Addedum to #218. I figured it might be useful in some cases as it would be in the `complete` macro.


Geal:
Thanks

Issue #181, opened 2016-01-25
nom::Err should implement std::error::Error


rorytrent:
This will allow usage of the nom::Err type with io::Error::new(io::ErrorKind, nom_err).

Geal:
Fixed in 40c0e55

Issue #66, opened 2015-08-03
Bit field parsers


Geal:
As mentioned by @rrichardson in [https://github.com/UpstandingHackers/hammer/issues/64#issuecomment-126709170](https://github.com/UpstandingHackers/hammer/issues/64#issuecomment-126709170), parsers that transform from bit positions to a tuple of fields would be useful:

```rust
pub fn be_bits_1<'a, A>(i: &[u8], u8) -> IResult<'a,&[u8], (A)>
pub fn be_bits_2<'a, A, B>(i: &[u8], u8, u8) -> IResult<'a,&[u8],( A, B )>
pub fn be_bits_2<'a, A, B>(i: &[u8], u8, u8, u8) -> IResult<'a,&[u8],( A, B, C )>.
//.. up to 8 or so and also for le

// + a small bit of  macro magic to streamline the chaining 

// So to parse something like a TCP header, one would do something like: 
chain!(
    src_prt::  be_u16 ~ 
    dst_prt :  be_u16 ~ 
    seq_num: be_u32 ~
    ack_num: be_u32 ~
    (offs, _, flags) : be_bits_3( 4, 3, 9 ) ~ 
    blah: blah ~
) 
```

Geal:
up, this might be useful

Issue #65, opened 2015-07-21
Difference with Parsec nomenclature


Geal:
Some of Parsec's functions are available in nom, but not all of them, and not always with the same name:
- [ ] [character level parsers](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Char.html)
- [ ] [permutation parsers](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Perm.html)
- [ ] [parserZero](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html) always fails
- [ ] [(<?>) or label](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html) replaces the error with a higher level error
- [ ] [lookAhead and try](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html) is named `peek!`
- [ ] [tokens](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html)
- [ ] [many](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html) is called `many0!`
- [ ] [skipMany](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Prim.html)
- [ ] [lexers](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Token.html)
- [ ] [count](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html) is called `count_fixed!`
- [ ] [between(open,end,p)](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html) is called `delimited!(open, p, end)`
- [ ] [option](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html)
- [ ] [optionMaybe](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html) is called `opt!`
- [ ] [optional](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html)
- [ ] [skipMany1](https://hackage.haskell.org/package/parsec-3.1.9/docs/Text-Parsec-Combinator.html)
[...]

Issue #40, opened 2015-05-22
Permutation parser


Geal:

filipegoncalves:
What exactly does this mean?

Geal:
That's a combinator already present in [Parsec](https://hackage.haskell.org/package/parsec-permutation-0.1.2.0/docs/Text-Parsec-Permutation.html).

Basically, it will return a tuple of results of its children, but its children can be called in arbitrary order. This is useful for formats like PNG or MP4.

PR #223, opened 2016-02-15
Minor standardizations

tstorch:
This is a standardization patch:

- standard pattern in matching IResult and Needed (i.e. Done(i,o), Error(e), Incomplete(x); Size(n) [varibale names and ordering])
- Contracting Incomplete(Unknown) and Incomplete(Size(n)) to  Incomplete(x) if possible
- submac, args -> submac1, args1 if submac2 resp. args2 is present

I know this is a lot, but this are mere renamings and reorderings.
I thought this was a good idea. In some parts of the code I always had to look up the variables...

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -138,7 +138,7 @@ macro_rules! tag_bits (
   ($i:expr, $t:ty, $count:expr, $p: pat) => (
     {
       match take_bits!($i, $t, $count) {
-        $crate::IResult::Incomplete(i) => $crate::IResult::Incomplete(i),
+        $crate::IResult::Incomplete(x) => $crate::IResult::Incomplete(x),
         $crate::IResult::Done(i, o)    => {
           if let $p = o {
             let res: $crate::IResult<(&[u8],usize),$t> = $crate::IResult::Done(i, o);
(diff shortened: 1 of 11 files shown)


Geal:
I'm sorry, I can't accept that patch. I don't feel it improves readability a lot, and it will mess with git blame's output (useful when you want to know the original idea behind a line).

Contracting `Incomplete(Unknown)` and `Incomplete(Size(n))` is fine, contracting `Incomplete` and `Error` where it makes sense too, but the rest does not add much value. If looking up the variable names makes the code look complex, you could find a way to break it up in smaller parts.

Moreover, on a more personal note, I do not have much time to spare for nom right now, and I cannot spend it reviewing huge patches like those.
If you have more time than me, how about using it on more interesting issues, like those (ordered by ascending difficulty):
- #181 implement `std::error::Error` for `nom::Err`
- #66 bit field parsers (this one should be pretty nice to do with `tuple`)
- #65 Providing a Parsec compatible API
- #34 document how to parse space separated formats, like languages. I feel there could be a magic combinator that will intersperse space parsers everywhere, but I have not put much thought to it yet
- #40 Permutation combinator. That one is probably very tricky, but will be interesting for some binary formats

There's also a lot of work to do around writing example parsers, testing the producers and consumers, and writing tutorials.

tstorch:
Sorry for bombing you with stuff. I just thought that the code could use some standardization and I personally do not read git history that much (read: only if its really necessary).

I'll have a look at the other stuff. Do you close PR or should I do it?

PR #232, opened 2016-03-04
Make take_until_s! work with strings that are not 3 characters long

conradev:
Me until I figured this out:


Geal:
Nice catch, thanks!

PR #237, opened 2016-03-16
Absolute macro references

DanielKeep:
This makes some `ErrorKind` references in macros absolute.

I noticed that *most* references to stdlib names are not absolute, and references to things like `::regex` aren't portable, either.  I wasn't sure if you'd want those fixed, too, so I just left things at nom's own names.

Incidentally, I also fixed all warnings I found when running tests.


Geal:
Thanks!

PR #238, opened 2016-03-19
Don't consume list separator prematurely

evestera:
When using the macros `separated_list!` and `separated_nonempty_list!` the separator is consumed before checking if there is another element. This prevents using it in cases like this:

        elems: separated_list!(tag!("."), id) ~
        tag!(".*") ,

Not consuming the separator until after another element is parsed fixes this. Fixed, and added a test for the fix.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1767,11 +1767,10 @@ macro_rules! separated_list(
                 if i2.len() == input.len() {
                   break;
                 }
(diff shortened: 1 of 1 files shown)


Geal:
Merged in https://github.com/Geal/nom/commit/d42cf5c7d0c1607f5676762c05c494b7250e76ae

Thanks!

Issue #243, opened 2016-04-03
Missing le_f32 function.


jeandudey:
There is a function to read f32 values in big endian (`be_f32`), but not for little endian (`le_f32`).

Geal:
It's not there because nobody has needed it yet ;)

Would you like to give it a try? That pull request should be easy to do.

jeandudey:
I will do it, seems easy :D.

Geal:
Thanks!

PR #244, opened 2016-04-04
Add le_f32 and le_f64 functions.

jeandudey:
Solves #243.


Geal:
Hi, could you add some unit tests for those functions?

Geal:
Hello!

How about some unit tests?

I know, I insist, but it's useful to know the base parsers always work as expected ;)

PR #248, opened 2016-04-11
Test: Fix a wrong test case for `is_oct_digit`

Hywan:
`is_hex_digit` is used instead of `is_oct_digit`.


Hywan:
Tests fail on nightly, don't know why, but it seems not to be related to this PR.

Geal:
Thanks!

Hywan:
'lcome :-)

PR #249, opened 2016-04-11
Fixed take_till! to be usable from the outside of nom crate

Xirdus:
Because of private trait macro shenanigans, it was impossible to use this macro due to compilation error. The attached test, for example, doesn't compile without this patch.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -491,7 +491,7 @@ macro_rules! take_till (
     {
       match $input.iter().position(|c| $submac!(c, $($args)*)) {
         Some(n) => $crate::IResult::Done(&$input[n..], &$input[..n]),
-        None    => $crate::IResult::Done(&$input[($input).input_len()..], $input)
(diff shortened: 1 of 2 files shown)


Xirdus:
The test doesn't compile on nightly.

```
<nom macros>:7:13: 7:53 error: module `util` is private
<nom macros>:7 & $ input [ $ crate:: util:: InputLength:: input_len ( & $ input ) .. ] , $
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/issues.rs:126:18: 20:36 note: in this expansion of take_till! (defined in <nom macros>)
tests/issues.rs:125:5: 127:7 note: in this expansion of named! (defined in <nom macros>)
```

Seems like they changed the privacy checks in macros to happen in the macro expansion scope instead of macro declaration scope. Works on stable and beta.

Xirdus:
I realized it's not private after all. Made better fix. Nightly still doesn't build, but it's not my fault anymore :)

Geal:
Thanks!

PR #250, opened 2016-04-13
Fix the doc string of take_bits

jespino:


Geal:
Already fixed in 344c68d
Thanks anyway!

PR #256, opened 2016-05-04
Add crushtool repo

cholcombe973:
I've created a nom parser that can decompile and recompile Ceph compiled crushmaps.


cholcombe973:
I'm not sure how the builds or coveralls could fail.  It's only a readme update

Geal:
Thanks! You should add a readme to your project repository, it's a bit raw right now :)

PR #257, opened 2016-05-05
Arithmetic expression parser to AST.

mspiegel:
Adapts tests/arithmetic.rs to generate an abstract syntax tree.


Geal:
Thanks, this looks great!

Arithmetic expressions are simple enough to be interpreted directly with a fold, but I have a feeling nom could be used for the next step as well. Since it is generic over the input type, you could write recognizers for a slice of a vector of tokens. I guess it will be the next test :)

PR #263, opened 2016-06-09
Fix clippy warnings on bit combinators

sourrust:
Small changes to `take_bits!` macro that were found in sourrust/flac#5.


Geal:
Thanks!

Issue #262, opened 2016-06-05
How to perform negative lookahead?


gtors:
I'm trying to translate this Scala code to Rust:

```scala
def kw(s: String) = s ~ !(letter | digit | "_")
```

But I can't find similar combinator like this http://www.lihaoyi.com/fastparse/#NegativeLookahead . 

```rust
fn kw(input: &[u8], s: &str) -> IResult<&[u8],&[u8]> {
    recognize!(tuple!(input, tag!(s)  ????))
    // In ideally, I think, it should be like this:
    // recognize!(tuple!(input, tag!(s), not!(lookahead!(alt!(alphanumeric | char!('_'))))))```
}
```

abiggerhammer:
FWIW this is basically Hammer's [`h_not`](https://github.com/UpstandingHackers/hammer/blob/master/src/parsers/not.c).

gtors:
@abiggerhammer thx for the tip

Geal:
merged

vwkd:
Was this ever implemented?

PR #264, opened 2016-06-15
Add negative lookahead

gtors:
#262

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1692,6 +1692,40 @@ macro_rules! peek(
   );
 );
 
+/// `not!(I -> IResult<I,0>) => I -> IResult<I, O>`
+/// returns a result only if the embedded parser returns Error or Incomplete
+///
+/// ```
+/// # #[macro_use] extern crate nom;
+/// # use nom::IResult::{Done, Error};
+/// # use nom::Err::{Position};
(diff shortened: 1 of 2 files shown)


Geal:
Hello,

could you add some unit tests with this?

src/macros.rs
```diff
@@ -1692,6 +1692,40 @@ macro_rules! peek(
...
+/// assert_eq!(r2, Error(Position(ErrorKind::Not, &b"e"[..])));
+/// # }
+/// ```
+#[macro_export]
+macro_rules! not(
+  ($i:expr, $submac:ident!( $($args:tt)* )) => (
+    {
+      match $submac!($i, $($args)*) {
+        $crate::IResult::Done(_, _)    => $crate::IResult::Error($crate::Err::Position($crate::ErrorKind::Not, $i)),
+        $crate::IResult::Error(_)      => $crate::IResult::Done($i, &b""[..]),
```
Geal: you should use `&($i)[..0]` instead of `&b""[..]`, because it is relative to the input instead of an empty string with a completely unrelated address (I got bit by bugs like this months ago). Other wise, you could just as easily return a `()`, since you do not intend to consume the value.

gtors:
Done

Geal:
Thanks!

PR #267, opened 2016-06-16
Add rest_s function

pwoolcoc:
Simple enough to write when needed, but I thought it should be in `nom` to go with the rest of the `_s` functions.

### DIFF
--- a/src/nom.rs
+++ b/src/nom.rs
@@ -529,6 +529,12 @@ pub fn rest(input: &[u8]) -> IResult<&[u8], &[u8]> {
     IResult::Done(&input[input.len()..], input)
 }
 
+/// Return the remaining input, for strings.
+#[inline]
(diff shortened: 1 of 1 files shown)


src/nom.rs
```diff
@@ -529,6 +529,12 @@ pub fn rest(input: &[u8]) -> IResult<&[u8], &[u8]> {
     IResult::Done(&input[input.len()..], input)
 }
 
+/// Return the remaining input, for strings.
+#[inline]
+pub fn rest_s(input: &str) -> IResult<&str, &str> {
+    IResult::Done("", input)
```
Geal: please use `&input[input.len()..]` as remaining input, since the slice's pointer is related to the input, while the empty string's address could be anywhere in the memory.

pwoolcoc:
@Geal fixed

Geal:
Thanks!

PR #270, opened 2016-06-22
Made documentation locations more clear.

Phlosioneer:
Puts them all under the the Documentation header, and adds the wiki as an explicit source of documentation.


Geal:
Hello, thanks for the suggestion, but I put the contact information at the top on purpose, so that people find it easily. So I'd prefer you do it the other way, putting the documentation link at the top with a clear message.

Geal:
the documentation links changed

PR #272, opened 2016-06-29
Fix #245

### LINKED ISSUE #245: Inconsistency between not_line_ending and line_ending
not_line_ending seems to match anything which is not \r or \n, whereas line_ending matches only \n.
I thing both should use the same set of line ending sequences.

See [this gist](https://gist.github.com/weiznich/eef28628d21a8ff5140bae5f84340ff1) for an example. The testcase full_line_windows fails for me with the following error message:
```
thread 'tests::full_line_windows' panicked at 'assertion failed: 
`(left == right)` (left: `Error(Position(Tag, [13, 10]))`, right: `Done([], ([97, 98, 99], [13, 10]))`)', lib.rs:29
```

weiznich:
line_ending should also match \r\n and not only \n

### DIFF
--- a/src/nom.rs
+++ b/src/nom.rs
@@ -54,7 +54,7 @@ pub fn not_line_ending(input:&[u8]) -> IResult<&[u8], &[u8]> {
   Done(&input[input.len()..], input)
 }
 
-named!(tag_ln, tag!("\n"));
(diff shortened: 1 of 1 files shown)


Geal:
Thanks!

PR #273, opened 2016-07-03
Add IResult::unwrap, IResult::unwrap_err, and IResult::unwrap_inc methods

fitzgen:
Similar to `Result::unwrap` and `Result::unwrap_err`.


Geal:
Thanks!

PR #276, opened 2016-07-09
Use named!() for the ini parser

acatton:
I think this was just old, but the custom `category` function could be defined this way.


Geal:
Nice one! It's a lot more readable that way.

PR #290, opened 2016-08-19
add a way to convert an IResult to a std::result::Result

Keruspe:


Keruspe:
This is more a RFC than a real pull request. Not sure which way you'd prefer to do things.

Geal:
I feel there should be two implementations: one that returns `Result<O, IError<I,E>>` and one that returns `Result<O, Err<I,E>>`. Can you add the other one?

Keruspe:
What do you think of this approach?

PR #296, opened 2016-09-02
Merge `_impl!` and `1!` macros into base macro

overdrivenpotato:
This should remove the need to import implementation macros when using the crate.


Geal:
Awesome work, thanks! This will simplify usage

PR #306, opened 2016-09-12
Fix Markdown issue in README

foophoof:
GitHub needs the final <code>```</code> to be on a line by itself.


foophoof:
This is now a duplicate of #312, which was merged, so closing this.

PR #307, opened 2016-09-14
Make not_line_ending polymorphic over input.

### LINKED ISSUE #285: not_line_ending should work on &[u8] and &str like alpha and other parsers

vickenty:
Make the function work with both &[u8] and &str.

Fixes #285.


Geal:
Sorry, I rewrote those functions in 26754f3 to use the new traits. Thanks for your help anyway!

PR #313, opened 2016-09-24
Change eol to test for newline first

keeperofdakeys:
If crlf is the first one to match, a single character will always return incomplete, even if it's a single '\n'. Change so that newline is the first test, so inputs of a single character are parsed correctly.


Geal:
Thanks!

PR #314, opened 2016-09-27
Document is_* tests for character types

jturner314:
These tests are especially important to document because the equivalent tests on Rust's `char` type account for Unicode, while these only check ASCII bytes.


Geal:
Thanks!

PR #315, opened 2016-10-03
Support for the Vec type with nightly Rust core

rudib:
Implementors need to provide the alloc and collections crates.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -1,17 +1,22 @@
 //! Parsers for applying parsers multiple times
 
+#[cfg(feature = "core")]
+pub type Vec<T> = ::collections::vec::Vec<T>;
+#[cfg(not(feature = "core"))]
+pub type Vec<T> = ::std::vec::Vec<T>;
+
 /// `separated_list!(I -> IResult<I,T>, I -> IResult<I,O>) => I -> IResult<I, Vec<O>>`
 /// separated_list(sep, X) returns Vec<X>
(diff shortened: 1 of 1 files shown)


Geal:
Hi,

Why isn't this working with the fake `std` module I declare when the `core` feature is activated? The `collections::vec` module is imported there: https://github.com/Geal/nom/blob/master/src/lib.rs#L106

rudib:
I have a no_std library that's built with xargo, here's the compilation error:
```
error[E0433]: failed to resolve. Maybe a missing `extern crate std`?
  --> /mnt/c/code/rust/sensors/src/buspirate/mod.rs:65:1
   |
65 | named!(u8_array< Vec<u8> >,
   | ^ Maybe a missing `extern crate std`?
   |
   = note: this error originates in a macro outside of the current crate
```
Using the ``::std`` prefix means that Rust should look for a crate named std, not a module, which you have defined. Even so, the std module from lib.rs is defined is private, as such the outside macro invocations won't be able to resolve to it. By changing ``::std`` to ``$crate::std`` and marking it as public (the outside world then sees it as ``nom::std``), I can get the cross-compiler to work with it.

I think that a cleaner solution would be to have a prelude module for collections and have it included only in macros that need them. I'll try to come up with a changeset for that.

Let's consider this solution outdated for now. I'm still learning how Rust expands macros from other libraries and their local imports.

PR #330, opened 2016-10-08
AsChar: rename is_0_to_9 to is_dec_digit

### LINKED ISSUE #310: Consistent method naming for AsChar
https://github.com/Geal/nom/blob/master/src/traits.rs#L49 should be `is_dec_digit`

Keruspe:
Fixes #310


Geal:
Thanks!

PR #331, opened 2016-10-08
Add an example test for parsing float

### LINKED ISSUE #325: example of a parser fo floating point numbers using `digit` and `recognize`
Recognize digits separated by a '.', then use `float::parse` on the resulting string.

Keruspe:
Fixes #325


ambaxter:
Would it make sense to test for the sign? (+/-).

Keruspe:
@ambaxter added

Geal:
Thanks!

PR #336, opened 2016-10-13
Add named_table! to match against many possible values

s3bk:
named_table! creates a generator that acts as a tag and matches one element of the input against the specified list. At the same time it declares an enum with all possible variants and returns the matched one.


Geal:
it feels a bit complicated compared to switch, but if you find a simpler approach it might be interesting. Closing this for now

PR #340, opened 2016-10-17
Use byteorder crate to parse integers and floating point numbers

Camille019:


SnirkImmington:
Is it worth adding a whole crate dependency to change 6 variable assignments?

Geal:
Hi!

Thanks for the pull request, but I'm not sure either that it warrants importing byteorder. Could you at least provide a little context on why you would need it? A pull request without any comment is a bit raw.

Geal:
I'd actually be interested in an assembly comparison of nom's and byteorder's method to convert ints :)

PR #344, opened 2016-10-20
Including doc comments in named! macro

keeperofdakeys:
This patch works, but needs some api considerations. Based off of work from https://github.com/Geal/nom/pull/317.


Geal:
Thanks!

Issue #328, opened 2016-10-08
remove usages of `chain` from documentation and examples


Geal:
replace it with do_parse

Geal:
replace it with `do_parse`

PR #346, opened 2016-10-22
Don't use chain! in nom internals

Keruspe:
Fixes part of #328

The only places where chain! is still listed/used are the Changelog, the nom-1.0 porting guide, the chain! documentation, the chain! unit tests, tsome arithmetic tests and the README.

Two other pull requests address the two latters, all the others should stay as is I guess.


Geal:
Thanks!

PR #347, opened 2016-10-22
Don't use chain! in arithmetic tests

Keruspe:
Fixes part of #328

This one is a little trickier, in order to drop the mut usage, I used fold_many which might be less intuitive.
I think it's a good start though, until we rewrite it another way maybe at some point.


Geal:
Thanks!

PR #352, opened 2016-10-28
Add MagicaVoxel .vox file parser

davidedmonds:


Geal:
Thanks, that's a great new addition :)

PR #359, opened 2016-11-04
Fix problem with IterIndices for str

jugglerchris:
The implementation of `IterIndices::index` for `&str` didn't correctly handle the case when the slice is exactly the right length.  Add a fix by checking for the right length if we get to the end of `char_indices()`.


Geal:
Thanks!

PR #361, opened 2016-11-06
Improve documentation for `alt!` and `alt_complete!`

azerupi:
I'm trying to improve the documentation a little and started with the `alt!` and `alt_complete!` combinators. Here is what I changed:

- Made the short description, shorter and easier to read.
  In my opinion, things like `alt!(I -> IResult<I,O> | I -> IResult<I,O> | ... | I -> IResult<I,O> ) => I -> IResult<I, O>` are too long to be in the short description. When all the combinators are listed on the front page of the docs it's easier to have a one line overview of what that combinator does, instead of a full type description.
- Added titles to differentiate between different sections of the page (e.g. example, manipulate results, ...).
- Changed the examples to be a little more readable by adding spaces, comments and making it a little more realistic.

I am not sure if you will like the changes, if not, feel free to close this PR. :smiley:

### DIFF
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -5,7 +5,7 @@
 /// ```
 /// All the parsers must have the same return type.
 ///
-/// If one of the parser returns `Incomplete`, `alt!` will return `Incomplete`, to retry
+/// If one of the parsers returns `Incomplete`, `alt!` will return `Incomplete`, to retry
 /// once you get more input. Note that it is better for performance to know the
 /// minimum size of data you need before you get into `alt!`.
 ///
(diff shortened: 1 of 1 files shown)


Geal:
Hey, thanks for your hard work on this, I like the theme around dragons and other monsters ;) 🐲 

I'd prefer keeping the type description at the top, because it makes it easier to find the combinator you want in the list. But I plan to make a separate page of the documentation containing a list of combinators with short explanation paragraphs, and a classification by theme.

azerupi:
Thanks :)

>I'd prefer keeping the type description at the top

I can understand that. However I would like to note for posterity that I personally find it very dense and hard to read. Especially when you have 50 of them in a row :wink: 

Considering that most combinators take `I -> IResutl<I, O>`, maybe it could help to define `P` as an alias for it? Reducing `alt!(I -> IResult<I,O> | I -> IResult<I,O> | ... | I -> IResult<I,O> ) => I -> IResult<I, O>` to `alt!(P1 | P2 | ... | Pn) => P`. I am not sure about the exact semantics, but it could be worth exploring to make the docs easier to read. What do you think?

lucab:
Just my 2¢: I was recently getting started with nom2 (to parse some linux procfs data) and I find @azerupi short descriptions much more easier to explore (at least as a newcomer).

I guess that (at least for me) the main difference is that not knowing the crate yet, having all the signatures in the index page makes it hard to navigate: I was trying to get a good overview of what's available in nom2, but I got derailed into cognitive overload trying to understand all those mostly-similar-but-slightly-different signatures.

Geal:
Thanks!

PR #376, opened 2016-11-22
Add bencode.rs to README.md

jbaum98:
I wasn't sure what category bencoding is, so I put it under Network protocol formats because of its use in torrent clients.


Geal:
Thanks!

PR #382, opened 2016-12-05
Rename add_error to add_return_error

nickbabcock:
I currently get a undefined `add_return_error!` on nom 2.0 and it [appears that it should be defined](https://github.com/Geal/nom/blame/75ad25152040b0e303eb2af7e962453f03fb6a68/doc/upgrading_to_nom_2.md#L140). This PR actually does the rename 😄 

EDIT: This would be a breaking change, not sure how you feel about that, since 2.0 was just released 😕

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -338,7 +338,7 @@ macro_rules! return_error (
 /// Add an error if the child parser fails
 ///
 /// While error! does an early return and avoids backtracking,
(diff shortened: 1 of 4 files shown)


Geal:
nice catch! This is effectively a breaking change, but since not many people use nom 2.0 yet, it should be fine. I know people can be very specific about their semver, but doing a 3.0 for this would be a bit ridiculous :)

PR #384, opened 2016-12-06
Make char! more general to accept &str.

### LINKED ISSUE #353: char! does not support &str

DarkDrek:
Fixes #353

### DIFF
--- a/src/character.rs
+++ b/src/character.rs
@@ -98,21 +98,28 @@ macro_rules! none_of_bytes (
   );
 );
 
-/// matches one character: `char!(char) => &[u8] -> IResult<&[u8], char>
+/// matches one character: `char!(char) => T -> IResult<T, char>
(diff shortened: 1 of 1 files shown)


Geal:
I avoided doing that change before because doing a comparison with only the first byte is wrong for chars in a &str. char and the other char-related combinators (one_of, etc) only really work on bytes right now.

ArtemGr:
> I avoided doing that change before because doing a comparison with only the first byte is wrong for chars in a &str.

Comparing with [Chars::next](https://doc.rust-lang.org/nightly/std/str/struct.Chars.html#next.v) should be okay though, shouldn't it?

Geal:
yes, it should

Geal:
this was fixed in a later refactoring. Thanks for the help, though!

PR #391, opened 2016-12-10
Error and Contrib doc updates

nickbabcock:
Minor updates to make the error documentation look better.

I also added a line in the contrib doc that'll help future me remember how to build nom's documentation 😄


Geal:
Thanks!

PR #400, opened 2016-12-18
Repair links in `home.md`

DirkyJerky:
.html -> .md, so the links actually point to something

### DIFF
--- a/doc/error_management.md
+++ b/doc/error_management.md
@@ -1,6 +1,6 @@
-% Error management
-
 # Error management
+
+## Error management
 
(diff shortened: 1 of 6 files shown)


Geal:
ok, so this is probably the third PR saying the links are wrong. They're not. They point to HTML files because they are meant for the online docs here: http://rust.unhandledexpression.com/nom/home.html

Those docs are generated with https://github.com/Geal/cargo-external-doc/ . It should be possible to make this tool transform .md links to .html automatically, but I have not looked into it.

Geal:
also, the `% title` is by design. This is how rustdoc finds the HTML title

PR #405, opened 2016-12-21
doc: Break long lines

lu-zero:
And try to prettify it a little.


lu-zero:
I'm starting with just one so you can tell me if you like the format :)

Geal:
the issue is that the macros list only displays the first line :/ But I should rewrite most of the docs, to make a markdown file describing combinators in a more readable way

PR #409, opened 2016-12-24
Moved Slice Implementation to Macros

tstorch:
This was something that seemed to offer itself to macrofying. Turned not out as nice as I wanted it to be, but better than nothing i guess. The clever people out there might be able to build a really nice version from this...


Geal:
thanks!

PR #410, opened 2016-12-24
InputLength known at compile time

tstorch:
We already know the input length at compile time. So why not use it?


tstorch:
Hmm, wait. The length might change during runtime :neutral_face:

PR #411, opened 2016-12-25
Add midi parser to list of parsers

derekdreery:


Geal:
thanks! what are you using midi for?

derekdreery:
At the moment it is just an experiment. It's the first time I've used nom, or parser combinators. I really like the principal though - how easy it is to test each small parser.

I think midi is a good example for where nom really excells - it has some unusual structures like a u32 that can be compressed using a continuation bit. If you implement these things by hand, you have to store some state yourself, wehereas using nom means it handles all the state for you :)

PR #412, opened 2016-12-30
Add explicit lifetimes for named! and named_attr!

bb010g:
Also normalizes pub and non-pub expansions. Previously only usable in
the private output-specified variation.


bb010g:
Not exactly sure what's happening with the coverage results. I'm not seeing why this would affect coverage on those lines (e.g. [these in multi.rs](https://coveralls.io/builds/9467178/source?filename=src%2Fbytes.rs#L956)).

Geal:
ok, so, what problem is it meant to solve? Rust has lifetime elision now, I removed those a long time ago.

bb010g:
Yeah, elision works. I'm stupid.

Issue #394, opened 2016-12-12
On Custom Errors


nickbabcock:
I've been playing around with nom and rust, I'm far from an expert but I wanted to address what I think is a major hurdle for nom users: errors. I think I've yet to see comprehensive error reporting with nom. Maybe it's just because I'm not familiar with parse combinators, but I feel like if recursive descent parsers can provide good error messages then so can parser combinators :smile:

Here's a toy example: decoding a utf8 string that is length prefixed and null terminated -- taken from a project I'm working on: [boxcars](https://github.com/nickbabcock/boxcars) 

```rust
/// Text is encoded with a leading int that denotes the number of bytes that
/// the text spans.
named!(text_encoded<&[u8], &str, BoxcarError>,
    fix_error!(BoxcarError,
    complete!(do_parse!(size: le_i32 >> data: apply!(decode_str, size) >> (data)))));

/// Reads a string of a given size from the data. The size includes a null
/// character as the last character, so we drop it in the returned string
/// slice. It may seem redundant to store this information, but stackoverflow
/// contains a nice reasoning for why it may have been done this way:
/// http://stackoverflow.com/q/6293457/433785
fn decode_str(input: &[u8], size: i32) -> IResult<&[u8], &str> {
    do_parse!(input, data: take_str!(size - 1) >> take!(1) >> (data))
}
```

What are all the errors that one may want to expose from just these two functions:

- Not enough data to read size of string (context: how much more data is needed)
- Nonsensical size of string (negative or too large)
- Not enough data to read string (context: how much more data is needed, and maybe even stringify what was read)
- String is not valid utf8 (context: what bytes were invalid and location of invalid bytes in string)
- String contains nonsensical characters (context: what's the string and invalid characters, and location of said characters)
- String does not end with a null character

So an enum like the following could be appropriate for errors (pseudocode):

```rust
#[derive(PartialEq, Debug, Clone, Copy)]
pub enum BoxcarError {
    SizeIncomplete(i32),
    InvalidSize(i32),
    StringIncomplete { decoded: String, needed: i32 },
    NonUtf8String { location: i32, invalid: &[u8] },
    InvalidString { decoded: String, invalid: char },
    MissingNullTermination
}
```

The end goal would be propagate all these errors (with the mentioned context) to upstream callers. Besides writing the actual code to map these errors, I've also encountered other problems:

[...]

Issue #308, opened 2016-09-15
value! - unable to infer enough type information about `_`


ArtemGr:
The following simplified code

```
named! (foo (&str) -> bool,
    chain! (
        tag_s! ("foo") ~
        b: alt_complete! (
            map! (tag_s! ("1"), |_: &str|->bool {true}) |
            value! (false)
        ),
        || b));
```

produces the following error:

```
error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [--explain E0282]
  --> <nom macros>:11:3
   |>
11 |> ( $ crate :: IResult :: Done ( $ i , $ res ) ) ;
   |>   ^^^^^^^^^^^^^^^^^^^^^^^^^^
<nom macros>:25:32: 32:43: note: in this expansion of value! (defined in <nom macros>)
<nom macros>:25:1: 25:54: note: in this expansion of alt_parser! (defined in <nom macros>)
<nom macros>:8:52: 9:23: note: in this expansion of alt_complete! (defined in <nom macros>)
src/main.rs:8:24: 141:42: note: in this expansion of alt_complete! (defined in <nom macros>)
<nom macros>:14:1: 17:39: note: in this expansion of chaining_parser! (defined in <nom macros>)
<nom macros>:2:3: 2:54: note: in this expansion of chaining_parser! (defined in <nom macros>)
src/main.rs:6:10: 5:34: note: in this expansion of chain! (defined in <nom macros>)
src/main.rs:5:1: 12:16: note: in this expansion of named! (defined in <nom macros>)
```

Here's the code in the playground: http://play.integer32.com/?gist=50f465df8a126e1a96fba226717d7a9b

---

The real use case is a little different. If you're curious:

```
named! (exploding_dice (&str) -> DiceOp, chain! (
  tag_s! ("!") ~ explode: alt_complete! (
    map! (pair! (alt! (tag_s! ("<") | tag_s! (">")), digit),
      |(le_ge, d): (&str, &str)| DiceOp::Explode {le_ge: if le_ge == ">" {1} else {-1}, face: Some (d.parse().expect ("!parse"))}) |
    map! (digit,
      |d: &str| {DiceOp::Explode {le_ge: 0, face: Some (d.parse().expect ("!parse"))}}) |
    value! (DiceOp::Explode {le_ge: 0, face: None})),
  || explode));
```

Geal:
[...]

PR #413, opened 2016-12-30
[WIP] Propagated error type.

### LINKED ISSUE #302: Using cond! together with count! unable to infer enough type information about `_`
Not sure if it's a bug in my code or in the library but if I'm trying to parse something using the following reduced example:

```
fn parse_branch_entries(input: &[u8]) -> IResult<&[u8], Option<Vec<u64>> > {
    chain!(input,
        entries: cond!(true, count!(le_u64, 3)),
        || { entries }
    )
}
```

it tells me:

```
error[E0282]: unable to infer enough type information about `_`
   --> src/parser/mod.rs:394:5
    |
394 |     chain!(input,
    |     ^ cannot infer type for `_`
    |
    = note: type annotations or generic parameter binding required
    = note: this error originates in a macro from the standard library
```

kdy1997:
Error types are implicitly passed around, and gives hints about type of error to rustc.

Fixes #302  #308 #394 

Some macros are currently broken. 
## Fix:
 
- [x] many1
- [x] many0
- [x] separated_list
- [x] separated_nonempty_list
- [ ] delimited
- [x] many_till
- [ ] and some more..

I need some help to fix delimited!(), what does tuple_parser!() do?
How can I get it's error value?

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -119,7 +119,7 @@ macro_rules! take_bits (
       use std::ops::Div;
       //println!("taking {} bits from {:?}", $count, $i);
       let (input, bit_offset) = $i;
-      let res : $crate::IResult<(&[u8],usize), $t> = if $count == 0 {
+      let res : $crate::IResult<(&[u8],usize), $t, _> = if $count == 0 {
         $crate::IResult::Done( (input, bit_offset), 0)
       } else {
         let cnt = ($count as usize + bit_offset).div(8);
@@ -169,14 +169,14 @@ macro_rules! tag_bits (
         $crate::IResult::Incomplete(i) => $crate::IResult::Incomplete(i),
         $crate::IResult::Done(i, o)    => {
           if let $p = o {
-            let res: $crate::IResult<(&[u8],usize),$t> = $crate::IResult::Done(i, o);
+            let res: $crate::IResult<(&[u8],usize),$t,_> = $crate::IResult::Done(i, o);
             res
           } else {
             $crate::IResult::Error(error_position!($crate::ErrorKind::TagBits, $i))
           }
         },
-        _                              => {
-          $crate::IResult::Error(error_position!($crate::ErrorKind::TagBits, $i))
+        $crate::IResult::Error(tag_bits_err) => {
+          $crate::propagate_error_type(tag_bits_err, $crate::IResult::Error(error_position!($crate::ErrorKind::TagBits, $i)))
         }
       }
     }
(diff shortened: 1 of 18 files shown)


Geal:
hi!

before going further, can you add tests to confirm that it fixes the issue? And explain how it works?

src/bits.rs
```diff
@@ -188,35 +188,45 @@ mod tests {
...
-    assert_eq!(take_bits!( (sl, 1), u8,   3 ), IResult::Done((&sl[0..], 4), 2));
-    assert_eq!(take_bits!( (sl, 6), u8,   3 ), IResult::Done((&sl[1..], 1), 5));
-    assert_eq!(take_bits!( (sl, 0), u16, 10 ), IResult::Done((&sl[1..], 2), 683));
-    assert_eq!(take_bits!( (sl, 0), u16,  8 ), IResult::Done((&sl[1..], 0), 170));
-    assert_eq!(take_bits!( (sl, 6), u16, 10 ), IResult::Done((&sl[2..], 0), 752));
-    assert_eq!(take_bits!( (sl, 6), u16, 11 ), IResult::Done((&sl[2..], 1), 1504));
-    assert_eq!(take_bits!( (sl, 0), u32, 20 ), IResult::Done((&sl[2..], 4), 700163));
-    assert_eq!(take_bits!( (sl, 4), u32, 20 ), IResult::Done((&sl[3..], 0), 716851));
-    assert_eq!(take_bits!( (sl, 4), u32, 22 ), IResult::Incomplete(Needed::Size(22)));
+    assert_eq!(take_bits!( (sl, 0), u8,   0 ), done((sl, 0), 0));
```
Geal: why do you replace `IResult::Done` with `done` here?

src/branch.rs
```diff
@@ -138,8 +138,8 @@ macro_rules! alt (
       match $subrule!( $i, $($args)* ) {
         $crate::IResult::Done(i,o)     => $crate::IResult::Done(i,$gen(o)),
         $crate::IResult::Incomplete(x) => $crate::IResult::Incomplete(x),
-        $crate::IResult::Error(_)      => {
-          alt!(__impl $i, $($rest)*)
+        $crate::IResult::Error(_alt_err)      => {
+          $crate::propagate_error_type(_alt_err, alt!(__impl $i, $($rest)*))
```
Geal: I'm a bit uneasy about this. Especially, I don't know how it will interact with the whitespace parsing combinators

src/traits.rs
```diff
@@ -17,6 +17,13 @@ pub trait InputLength {
   fn input_len(&self) -> usize;
 }
 
+impl<T> InputLength for [T] {
```
Geal: why this?

src/util.rs
```diff
@@ -235,7 +235,7 @@ pub fn slice_to_offsets(input: &[u8], s: &[u8]) -> (usize, usize) {
 
 #[cfg(not(feature = "core"))]
 #[cfg(feature = "verbose-errors")]
-pub fn prepare_errors<O,E: Clone>(input: &[u8], res: IResult<&[u8],O,E>) -> Option<Vec<(ErrorKind<E>, usize, usize)> > {
+pub fn prepare_errors<O,E: Clone>(input: &[u8], res: IResult<&[u8], O, E>) -> Option<Vec<(ErrorKind<E>, usize, usize)> > {
```
Geal: please avoid cosmetic changes in unrelated parts of the code

tests/issues.rs
```diff
@@ -98,7 +98,9 @@ fn issue_142(){
 
 #[test]
 fn usize_length_bytes_issue(){
-  length_bytes!(b"012346", be_u16);
+  named!(typed_iresult, length_bytes!(be_u16));
```
Geal: why touch that test?

tests/test1.rs
```diff
@@ -40,5 +40,6 @@ fn is_not() {
 #[test]
 fn exported_public_method_defined_by_macro() {
   let a = &b"ab12cd\nefgh"[..];
-  assert_eq!(not_line_ending(a), IResult::Done(&b"\nefgh"[..], &b"ab12cd"[..]));
+  let expected: IResult<_, _> = IResult::Done(&b"\nefgh"[..], &b"ab12cd"[..]);
```
Geal: why change that test?

kdy1997: default parameter does not works for function, so

return types for codes that calls macro directly from assert_eq() cannot be inferred.

```rust

```

In real world usecase, it would not affect any usage, unless users write code like

```rust
assert_eq!(macro_in_nom!("args.."), Done(...))
```

src/traits.rs
kdy1997: I'm not sure about it. Maybe crlf? I configured vscode to use lf always, but it somehow prefers crlf. I'll try to remove it from pull request.

src/bits.rs
kdy1997: As commented below, type cannot be inferred if macro is used directly from assert_eq!()

kdy1997:
I'm stucked while making escaped function. Does original macro accept only slices? And it seems to match against u8, but matching non-ascii character as byte isn't safe for utf8. UTF8 extended characters have 1 on it's first bit for ascii-compatibility, so it's safe to match as byte only if it's in [0, 127].

Geal:
ok, sorry for the delay, I will now get to work on this :)

escaped cannot work yet with &str, I planned to do that for 2.0, but didn't have the time to do it: https://github.com/Geal/nom/issues/300

Geal:
fixing inference for escaped will be difficult :/

Geal:
I really don't know how to fix that one, except by specifying the type directly like this: `named!(esc, escaped!(call!(alpha::<&[u8],u32>), '\\', one_of!("\"n\\")));`

I found a better way to write escaped in the process, though

Geal:
I have to stop for today. I have a feeling the solution will be in fixing the error type for basic parsers like alpha or tag, and let users do their fix_error above them, instead of trying to adapt in a parent parser to the custom error type coming from the child parser.

kdy1997:
I'm really sorry, but I can't help because I'm far away from my home and I can't use any pc.
I'm not sure when it would be allowed.. Maybe on march? But even this depends.


I agree that propagating error types from basical parsers is much important task.

Geal:
@kdy1997 why did you close this PR?

kdy1997:
Oh.. sorry for my mistake. I think I accidentally touched it while reading from mobile.

Geal:
ok, I'm relieved :)
I might have time to work on it next week

fflorent:
Any news on this?

Florent

Geal:
hey, so, there was a huge refactoring for nom 4 to fix this issue, so I don't think I'll be able to merge that PR in the end. Thank you for your work still!

PR #420, opened 2017-01-07
Add move_closure.

AndyShiue:
I guess this would probably be useful.


Geal:
in which case do you need it?

Geal:
Closing this for now, the closure combinator is not used too much in the end

PR #426, opened 2017-01-21
[wip] Add example to `one_of`

derekdreery:


derekdreery:
I'll add the str support to this PR, so closing #429

derekdreery:
Having looked at this some more, it seems sensible to break backwards compatibility, to make the signature like `is_not!`, @Geal do you think this is the correct way to do this?

Geal:
why would it break backward compatibility?

derekdreery:
hmm I didn't understand how general the new-style is, I'm still trying to understand what the macro will take and produce.

Geal:
could you try again, now that I pushed 59ef6901d012be8d804490 ?

Geal:
Thanks!

PR #429, opened 2017-01-23
Add str version of one_of

derekdreery:
Adds a string slice version of `one_of` called `one_of_s`. Slight difference to `one_of`, in that it returns a `char` rather than a `&str` (`one_of` returns a `&[u8]` with a single element).

Handles multibyte chars, as seen in doc_test.

### DIFF
--- a/src/str.rs
+++ b/src/str.rs
@@ -224,6 +224,38 @@ macro_rules! take_until_s (
   );
 );
 
+/// Matches one of the characters in a string
+///
+/// # Example
+/// ```
+/// # #[macro_use] extern crate nom;
+/// # use nom::IResult;
+/// # fn main() {
+/// let test = one_of_s!("汉 some more input", "汉012");
+/// assert_eq!(test, IResult::Done(" some more input", '汉'));
+/// # }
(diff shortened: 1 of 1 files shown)


Geal:
Hi!
thank you for working on this. Could you instead convert the current `one_of` to be independent from the input type like other parsers?

derekdreery:
I'll have a look.

derekdreery:
There are some issues around `str`, as index operations (e.g. `my_str[2..]`) work on the underlying bytes, and panic if the position doesn't correspond to a char boundary. That's why the patch uses the `chars` iterator to get at the first character. Do you know if it is possible to to work around this, in order to make the macro generic?

Geal:
You can use the traits implemented on input types: https://github.com/Geal/nom/blob/master/src/traits.rs#L141

see for example the implementation of `is_not!`: https://github.com/Geal/nom/blob/master/src/bytes.rs#L92-L116

the `position` method will give you the byte offset of a char for `&str`.

derekdreery:
I'll add this to the other PR.

PR #430, opened 2017-01-23
Added be_u24 method

valarauca:
In TLS1.0+ there are payload fields who's length is set via a 24bit unsigned integer value.

Kind of a strange method I agree. I'm not sure if you prefer `usize` or `u32`

### DIFF
--- a/src/nom.rs
+++ b/src/nom.rs
@@ -340,6 +340,17 @@ pub fn be_u16(i: &[u8]) -> IResult<&[u8], u16> {
   }
 }
 
+/// Recognizes big endian unsigned 3 byte integer
+#[inline]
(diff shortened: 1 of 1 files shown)


Geal:
u32 should be fine. Can you add the little endian method as well?

valarauca:
Where should I commit the tests btw?

Geal:
the tests should be in the same file with the others: https://github.com/Geal/nom/blob/master/src/nom.rs#L871-L956

Geal:
ping :)

valarauca:
Hey sorry in the middle of a move for work. I'll follow up on this when I
get the chance

Geal:
Thanks!

PR #439, opened 2017-02-09
Enable the str-related traits and modules when using the "core" feature

iBelieve:
These traits and modules all seem to work fine under no_std on nightly Rust.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,7 +105,7 @@ extern crate test;
 #[cfg(feature = "core")]
 mod std {
 #[macro_use]
-  pub use core::{fmt, iter, option, ops, slice, mem};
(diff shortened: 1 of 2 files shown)


iBelieve:
Hmm, looks like coveralls is crashing in the Travis CI job:

```
$ ./kcov/build/src/kcov --verify --coveralls-id=$TRAVIS_JOB_ID --exclude-pattern=/.cargo target/kcov target/debug/nom-*
kcov: warning: kcov: WARNING: kcov has been built without libbfd-dev (or
kcov: binutils-dev), so the --verify option will not do anything.
kcov: Process exited with signal 11 (SIGSEGV) at 0x55555571a92e
```

Geal:
merged, thanks!

PR #443, opened 2017-02-13
IResult: implement or() similar to Option.or()

Keruspe:
For now, implementing some alternative "whitespace" logic can be tricky as functions like take_until & co give you an error when i[0] doesn't satisfy the condition, and sep! then bubbles-up the error. opt! is not acceptable here as sep! doesn't like Option either.
With this, it's simple enough to implement a new alternative with stuff like:

pub fn skip_non_vowels (i: &[u8]) -> IResult<&[u8], &[u8]> {
is_not!(i, b"aeiouy").or(IResult::Done(i, &i[0..0]))
}

and then

sep!(i, skip_non_vowels, blah)

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -120,6 +120,14 @@ impl<I,O,E> IResult<I,O,E> {
     }
   }
 
+  pub fn or(self, other: IResult<I, O, E>) -> IResult<I, O, E> {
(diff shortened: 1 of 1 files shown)


Geal:
Thanks

PR #446, opened 2017-02-19
Added documentation for delimited macro

timlyo:
I've added a simple example for the delimited macro. Is this useful? I'll add a few more as I learn this library if it is.


Geal:
Thanks!

PR #447, opened 2017-02-21
hex_u32 should accept uppercase digits

JayKickliter:
Since `hex_digit` accepts `ABCDEF`, I see no reason why `hex_u32` shouldn't. This caused a real-world bug.


Geal:
ooh, that's bad. Sorry for this, and thank you for fixing it!

JayKickliter:
No problem. Thanks for the library, it's working out great.

Issue #437, opened 2017-02-07
float, float_s, double and double_s can not recognize numbers without dots


c410-f3r:
Current implementation does not recognize numbers like `123e-10` or `123`.
I am also closing #435 since this matter needs more discussion.

federicomenaquintero:
I just ran into this for librsvg.  I added some tests:

```
  #[test]
  fn float_test() {
    assert_eq!(float(&b""[..]),            Incomplete(Needed::Size(1)));

    assert_eq!(float(&b"0"[..]),           Done(&b""[..], 0.0));
    assert_eq!(float_s(&"0"[..]),          Done(&""[..], 0.0));

    assert_eq!(float(&b"-"[..]),           Incomplete(Needed::Unknown));
    assert_eq!(float_s(&"-"[..]),          Incomplete(Needed::Unknown));

    assert_eq!(float(&b"-0"[..]),          Done(&b""[..], 0.0));
    assert_eq!(float_s(&"-1"[..]),         Done(&""[..], -1.0));

    assert_eq!(float(&b"123"[..]),         Done(&b""[..], 123.0));
    assert_eq!(float_s(&"123"[..]),        Done(&""[..], 123.0));

    assert_eq!(float(&b"-123"[..]),        Done(&b""[..], -123.0));
    assert_eq!(float_s(&"-123"[..]),       Done(&""[..], -123.0));

    assert_eq!(float(&b"+3."[..]),         Done(&b""[..], 3.0));
    assert_eq!(float_s(&"+3."[..]),        Done(&""[..], 3.0));

    assert_eq!(float(&b"+3.14"[..]),       Done(&b""[..], 3.14));
    assert_eq!(float_s(&"3.14"[..]),       Done(&""[..], 3.14));

    assert_eq!(float(&b".5"[..]),          Done(&b""[..], 0.5));
    assert_eq!(float_s(&".5"[..]),         Done(&""[..], 0.5));

    assert_eq!(float(&b"-.5"[..]),         Done(&b""[..], -0.5));
    assert_eq!(float_s(&"-.5"[..]),        Done(&""[..], -0.5));

    assert_eq!(float(&b"-1.234E12"[..]),   Done(&b""[..], -1.234E12));
    assert_eq!(float_s(&"-1.234E12"[..]),  Done(&""[..], -1.234E12));

    assert_eq!(float(&b"-1.234E-12"[..]),  Done(&b""[..], -1.234E-12));
[...]

PR #448, opened 2017-02-22
float() / float_s() / double() / double_s(): Accept numbers without decimals

federicomenaquintero:
Fixes https://github.com/Geal/nom/issues/437

We add more detailed tests, and add support for floating-point numbers
that are written without decimals.

### DIFF
--- a/src/nom.rs
+++ b/src/nom.rs
@@ -602,9 +602,10 @@ pub fn float(input: &[u8]) -> IResult<&[u8],f32> {
     recognize!(
       tuple!(
         opt!(alt!(tag!("+") | tag!("-"))),
-        alt!(
-          delimited!(digit, tag!("."), opt!(digit))
-          | delimited!(opt!(digit), tag!("."), digit)
+        alt_complete!(
+          delimited!(opt!(digit), tag!("."), digit)
+          | preceded!(digit, tag!("."))
+          | digit
         ),
         opt!(complete!(tuple!(
           alt!(tag!("e") | tag!("E")),
@@ -625,9 +626,10 @@ pub fn float_s(input: &str) -> IResult<&str,f32> {
     recognize!(
       tuple!(
         opt!(alt!(tag!("+") | tag!("-"))),
-        alt!(
-          delimited!(digit, tag!("."), opt!(digit))
-          | delimited!(opt!(digit), tag!("."), digit)
+        alt_complete!(
+          delimited!(opt!(digit), tag!("."), digit)
+          | preceded!(digit, tag!("."))
+          | digit
         ),
         opt!(complete!(tuple!(
           alt!(tag!("e") | tag!("E")),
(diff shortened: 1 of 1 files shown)


src/nom.rs
```diff
@@ -602,9 +602,10 @@ pub fn float(input: &[u8]) -> IResult<&[u8],f32> {
     recognize!(
       tuple!(
         opt!(alt!(tag!("+") | tag!("-"))),
-        alt!(
-          delimited!(digit, tag!("."), opt!(digit))
-          | delimited!(opt!(digit), tag!("."), digit)
+        alt_complete!(
+          delimited!(opt!(digit), tag!("."), digit)
+          | preceded!(digit, tag!("."))
+          | digit
```
sdroege: That's basically the same as https://github.com/Geal/nom/blob/master/benches/json.rs#L27 now, just as a general comment. Change looks good to me!

src/nom.rs
```diff
@@ -602,9 +602,10 @@ pub fn float(input: &[u8]) -> IResult<&[u8],f32> {
     recognize!(
       tuple!(
         opt!(alt!(tag!("+") | tag!("-"))),
-        alt!(
-          delimited!(digit, tag!("."), opt!(digit))
-          | delimited!(opt!(digit), tag!("."), digit)
+        alt_complete!(
```
Geal: using `alt_complete!` can break the behaviour of the parser. If the data is`123.4` but you get it as `123` then `.4` (because of buffering, network packets or other reasons), you would parse it as `123` then a parser that fails on the unexpected `.4`.

I used it in benches/json.rs for convenience, but in the general case it does not work :/

federicomenaquintero: What's the strategy to handle that case?  How would one say, "I need one more character to see if it is part of the number, or if the production ends here"?

joshtriplett: Other streaming parsers I've seen, such as HTTP parsers, handle cases like this by returning an "I need more data" indication to the caller (ideally with some idea of "how much more data"), which has to obtain more data and call back into the parser.

federicomenaquintero: Nom already does that with IResult::Incomplete.  But in this case, the buffer ends and the parser is at a state where it *could* be finished: think of reading "123.45" at the end of the buffer, but the next characters (which are unavailable yet) are "67".

digit() already has a check for input_length == 0 when it starts right at the end of the buffer.  Maybe if it runs out of buffer in the for loop it should also return Incomplete?

(What would happen after that, both for the "here's some more data" case and for the "nope, that's the only buffer we'll have" case?)

federicomenaquintero: Are there tests for streaming?  Something like a test_parser (parser, "some data") that tests the parser with a single buffer with the whole string, and with a streaming wrapper that feeds it 1 byte at a time?

sdroege: Probably something with eof. Your parser wants more digits until eof or a non-digit is found

Geal: ok, so, anyone want to finish this PR? :)

marshallpierce:
I was just bitten by this problem; is it still in need of someone to take it over?

src/nom.rs
richardwhiuk: Should the parser differentiate between `IncompleteOrDone` or `IncompleteOrError` - which would allow the `complete!` macro to resolve it?

Geal:
I think we can close this now, with the code that landed in 99e4cea

PR #449, opened 2017-02-24
Use the derive named 'deprecated' for chain!

Kerollmops:


Geal:
ooooh this is nice! I didn't know we could do that now :)

PR #458, opened 2017-03-07
Support &str input for escape* 

### LINKED ISSUE #300: escaped_*! don't support &str

tmccombs:
Fixes #300.

Unfortunately, working on this I realized that `escaped_transform` only works with verbose-errors, because without it the type inferrer can't determine the E type of IResult in some places. I tried to fix that as well, but couldn't find a good solution. I also noticed that there are several unit tests that fail with the `verbose-errors` feature on. I fixed the tests related to the `escaped*` macros.


Geal:
Hey!
I'm sorry it took so much time to merge this PR. It will appear in the next major release of nom, 4.0 :)

PR #461, opened 2017-03-11
perf(tests/ini{,_str}): remove unnecessary cloning and refactor

utkarshkukreti:
I was benchmarking this parser against one [I wrote in munch.rs](https://github.com/utkarshkukreti/munch.rs/compare/dcc1f29a1d858d684c61e116169bf1f2f610e073...84356c91d5974249b84c4149f46989dd85a3e46b) and found some low hanging fruits to speed up this one, specifically removing the unnecessary cloning of the `HashMap` returned by `keys_and_values`.

(If you want, I can also add the benches to these examples. I [already have them written](https://github.com/utkarshkukreti/nom/commit/5c3376daa2e1f8409590ea1cac55d4bc3d0e4436) but I didn't know whether you want benches for these in the repo.)

### DIFF
--- a/tests/ini.rs
+++ b/tests/ini.rs
@@ -40,14 +40,9 @@ named!(key_value    <&[u8],(&str,&str)>,
 named!(keys_and_values_aggregator<&[u8], Vec<(&str,&str)> >, many0!(key_value));
 
 fn keys_and_values(input:&[u8]) -> IResult<&[u8], HashMap<&str, &str> > {
-  let mut h: HashMap<&str, &str> = HashMap::new();
-
   match keys_and_values_aggregator(input) {
     IResult::Done(i,tuple_vec) => {
-      for &(k,v) in &tuple_vec {
-        h.insert(k, v);
-      }
-      IResult::Done(i, h)
+      IResult::Done(i, tuple_vec.into_iter().collect())
     },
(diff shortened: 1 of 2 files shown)


Geal:
Thanks! That parser was not built with performance in mind, but this also makes the code simpler. Can you add the benchmarks to nom too? I was thinking of adding more benchmarks to track if there were performance regressions.

Geal:
BTW, how did the parser libraries fare against each other? :)

utkarshkukreti:
I'll try to send a PR with the benchmarks as soon as possible. The way I did it in the tmp branch I linked to above will make the tests not run on stable. If I add it in a new file in `benches`, I cannot load `tests/ini.rs` as a module with custom path because it includes `macro_use` with `extern crate` which is only allowed at crate root:

```
error[E0468]: an `extern crate` loading macros must be at the crate root
 --> benches/../tests/ini.rs:3:1
  |
3 | extern crate nom;
  | ^^^^^^^^^^^^^^^^^
```
Do you have any suggestions for this?

<hr>

For the benchmark I linked above, I'm getting around 96 MB/s for `ini_str` + nom and 140 MB/s for munch. I don't know nom well enough to say why it's slower.

utkarshkukreti:
Fixed the compilation errors and opened #471!

PR #464, opened 2017-03-15
Make inner errors in many_till! visible (fix #463)

### LINKED ISSUE #463: Getting errors through take_while! or similar loop
I have something that looks like this:

    many_till!(complex_subparser,
               simple_end_marker)

What I was hoping to get is something like the following pseudocode:

    let mut result = vec![];
    while(simple_end_marker is not matching) {
        result.append(try!(complex_subparser))
    }
    result

That is, as long as the simple_end_marker does not match, I expect the complex_subparser to match. If it doesn't, I want the error from it.  Currently, if both the simple_end_marker and the complex_subparser fails, all I get is a `ManyTill` error.

I'm using the `verbose-errors` feature.

My actual code, for context: https://github.com/kaj/ructe/blob/feature/improve-error-handling/src/template.rs#L63

kaj:
Make `many_till!()` return a NodePosition error on failure, so that the inner error that caused it to fail is visible to the caller.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -297,7 +297,7 @@ macro_rules! many1(
 ///    let res_b: (Vec<&[u8]>, &[u8]) = (Vec::new(), &b"efgh"[..]);
 ///    assert_eq!(multi(&a[..]), Done(&b"abcd"[..], res_a));
 ///    assert_eq!(multi(&b[..]), Done(&b"abcd"[..], res_b));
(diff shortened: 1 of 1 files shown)


kaj:
This changes the `many_till!()` macro in a way that may break existing error handling in crates using nom.  Maybe a separate version should be used instead?  If so, what should it be called?

Geal:
Thanks!

PR #466, opened 2017-03-16
switch: wrap cases in a Some()

### LINKED ISSUE #460: support exhaustive matching in switch!
Currently, nom always adds a catchall "_" to the generated code.
Either having another macro like switch_exhaustive or not dealing with the catchall would be great

Keruspe:
This changes basically nothing except that _ becomes a valid case even with an exhaustive listing as it would match None (even if that case can never happen)

Fixes #460


Geal:
Thanks! Can you add something in the doc about exhaustive matches?

PR #467, opened 2017-03-19
Fix documentation for take_while1.

tmccombs:
Documentation says the predicate takes a `&[T]` when in actuality it takes a `T` (as one would expect).


Geal:
Good catch, thanks!

PR #470, opened 2017-03-21
Fixed markdown in README.md

denisbider:
Fixed inconsistent begin/end markers for code segment that caused it to not be terminated.


Geal:
Thanks!

PR #471, opened 2017-03-21
Add a simple benchmark to both ini and ini_str parsers.

utkarshkukreti:
These can be run on Rust Nightly using:

    cargo bench --features nightly --test ini

and

    cargo bench --features nightly --test ini_str

On my machine, these currently run at around 99 MB/s and 101 MB/s
respectively (best of many runs).

<hr>

Just for reference, a similar benchmark in munch.rs [[1]] [[2]] runs at 140 MB/s on the same machine as of the current HEAD (149 MB/s with an optimization I'm testing right now and will push soon) as mentioned in the original PR: https://github.com/Geal/nom/pull/461#issuecomment-287717125.

<hr>

@Geal let me know if I got the `cfg`s right!

[1]: https://github.com/utkarshkukreti/munch.rs/blob/62907a57d0d598bd5570c4726bbd0251bd66152f/benches/ini.rs
[2]: https://github.com/utkarshkukreti/munch.rs/blob/62907a57d0d598bd5570c4726bbd0251bd66152f/examples/ini.rs

### DIFF
--- a/tests/ini.rs
+++ b/tests/ini.rs
@@ -1,3 +1,7 @@
+#![cfg_attr(feature = "nightly", feature(test))]
+
+#[cfg(feature = "nightly")]
+extern crate test;
 
 #[macro_use]
 extern crate nom;
@@ -221,4 +225,21 @@ key4 = value4
   expected_h.insert("abcd",     expected_1);
   expected_h.insert("category", expected_2);
   assert_eq!(res, IResult::Done(ini_after_parser, expected_h));
+}
+
+#[cfg(feature = "nightly")]
+#[bench]
+fn bench_ini(b: &mut test::Bencher) {
+  let str = "[owner]
+name=John Doe
+organization=Acme Widgets Inc.
+
+[database]
+server=192.0.2.62
+port=143
+file=payroll.dat
+";
+
+  b.iter(|| categories(str.as_bytes()).unwrap());
+  b.bytes = str.len() as u64;
 }
--- a/tests/ini_str.rs
+++ b/tests/ini_str.rs
@@ -1,3 +1,7 @@
+#![cfg_attr(feature = "nightly", feature(test))]
+
+#[cfg(feature = "nightly")]
+extern crate test;
 
 #[macro_use]
 extern crate nom;
@@ -238,4 +242,21 @@ key4 = value4
   expected_h.insert("abcd",     expected_1);
   expected_h.insert("category", expected_2);
   assert_eq!(res, IResult::Done("", expected_h));
+}
+
+#[cfg(feature = "nightly")]
+#[bench]
+fn bench_ini(b: &mut test::Bencher) {
+  let str = "[owner]
+name=John Doe
+organization=Acme Widgets Inc.
+
+[database]
+server=192.0.2.62
+port=143
+file=payroll.dat
+";
+
+  b.iter(|| categories(str).unwrap());
+  b.bytes = str.len() as u64;
 }


Geal:
Thanks!

Geal:
so, I took a look, and there were a few points where the original INI code could be simplified. And now the nom benchmark is on par with the munch one:
- munch: 849 ns/iter (+/- 153) = 129 MB/s
- nom: 788 ns/iter (+/- 81) = 138 MB/s

(sometimes faster than munch, sometimes slower, there's some variance)

The part that I find interesting is there: https://github.com/utkarshkukreti/munch.rs/blob/62907a57d0d598bd5570c4726bbd0251bd66152f/examples/ini.rs#L26

Since you use iterators, it appears rustc is smart enough to do the collect in the same loop as the rest, while the nom code has to first produce a Vec, then collect it into a HashMap. So it is a bit frustrating for me, since nom cannot get that specific performance boost.

Maybe a collect operator could be useful, what do you think?

Anyway, good job on munch :)

Geal:
By the way, how do you handle the `Incomplete` case? When the data is valid but there is not enough?

utkarshkukreti:
Nice! I'm getting a fairly consistent result of 135 MB/s for nom now and 140 MB/s for munch.

> Since you use iterators, it appears rustc is smart enough to do the collect in the same loop as the rest, while the nom code has to first produce a Vec, then collect it into a HashMap. So it is a bit frustrating for me, since nom cannot get that specific performance boost.

Nope, I'm not using iterators there. `repeat` _is_ generating a temporary Vec right now and the `.map` is run after that. munch does have a `fold` that can be used to collect into anything directly. With the following change I'm getting a nice almost 20% speed up -- 165 MB/s:

```diff
diff --git a/examples/ini.rs b/examples/ini.rs
index a068c4c..21d7fe7 100644
--- a/examples/ini.rs
+++ b/examples/ini.rs
@@ -20,10 +20,16 @@ pub fn parse(str: &str) -> Result<Value, (usize, munch::str::Error<'static>)> {
     let value = P(TakeWhile(|ch| ch != '\n' && ch != ';'));
 
     let kv = (key << s() << '=' << s(), value << s() << Optional(comment) << ws());
-    let kvs = kv.repeat(..).map(|vec| vec.into_iter().collect());
+    let kvs = kv.repeat(..).fold(|| HashMap::new(), |mut hm, (k, v)| {
+        hm.insert(k, v);
+        hm
+    });
 
     let section = (header, kvs);
-    let sections = section.repeat(..).map(|vec| vec.into_iter().collect());
+    let sections = section.repeat(..).fold(|| HashMap::new(), |mut hm, (k, v)| {
+        hm.insert(k, v);
+        hm
+    });
 
     match (ws() >> sections << End).parse(str, 0) {
         Ok((_, output)) => Ok(output),
```

I'm not sure if I'll make `repeat` generic w.r.t. the container in munch or just let the user use `fold` for these cases.

> Anyway, good job on munch :)

Thanks. :)

utkarshkukreti:
> By the way, how do you handle the `Incomplete` case? When the data is valid but there is not enough?

I do not. I think some (all?) of these can be inferred afterwards based on the exact error variant, the position of the failure, and the input data. For example, if the error is `TakeWhile1` and the position of error == input.len(), then it can be considered Incomplete, otherwise it's an actual error. Not sure if this will work through the combinators like `Repeat` though. I haven't thought much about this yet since I don't need this feature for the projects I'm going to use munch for.

Geal:
It's a bit more complex than that :)
Like, for many0, the error and incomplete cases have different results: error means returning the vec and continue parsing, while incomplete means return incomplete and stop parsing.
It's a crucial feature for me, to write streaming parsers, but it has annoyed a lot of users who only work on complete inputs. And I guess it has a performance hit as well, since I also calculate how much data is needed as it bubbles up the parser tree.

utkarshkukreti:
I did add a consumed vs not consumed distinction so that `.repeat(..)` doesn't backtrack if the internal parser consumed some bytes and then returned an error. The unit tests for this are not readable at all so here's a better example:

```rust
extern crate munch;

use munch::*;
use munch::str::*;

fn main() {
    let mut p = (('a', 'b').repeat(..), Any);
    for str in &["", "a", "ab", "aba", "abab", "abc", "abcd"] {
        println!("{:?} {:?}", str, p.parse(str, 0));
    }
}
```

```
"" Err((0, Any))
"a" Err((1, Char('b')))
"ab" Err((2, Any))
"aba" Err((3, Char('b')))
"abab" Err((4, Any))
"abc" Ok((3, ([('a', 'b')], 'c')))
"abcd" Ok((3, ([('a', 'b')], 'c')))
```

"aba" does not return Ok here even though it could because the parser consumed a part of the input before returning an error. I see `many0` does a similar thing where `Error` is ignored while `Incomplete` causes `many0` to return `Incomplete`.

There's a `Try` in munch which resets the input position on any failure, similar to `complete!` in nom as far as I can see.

```
    let mut p = ((Try(('a', 'b'))).repeat(..), Any);
```

```
"" Err((0, Any))
"a" Ok((1, ([], 'a')))
"ab" Err((2, Any))
"aba" Ok((3, ([('a', 'b')], 'a')))
"abab" Err((4, Any))
"abc" Ok((3, ([('a', 'b')], 'c')))
"abcd" Ok((3, ([('a', 'b')], 'c')))
```

utkarshkukreti:
> It's a crucial feature for me, to write streaming parsers

Do you have any simple but still somewhat useful example of a streaming parser built with nom? I'd like to see if munch can be used for that, even if it takes more code.

Geal:
See https://github.com/Geal/flavors

It's an FLV parser

Le 23 mars 2017 16:44, "Utkarsh Kukreti" <notifications@github.com> a
écrit :

PR #477, opened 2017-03-25
Fix Markdown Syntax

michelboaventura:
This error is here since 2015 but for some reason showed up on github's markdown parser just now.


michelboaventura:
Sorry, just now I realize there is two other OR fixing the same issue. I'm closing this one.

PR #478, opened 2017-03-26
Fix dead links in doc/home.md

BurningMind:
Modify links from the erroneous ".html" extension to the correct ".md" extension


Geal:
hi!
thanks for the PR, but it's not an error: the links point to html files because they're meant for generation by https://github.com/geal/cargo-external-doc

BurningMind:
Woops! Sorry about that

PR #475, opened 2017-03-23
Remove extraneous backtick in README.md

cpick:
Fixes GitHub's rendering of the trailing markdown.

PR #479, opened 2017-03-26
Remove backtick to fix structure of README.md

hansott:


hansott:
See https://github.com/Geal/nom/pull/475 & https://github.com/Geal/nom/pull/470

PR #480, opened 2017-03-26
Fixed formatting error with codeblock

Karl-D-Asmussen:
Was gosh-darn near unreadable!

Anyway, now the readme is nice again.

PR #483, opened 2017-03-27
Remove extraneous backtick from README.md

ameliaquining:
A code fence contained four backticks. This caused formatting problems.


chengsun:
Duplicate of #480

Geal:
Fixed by #470

Issue #474, opened 2017-03-23
alt! sometimes fails to compile with 


tmccombs:
If I try to compile `named!(endline<char>, alt!(char!('\n') | value!('\0', eof!())));`  I get:

```
error[E0282]: unable to infer enough type information about `E`
  --> src/desktop/parser.rs:42:1
   |
42 | named!(endline<char>, alt!(char!('\n') | value!('\0', eof!())));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for `E`
   |
   = note: type annotations or generic parameter binding required
   = note: this error originates in a macro outside of the current crate
```

however,  `named!(endline<char>, alt!(value!('\0', eof!()) | char!('\n')));` works fine.

I'm not entirely sure what causes this, or how to fix it.

PR #484, opened 2017-03-27
Simplification and fixes to the `alt!` macro

chengsun:
Two changes to the `alt!` macro:

1. **Code simplification**: Simplify `alt!` macro by using explicit terminator token.  
`alt!(__impl $i, ...)` always terminates with an explicit `__end` token now.
This saves us from having to duplicate each macro clause twice, as each
non-terminating pattern now always takes the form `pattern | rest`.

2. **Fix bug #474**: Make `alt!` unify generic type `E` to prevent underspecification.  
Add a compiler-time hack to ensure that the `E` type of a subexpression is
not underspecified, leading to the error "unable to infer enough type
information about `E`".


chengsun:
I'm still new to Rust by the way, let me know if there's a better way of implementing either commit.

dignati:
Any update on this? I just run into this (with too much code to just paste in here). Reversing the order of the statements inside `alt!` as in #474 did not work for me. Is there any workaround until this is fixed?

Geal:
about the type inference error, I'm working on it, this will go in the next nom release

src/branch.rs
```diff
@@ -133,13 +133,22 @@ macro_rules! alt (
...
         $crate::IResult::Done(i,o)     => $crate::IResult::Done(i,$gen(o)),
         $crate::IResult::Incomplete(x) => $crate::IResult::Incomplete(x),
-        $crate::IResult::Error(_)      => {
-          alt!(__impl $i, $($rest)*)
+        $crate::IResult::Error(e)      => {
+          let out = alt!(__impl $i, $($rest)*);
+
+          // Compile-time hack to ensure that res's E type is not under-specified.
+          // This all has no effect at runtime.
+          fn unify_types<T>(_: &T, _: &T) {}
```
Geal: how does this affect compilation time?

Geal: maybe the `unify_type` function could be defined elsewhere once, instead of in every recursive call of alt

src/branch.rs
```diff
@@ -149,45 +158,13 @@ macro_rules! alt (
...
-        $crate::IResult::Incomplete(x) => $crate::IResult::Incomplete(x),
-        $crate::IResult::Error(_)      => {
-          alt!(__impl $i)
-        }
-      }
-    }
-  );
-
-  (__impl $i:expr) => (
+  (__impl $i:expr, __end) => (
```
Geal: great trick to simplify alt, thanks!

PR #486, opened 2017-03-27
Check if requested length would overflow in do_parse, chain and tuple

### LINKED ISSUE #481: Panic in do_parse with take
Hi,

The following parser:
```rust
named!(parser02<&[u8],()>,
    do_parse!(
        hdr: take!(1) >>
        data: take!(18446744073709551615) >>
        ( () )
    )
);

#[test]
fn test02() {
    let data = b"0\x88\xff\xff\xff\xff\xff\xff\xff\xff00\x0f\x02\x000\x00\x00\x00\x00\x00\x0000\x0f\x00\xff\x0a\xbb\xff";
    let _ = parser02(data);
}
```

fails with an error:
```
thread 'test02' panicked at 'attempt to add with overflow', <do_parse macros>:33
```

The backtrace is not interesting, so not reported here.

The bug was originally discovered in https://github.com/rusticata/x509-parser/issues/1

chifflier:
This fixes #481 by using `overflowing_add`, in case the requested value would overflow when computing `Needed(consumed + needed)`.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -725,7 +725,11 @@ macro_rules! length_bytes(
           let nb = nb as usize;
           if i1.input_len() < nb {
             use $crate::Offset;
-            $crate::IResult::Incomplete($crate::Needed::Size($i.offset(i1) + nb))
+            let (size,overflowed) = $i.offset(i1).overflowing_add(nb);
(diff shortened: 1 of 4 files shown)


Geal:
Thanks!

PR #487, opened 2017-04-02
Indent parser implementations wrt the category

Victor-Savu:
The list of examples in the "Parsers written with nom" section is not properly indented, so the libraries appear in the same level as their categories. Increase the indent of the libraries to appear as sublists of the respective categories.


Geal:
Thanks!

PR #489, opened 2017-04-07
Return Incomplete from separated{,_nonempty}_list

cpick:
Previously, if the input ran out in the middle of a separator or
non-first element, `separated_list` and `separated_nonempty_list` would
return `Done` with the already parsed elements.

Imagine a scenario where the full input is "(a,a)", but only "(a," has
been loaded so far.

When attempting to parse with:
`delimited!(
    tag!("("),
    separated_list!(tag!(","), tag!("a")),
    tag!(")")
)`

The inner `separated_list` would return
`Done(&b","[..], vec![&b"a"[..]])` and the outer `delimited` would fail
when it saw "," instead of the expected ")".

With this patch, inner `separated_list` returns `Incomplete(_)` allowing
the caller to supply more input.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -1,7 +1,7 @@
 //! Parsers for applying parsers multiple times
 
 /// `separated_list!(I -> IResult<I,T>, I -> IResult<I,O>) => I -> IResult<I, Vec<O>>`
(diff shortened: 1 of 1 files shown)


Geal:
Thanks! good catch :)

PR #490, opened 2017-04-07
Return Incomplete from escaped

cpick:
Previously, if the input ran out while looking for an escaped value, `escaped`
would return `Error`.

Returning `Incomplete` allows the caller to supply more input.


Geal:
Thanks!

PR #495, opened 2017-04-13
Implement `unwrap_or` for IResult

evuez:
Same as `unwrap_or` for `std::option::Option`.

https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or


Geal:
Thanks!

PR #497, opened 2017-04-19
Only mark FileProducer as Eof on full buffer

cpick:
Previously, if `FileProducer.refill()` was called twice in a row the second
would read 0 additional bytes and assume that was due to Eof when really there
was just no more room left in the buffer.

When `FileProducer.apply()` returns `ConsumerState::Continue(Move::Await(_))`
it could be due to either 1) the buffer needing a refill or 2) the buffer not
being large enough.  There's no way for the caller to differentiate between
these states.

A sensisible plan may be to try to `apply()` again, in the case of (1) it will
`refill()` the buffer and continue on it's way.  However, if it's (2) then the
buffer will already be full `refill()` will read 0 bytes and erroneously mark
the producer at Eof.  There is no way to reset or recover from this condition.

Now that `refill()` has been fixed, the caller can respond to subsequent
`Move::Await(_)`s by `resize()`ing the producer's buffer and trying again.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -261,7 +261,15 @@ impl FileProducer {
     shift(&mut self.v, self.start, self.end);
     self.end = self.end - self.start;
     self.start = 0;
-    match self.file.read(&mut self.v[self.end..]) {
+
+    let remaining = &mut self.v[self.end..];
+
+    // already full, prevents erroneous Eof below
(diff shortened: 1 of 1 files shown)


Geal:
Thanks! The producers and consumers should probably be refactored and moved to a separate crate. They're a bit awkward to use right now

PR #501, opened 2017-05-02
Remove AsChar requirement for Item in InputIter

0xd34d10cc:
to avoid boilerplate code when parsing non-bytes and non-string input (like [here](https://github.com/Rydgel/monkey-rust/blob/master/lib/lexer/token.rs#L64-L126)).


Geal:
Thanks! Handling token streams is a great use of nom :)

PR #506, opened 2017-05-10
Simplify type declaration and usage

Hywan:
Just a simplification for type declaration and simplification. No deep change, just ease the maintenance and the reading. Patch is straightforward.


Geal:
Thanks, it's indeed easier to read

PR #507, opened 2017-05-10
Use `memchr` in `FindToken` and `FindSubstring` implementation

Hywan:
As said on Gitter, this is the first step for a `memchr` usage in the `FindToken` implementation. So far, no performance boost has been observed. Maybe benchmarks are not designed to show this kind of performance change yet.

This is a WIP.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -9,6 +9,8 @@ use std::str::CharIndices;
 use std::str::FromStr;
 use std::str::from_utf8;
 
+use memchr::memchr;
+
 
 /// abstract method to calculate the input length
 pub trait InputLength {
@@ -404,10 +406,7 @@ pub trait FindToken<T> {
 
 impl<'a> FindToken<&'a[u8]> for u8 {
   fn find_token(&self, input: &[u8]) -> bool {
-    for &i in input.iter() {
-      if *self == i { return true }
-    }
-    false
+    memchr(*self, input).is_some()
   }
 }
 
@@ -419,10 +418,7 @@ impl<'a> FindToken<&'a str> for u8 {
 
 impl<'a,'b> FindToken<&'a[u8]> for &'b u8 {
   fn find_token(&self, input: &[u8]) -> bool {
-    for &i in input.iter() {
-      if **self == i { return true }
-    }
-    false
+    memchr(**self, input).is_some()
   }
 }
 
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -9,7 +9,7 @@ use std::str::CharIndices;
 use std::str::FromStr;
 use std::str::from_utf8;
 
-use memchr::memchr;
+use memchr;
 
 
 /// abstract method to calculate the input length
@@ -406,7 +406,7 @@ pub trait FindToken<T> {
 
 impl<'a> FindToken<&'a[u8]> for u8 {
   fn find_token(&self, input: &[u8]) -> bool {
-    memchr(*self, input).is_some()
+    memchr::memchr(*self, input).is_some()
   }
 }
 
@@ -418,7 +418,7 @@ impl<'a> FindToken<&'a str> for u8 {
 
 impl<'a,'b> FindToken<&'a[u8]> for &'b u8 {
   fn find_token(&self, input: &[u8]) -> bool {
-    memchr(**self, input).is_some()
+    memchr::memchr(**self, input).is_some()
   }
 }
 
@@ -447,15 +447,31 @@ impl<'a,'b> FindSubstring<&'b [u8]> for &'a[u8] {
     let substr_len = substr.len();
 
     if substr_len == 0 {
-        return None;
-    }
+      None
+    } else if substr_len == 1 {
+      memchr::memchr(substr[0], self)
+    } else {
+      let max = self.len() - substr_len;
+      let mut offset = 0;
+      let mut haystack = &self[..];
 
-    for (index,win) in self.windows(substr_len).enumerate() {
-      if win == substr {
-        return Some(index)
+      while let Some(position) = memchr::memchr(substr[0], haystack) {
+        offset += position;
+
+        if offset > max {
+          return None
+        }
+
+        if &haystack[position..position + substr_len] == substr {
+          return Some(offset)
+        }
+
+        haystack  = &haystack[position + 1..];
+        offset   += 1;
       }
+
+      None
     }
-    None
   }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,9 @@ optional = true
 version = "^0.2.2"
 optional = true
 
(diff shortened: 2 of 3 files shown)


Geal:
could you rebase the tweet? I added a new benchmark yesterday for the arithmetic parser. Its slow, but we can probably see improvements.
Does memchr work in no_std?

Hywan:
https://github.com/BurntSushi/rust-memchr#no_std

Geal:
nice :)

Hywan:
`FindSubstring` is only used in the `take_until_and_consume!` macro so far. I used it to parse block comments (`preceded!(tag!("/*"), take_until_and_consume!("*/"))`). Here are my results so far:

```
# before optimisation:
test … ... bench:      73,433 ns/iter (+/- 3,869)
# after optimisation:
test … ... bench:      15,986 ns/iter (+/- 3,068)
```

(applied on a comment with 1024 paragraphs to see the difference).

I reckon we can do better with the `regex` crate for the fallback (but that might not obvious). Why? Because I hope they have better strategies than us to find a substring :-].

Geal:
it looks like the benchmark comparison is worse https://travis-ci.org/Geal/nom/jobs/232355649#L2637-L2643 could you check why?

Hywan:
I don't have such a difference on my machine:

```
 name                       control ns/iter   variable ns/iter  diff ns/iter  diff % 
 bench_ini                  750 (145 MB/s)    737 (147 MB/s)             -13  -1.73% 
 bench_ini_key_value        39 (461 MB/s)     40 (450 MB/s)                1   2.56% 
 bench_ini_keys_and_values  61 (721 MB/s)     62 (709 MB/s)                1   1.64% 
 bench_ini_str              1,012 (107 MB/s)  1,043 (104 MB/s)            31   3.06% 
 json_bench                 1,486             1,547                       61   4.10% 
 one_test                   512               484                        -28  -5.47% 
```

What's the expected threshold?

Hywan:
With another bench:

```
 name                       control ns/iter   variable ns/iter  diff ns/iter  diff % 
 bench_ini                  750 (145 MB/s)    727 (149 MB/s)             -23  -3.07% 
 bench_ini_key_value        39 (461 MB/s)     39 (461 MB/s)                0   0.00% 
 bench_ini_keys_and_values  61 (721 MB/s)     60 (733 MB/s)               -1  -1.64% 
 bench_ini_str              1,012 (107 MB/s)  1,021 (106 MB/s)             9   0.89% 
 json_bench                 1,486             1,542                       56   3.77% 
 one_test                   512               495                        -17  -3.32% 
```

There is almost no difference.

Geal:
It can also be Travis being unreliable. I'll test on my machine too

Geal:
I see no sensible differences. I should probably make benchmarks with larger parsers and more data, since it's too easy to be unstable at the nanosecond level

Hywan:
`memchr2` does not behave as I would like, please wait.

Hywan:
`memchr2` does not search for two consecutive needles, it just searches for two needles. This patch is broken.

Hywan:
Fixed!

Hywan:
Tests are green, coverall are red. Next steps?

Geal:
Thanks!

src/traits.rs
```diff
@@ -448,12 +444,34 @@ pub trait FindSubstring<T> {
...
+    if substr_len == 0 {
+      None
+    } else if substr_len == 1 {
+      memchr::memchr(substr[0], self)
+    } else {
+      let max = self.len() - substr_len;
+      let mut offset = 0;
+      let mut haystack = &self[..];
+
+      while let Some(position) = memchr::memchr(substr[0], haystack) {
```
fflorent: Something I learned today: memchr provides a iterator which maintains the state of the remaining haystack. See:
https://docs.rs/memchr/1.0.1/memchr/struct.Memchr.html

Some of the logic here could be removed in favor to this iterator if I am not wrong :).

fflorent: The alternative implementation would have been:
````rust
impl<'a,'b> FindSubstring<&'b [u8]> for &'a[u8] {
  fn find_substring(&self, substr: &'b[u8]) -> Option<usize> {
    let substr_len = substr.len();

    if substr_len == 0 {
      None
    } else if substr_len == 1 {
      memchr::memchr(substr[0], self)
    } else {
      let max = self.len() - substr_len;
      let haystack = &self[..max + 1];

      memchr::Memchr::new(substr[0], haystack)
          .find(|offset| &self[(*offset)..(offset+substr_len)] == substr)
    }
  }
}
````

But it doesn't pass the tests currently. Some error I don't identify yet occurs.

Not sure also how to benchmark. What do you think of it?

Hywan: Good catch! Let's try with a PR then.

Hywan: PS: Use `position()` instead of `find` because we expect `Option<usize>`, not `Option<u8>`.

Hywan: PPS: Actually, `position()` passes an item to the closure, not the offset, so it might be more difficult to adapt the existing algorithm to an iterator :-/.

fflorent: I think you miss the fact that the Memchr iterator doesn't iterate the occurrences but the position of these occurrences. Quoting the [documentation](https://docs.rs/memchr/1.0.1/memchr/struct.Memchr.html):
> Creates a new iterator that yields all positions of needle in haystack.

It also mentions that `type Item = usize`, and `find` returns `Option<Self::Item>` :).

Hywan: Damn, I am really tired…

fflorent:
Haha :)Geal/nom a écrit :@Hywan commented on this pull request.



In src/traits.rs:
> @@ -448,12 +444,34 @@ pub trait FindSubstring<T> {
 
 impl<'a,'b> FindSubstring<&'b [u8]> for &'a[u8] {
   fn find_substring(&self, substr: &'b[u8]) -> Option<usize> {
-    for (index,win) in self.windows(substr.len()).enumerate() {
-      if win == substr {
-        return Some(index)
+    let substr_len = substr.len();
+
+    if substr_len == 0 {
+      None
+    } else if substr_len == 1 {
+      memchr::memchr(substr[0], self)
+    } else {
+      let max = self.len() - substr_len;
+      let mut offset = 0;
+      let mut haystack = &self[..];
+
+      while let Some(position) = memchr::memchr(substr[0], haystack) {

Damn, I am really tired…

src/traits.rs
Hywan: @fflorent Did you open a PR for that?

fflorent: Damn, sorry for having forgotten this topic.

I create the PR right now. Tell me what you feel about the changes :).

Florent

fflorent: Done : https://github.com/Geal/nom/pull/565

PR #513, opened 2017-05-15
`find_substring` returns `None` if empty substr

Hywan:
Calling `input.find_substring(b"")` will make the thread to panic with this error:

> thread '…' panicked at 'assertion failed: size != 0', src/libcore/slice/mod.rs:337

This is not possible to create windows of length 0.


Hywan:
Close in favor of #513.

PR #515, opened 2017-05-19
allow non primitive bits encoded integers

nizox:
`take_bits!` returns a bits encoded value of the type passed to the macro by argument. This type must implement some operator traits (Shr, Shl, AddAssign) and a conversion from `u8`. Only rust integer and float primitive types can do this conversion according to numeric cast rules.

These changes make the conversion explicit in order to allow non primitive types. It is then possible to parse `num::bigint::BigInt` integers with `take_bits!`.


Geal:
nice find, thanks :)

PR #516, opened 2017-05-19
fix doc typo

grissiom:
Both take_till_s and take_till1_s recieve function of type "char ->
bool" instead of "&str -> bool".


Geal:
Thanks!

PR #519, opened 2017-05-24
Removed leading Zeros in Ranges

tstorch:
Hope this is something you`re (mildly) interested in:

Removing leading zeroes in ranges helped making the code a bit
more intuitive. Also this made Range<usize> as a trait
superflous in some cases.

ATTENTION: This also ment in two case to switch from RangeFrom to
RangeFull. This might break some code which does not rely on `str`
ot `[T]` for which the Slice trait is implemented for all kinds of
Ranges by nom.

### DIFF
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -718,7 +718,7 @@ mod tests {
         let res:$crate::IResult<&[u8],&[u8]> = if $i.len() < cnt {
           $crate::IResult::Incomplete($crate::Needed::Size(cnt))
         } else {
-          $crate::IResult::Done(&$i[cnt..],&$i[0..cnt])
+          $crate::IResult::Done(&$i[cnt..],&$i[..cnt])
         };
         res
       }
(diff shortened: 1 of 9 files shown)


Geal:
Hi! It is great that you want to spend time on nom, but I'd prefer you work on bugs and features than refactorings. As an informal rule, a refactoring is useful when it improves the architecture of the code to either reduce the possibility of error or facilitate feature development.
This PR does not really improve the code. This is the same issue with your #186 PR, which replaced a correct and straightforward implementation with a convoluted one for no clear benefit.
I want to inist that there's no animosity in this, I really appreciate that you take some time to work on nom. But unsolicited refactorings like these just increase my workload and add no value.

tstorch:
I'm sorry to hear that. For me the added value is readability (and dropping some previously necessary trait restrictions on some functions). Regarding #186 : Code seems to be very subjective as I thought my code was more straight forward :smiley: (judging by the PR it even fixed some issues)

When do I know, when a PR has added value? I have a patch  (unfortunately based on this commit, but this can be changed) laying around consolidating some stuff into macros.

PR #529, opened 2017-06-24
fixed parsing of factors without spaces

### LINKED ISSUE #528: broken factor parsing combinator in tests/arithmetic_ast.rs
The implementation of the `factor` parser is broken for integers without any spaces because the `opt!` is not wrapped in a `complete!` macro. The rationale for using `complete!` is documented [here](https://github.com/Geal/nom/issues/406#issuecomment-274256680)

jaupe:
This fixes #528

### DIFF
--- a/tests/arithmetic_ast.rs
+++ b/tests/arithmetic_ast.rs
@@ -64,7 +64,7 @@ named!(factor< Expr >, alt_complete!(
     map!(
       map_res!(
         map_res!(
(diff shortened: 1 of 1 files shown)


Geal:
inserting `ws!` here does not make much sense. Since you wanted to wrap `opt!` with `complete!`, why not make another parser for this and replace `opt!(multispace)` in the expression?

Geal:
fixed in 47deaa1

PR #532, opened 2017-07-02
Create a bytes! macro, inverse of bits!

### LINKED ISSUE #158: Make chaining `take_bits!` with other non-bit macros more ergonomic
Currently I'm trying to use the `take_bits!` macro in a `chain!` with other macros that work on bytes, not bits, and I'm using other functions that could be replaced with take bits, but it would be better not to (i.e. I'd rather use `be_u16` instead of `take_bits!(u16, 16)`).

It seems like this isn't possible because the type signatures don't match up. I'm not sure what the best solution/workaround for this is.

gmorenz:
Currently we can go from the world of byte parsers to the world of bit parsers, but not back (in any convenient way). This lets you go back. An example of parsing an ipv4 packet with it is [here](https://gitlab.com/snippets/1666733).

The actual implementation is pretty much a direct copy of the `bits!` macro with a few bits changed.

Possibly fixes #158?

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -45,7 +45,7 @@ macro_rules! bits_impl (
     {
       let input = ($i, 0usize);
       match $submac!(input, $($args)*) {
-        $crate::IResult::Error(e)                            => {
+        $crate::IResult::Error(e) => {
           let err = match e {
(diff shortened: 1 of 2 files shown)


Geal:
Thanks!

PR #540, opened 2017-07-16
Don't use i2 after moving it to $submac

jethrogb:


Geal:
was there a compilation error without this?

jethrogb:
Yes. I've been trying to use iterators instead of slices. You'll get a "use of moved value: i2" error without this patch.

Geal:
ok, I thought using iterators as input type would not work because they would need to be cloned in the branching combinators like `alt!`, but I just remembered it's actually how it works now in those combinators. Because some input types might not be `Copy`, it was added, and for `&[u8]` and `&str`, cloning boils down to copying.
The change does not break anything, so go for it :)
Also, would you like to write something about how you use iterators as input type for nom? You're using that to parse a list of tokens, right?

jethrogb:
I was using slices (`&[Token]`) before. This is so far just an experiment. And yeah, there's a bunch of clone already used, this was the only change needed to make it work. Of course, it's possible there are macros I'm not using that still don't work.

One of these days... I'll write something up about `cexpr` (but keep in mind I said that last year too...). The iterator thing was pretty straightforward actually. I spent more time trying to box up a cloneable iterator trait object than I spent making nom work.

PR #541, opened 2017-07-16
Add native endianness.

kurnevsky:
It might be useful if you don't care about endianness (e.g. parsing random ids).


Geal:
why do you need this when there are already the `be_*` functions?

kurnevsky:
Because these functions will convert differently depends on target architecture. And since they use just transmute they might be faster then be_*. Also it might be useful to show that you don't care about endianness in this particular case.

kurnevsky:
Reimplemented all parsers, finally. But tests don't compile yet.

PR #549, opened 2017-08-03
Add Examples To Documentation

bheisler:
I've added some examples to the doc comments in the bits, bytes and character modules for macros that didn't have examples previously. I've also reworked the prose descriptions and format of the doc comments in the bits module, but if you'd like that to be reverted or done as a separate pull-request that's fine.


bheisler:
I'm not sure why code coverage would go down; all this change does is add more tests. Perhaps someone from the project could clarify.

Geal:
Thanks!

PR #555, opened 2017-08-12
Use stdlib compile_error! in compiler_error! macro

roblabla:
Removes the dependency on `compiler_error!`, which didn't work on my version of nightly for whatever reason (it complains that it can't find the `compiler_error!` macro).

`compile_error!` is stable and will land in 1.20: https://github.com/matthewhammer/rust/commit/87483f93cea2a2b7b7649f3c2e88ce7d2ce9801f

Macro docs : https://doc.rust-lang.org/nightly/std/macro.compile_error.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Changed
 
+## Unreleased
+
+### Changed
+
+- the `compiler_error` macro now uses the stdlib `compile_error` macro on nightly.
+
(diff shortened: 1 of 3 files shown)


lu-zero:
Great :) !

Geal:
Hello, thank you for the PR. In the end, the `compiler_error` macro was removed in 1d62d99a and completely replaced with `compile_error`

Issue #561, opened 2017-08-18
type annotations needed


epilys:
```
named!(value<std::vec::Vec<&str>>,
                      do_parse!(
                          first_line: map_res!(is_not_s!("\n"), std::str::from_utf8) >>
                          rest: many_m_n!(0, 1, separated_list!(tag!("\n\t"), map_res!(take_while!(call!(|c| c != '\n' as u8)), std::str::from_utf8))) >>
                          ({
                              let mut v : Vec<&str> = rest.iter().fold(Vec::new(), |mut a:Vec<&str>,ref mut x| {
                                  a.append(x); a});
                              v.insert(0, first_line); v}) 
                          )
                      );

```

```
   Compiling parser v0.1.0 ()
note: trace_macro
  --> main.rs:27:1
   |
27 | / named!(value<std::vec::Vec<&str>>,
28 | |                       do_parse!(
29 | |                           first_line: map_res!(is_not_s!("\n"), std::str::from_utf8) >>
30 | |                           rest: many_m_n!(0, 1, separated_list!(tag!("\n\t"), map_res!(take_while!(call!(|c| c != '\n' as u8)), std::str::from_utf8))) >>
...  |
35 | |                           )
36 | |                       );
   | |________________________^
   |
   = note: expanding `named! { value < std :: vec :: Vec < & str >> , do_parse ! (
           first_line : map_res ! ( is_not_s ! ( "\n" ) , std :: str :: from_utf8 ) >>
           rest : many_m_n ! (
           0 , 1 , separated_list ! (
           tag ! ( "\n\t" ) , map_res ! (
           take_while ! ( call ! ( | c | c != '\n' as u8 ) ) , std :: str :: from_utf8 )
           ) ) >> (
           {
           let mut v : Vec < & str > = rest . iter (  ) . fold (
           Vec :: new (  ) , | mut a : Vec < & str > , ref mut x | { a . append ( x ) ; a
           } ) ; v . insert ( 0 , first_line ) ; v } ) ) }`
   = note: to `# [ allow ( unused_variables ) ] fn value < 'a > ( i : & 'a [ u8 ] ) -> $crate
           :: IResult < & 'a [ u8 ] , std::vec::Vec<&str> , u32 > {
           do_parse ! (
           i , first_line : map_res ! ( is_not_s ! ( "\n" ) , std :: str :: from_utf8 )
           >> rest : many_m_n ! (
           0 , 1 , separated_list ! (
[...]

PR #562, opened 2017-08-19
src/multi.rs: fix 'type annotations needed' error

epilys:
separated_list!() gives an "error[E0282]: type annotations needed", with
  cannot infer type for `E`
  consider giving `ret` a type

This stops the error.


epilys:
fixes #561

PR #563, opened 2017-08-19
Ensure the docs for `do_parse!` are visible

sgrif:
The documentation for `do_parse!` was not visible without scrolling, as
the matchers take up the entire screen (the majority of which are not
meant to be visible to the user). This hides all matchers which start
with `__impl`, to ensure that the documentation appears on the page.

Before:

screen shot 2017-08-19 at 1 57 34 pm

After:

screen shot 2017-08-19 at 1 58 04 pm

### DIFF
--- a/src/sequence.rs
+++ b/src/sequence.rs
@@ -360,15 +360,38 @@ macro_rules! delimited(
 ///
 #[macro_export]
 macro_rules! do_parse (
-  (__impl $i:expr, $consumed:expr, ( $($rest:expr),* )) => (
+  ($i:expr, $($rest:tt)*) => (
+    {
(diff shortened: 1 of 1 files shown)


Geal:
the annoying thing with making a separate impl parser is that if someone selectively imports macros instead of all nom macros, then they'll get errors because of missing macros that they can't find in the docs. Some people do that kind of import, but I don't know how many, and there should be less of them since there's now no more conflict with log's macros.

I think it could be done for do_parse, still, since the docs clearly look bad that way

sgrif:
Luckily, they will get a very specific error message telling them the missing macro name in that case. I agree it's a papercut, but I think it's a small one and impacts a smaller portion of users than the documentation being in this state.

PR #565, opened 2017-08-20
Rework of FindSubstring using iterator instead

fflorent:
Use an iterator of Memchr for FindSubstring, accordingly to the discussion in #507.

Don't hesitate to bikeshed. Also I won't be offended if you don't feel that's worthy :).

Florent


fflorent:
@Hywan In order to check how fast is this implementation compared to the previous one, do you still have the bench test for this https://github.com/Geal/nom/pull/507#issuecomment-301423469?

Thanks in advance :)

Hywan:
@fflorent No I don't, sorry :-(.

src/traits.rs
```diff
@@ -467,25 +467,18 @@ impl<'a,'b> FindSubstring<&'b [u8]> for &'a[u8] {
       memchr::memchr(substr[0], self)
     } else {
       let max = self.len() - substr_len;
```
Hywan: Add `+1` directly to the definition of `max`.

src/traits.rs
```diff
@@ -467,25 +467,18 @@ impl<'a,'b> FindSubstring<&'b [u8]> for &'a[u8] {
...
-        if &haystack[position..position + substr_len] == substr {
-          return Some(offset)
-        }
-
-        haystack  = &haystack[position + 1..];
-        offset   += 1;
-      }
-
-      None
+      let haystack = &self[..max + 1];
```
Hywan: And then remove `+1` here.

fflorent: Done, thanks!

fflorent:
I have committed a banchmark in `specific.rs` for `take_until_and_consume`. 

Here are the results:

```
-- before
running 1 test
test take_until_and_consume_bench ... bench:         272 ns/iter (+/- 6)

-- after
running 1 test
test take_until_and_consume_bench ... bench:         310 ns/iter (+/- 3)
````

So… not really good I would say. There might be tricks to catch up the previous performance, I am doing tests.

fflorent:
A quick look didn't give me any hint on how to keep the optimization. I guess such a gap of performance between the previous solution and this proposal excludes the latter.

May I close the PR? Or do you (@Hywan, @Geal) have some ideas on your side?

Hywan:
Using an iterator would be better for safety and readiness for sure. However I am not able to obtain better performances with the iterator API.

It's up to @Geal to decide :-).

PR #566, opened 2017-08-22
Add pub(crate) support

derekdreery:
Adds pub(crate) support where pub support is present in `macros.rs`.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -89,8 +89,10 @@ macro_rules! closure (
 /// // will use &[u8] as input type (use this if the compiler
 /// // complains about lifetime issues
 /// named!(my_function<&[u8]>,            tag!("abcd"));
-/// //prefix them with 'pub' to make the functions public
+/// // prefix them with 'pub' to make the functions public
 /// named!(pub my_function,               tag!("abcd"));
+/// // prefix them with 'pub(crate)' to make the functions public within the crate
+/// named!(pub(crate) my_function,               tag!("abcd"));
 /// ```
 #[macro_export]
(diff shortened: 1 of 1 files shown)


derekdreery:
Currently there are no extra tests - I'll  see if I can add some.

Geal:
hi! apparently the test cannot compile on current stable. IIRC that feature is still experimental, right?

derekdreery:
Aah I thought it was already stable. You are correct. Can I mark this PR as blocked by upstream until the syntax becomes stable?

derekdreery:
I've tested and it works on beta - not stable.. I think it was stabilized in 1.19, but maybe there are some changes in macros to support it in 1.20. Anyway, all tests pass in 1.20, so when that's released this should be mergeable.

derekdreery:
1.20 has been released now, so this should work

derekdreery:
Version 1.20 is out, so this should no longer be blocked. I need to trigger a ci reload.

derekdreery:
I think there's an unstable 'vis' identifier, that would make all these variants unnecessary

CAD97:
[Tracking issue for `:vis` macro matcher](https://github.com/rust-lang/rust/issues/41022)

Seems to be almost en route to stabilization.

Geal:
Thank you! This is getting in nom at last :)

derekdreery:
Note that this PR doesn't actually use `:vis`, it duplicates the code for the different varieties. the code could be reduced considerably by using the now-stabilized `:vis` macro ident type.

PR #567, opened 2017-08-25
Adding link to GZIP parser

nharward:
Great framework! May try MP3 next, my last one was hand-written in Erlang but nom has got to be easier.


Geal:
awesome!

PR #571, opened 2017-08-30
Make rest parser generic

### LINKED ISSUE #559: rest does not work on input other than &[u8]
Unlike some other functions, rest only accepts &[u8] and has a separate rest_s for &str.

kamarkiewicz:
Hi! I think the `rest` parser should work with various input types. I've changed signature of rest to be more generic over input. The `rest_s` parser is from now on a good candidate to deprecation.

It is my second PR to `nom`, the first one was not merged, so please do a review and tell me what should I change to make it mergeable.

Fixes #559


Geal:
Thanks!

PR #574, opened 2017-08-30
Make escaped! and escaped_transform! usable with method!

dmski:
escaped! and escaped_transform! expanded to something that wasn't an expression (due to there being a return statement there).
This made them impossible to use with method!.
Just removing returns seems to have helped.

I've also had to fix some tests compilation error that only occured when running "cargo test --features verbose-errors"


Geal:
those `return` were removed in later versions, closing this now

Issue #522, opened 2017-06-01
Documentation effort


Geal:
As we know, the [nom documentation](https://docs.rs/nom) is not that great. It has been improving, but not fast enough, and not with a satisfying quality. It's getting to the point where [anytime someone talks about nom, someone else chimes in to say the documentation is crap](https://www.reddit.com/r/rust/comments/6efmov/rust_toronto_meetup_introduction_to_parsing_using/). This is unacceptable and a big failure on my part.

So, let's improve things:
- the [first page](https://docs.rs/nom/3.0.0/nom/index.html) is daunting. There's a big example that uses advanced features like `fold_many0`, that does not help comprehension. Maybe the [JSON example](https://github.com/Geal/nom/blob/master/tests/json.rs) would be better? (provided we remove the float parsers and replace with the builtin ones). There should be an obvious path: "where do I go if I want to parse this kind of format?" "I know how to create a first parser, how do I find the combinators I need?" "I got this error, how do I fix it?"
- the big list of combinators is not helpful to beginners. Once you know how nom works, displaying a kind of function definition works to explain what it does, but it's not enough in a lot of cases. It does not help that we have limited room for the text of each macro on the first page. We could add more documentation inside modules, like the [whitespace module](https://docs.rs/nom/3.0.0/nom/whitespace/index.html), but how do we make sure beginners will check those out instead of going directly to the macros list?
- the [guides](http://rust.unhandledexpression.com/nom/home.html) have broken links. They're available on my own doc hosting, but docs.rs cannot render them, so the links are broken.
- the story around `named!`, `named_args!`, custom functions, methods, is confusing. Those combinators were designed to hide how the input is passed to parsers and reduce confusion, but they do not help much. Since they're the first combinators people will use, they need to be explained clearly.
- most combinators only have a one liner as documentation. I don't think large code examples will help so much there, they don't improve readability. But some small examples, with maybe some kind of table with the expected result for some inputs, would be better to get a feel of how combinators work.

This will be a lot of work, and I probably won't be able to do it all alone, but it has to be done.

sharazam:
@Geal Like I said on reddit, it needs a lot of work. I also said that I'd do a PR, but for now I've just rewritten the lib.rs file - if you could look over it, that'd be nice: https://github.com/sharazam/nom-documentation-readme/blob/master/README.md

It's not finished, largely based on a transcript of the Toronto video.

[...]

PR #576, opened 2017-08-31
Expand documentation for bits! combinator

kamarkiewicz:
I've expanded the documentation of `bits!` combinator as some users have problems with its right usage.
English is not my native lang (Rust is 😄), so please free to correct me.

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -13,7 +13,7 @@
 
 
 /// `bits!( parser ) => ( &[u8], (&[u8], usize) -> IResult<(&[u8], usize), T> ) -> IResult<&[u8], T>`
-/// transforms its byte slice input into a bit stream for the underlying parsers
+/// Transforms its byte slice input into a bit stream for the underlying parsers.
 ///
 /// ```
 /// # #[macro_use] extern crate nom;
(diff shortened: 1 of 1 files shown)


src/bits.rs
```diff
@@ -13,19 +13,40 @@
...
-///  let sl    = &input[..];
+///  let bytes = vec![0b1010_1010, 0b1111_0000, 0b0011_0011];
+///  let slice = &bytes[..];
+///
+///  assert_eq!( take_3_bits( slice ), Done(&slice[1..], 5) );
+/// # }
+/// ```
+///
+/// The consumed bits counter is local to a `bits!` macro invocation, so the remaining result
+/// is not `(&[u8], usize)` anymore. Second element is bit index in the stream - esentional to parse
```
leogomes: esentional  -> essential

kamarkiewicz: Thanks, fixed!

kamarkiewicz:
I found a documentation effort committed by @bheisler [here](https://github.com/Geal/nom/issues/522#issuecomment-318531773) and I think he made a really great work on this! Check his proposition first. This little PR describes only single parser and is not in contradiction to bheisler's work, just clarifies some use cases.

src/bits.rs
```diff
@@ -13,19 +13,40 @@
...
+/// The consumed bits counter is local to a `bits!` macro invocation, so the remaining result
+/// is not `(&[u8], usize)` anymore. Second element is bit index in the stream - essential to parse
+/// all the bits or else they will be dropped.
 ///
-///  assert_eq!(take_3_bits( sl ), Done(&sl[1..], 5) );
+/// ```
+/// # #[macro_use] extern crate nom;
+/// # use nom::IResult::Done;
+/// # fn main() {
+///  named!( take_z_word<(u8, u16)>,
```
willmurphyscode: The name `take_z_word` is from a project I was working on when I asked the question that led to this documentation change. (I am writing a parser for files from the game Zork.) But this name will not mean anything to most people. Is there a more generic name we could use?

src/bits.rs
```diff
@@ -13,19 +13,40 @@
...
+///      bits!(
+///          tuple!( take_bits!( u8, 5 ), take_bits!( u16, 10 ) )
+///      )
+///      // the last one bit of input remains unconsumed (as 5 + 10 != 16) and bit index is dropped
```
willmurphyscode: The phrase `the last one bit of input remains unconsumed` makes me think that it will still be available to the next parser. But actually nom consumed the whole byte, and then ignored the last bit of it. Would it make more sense to say "ignored" or "skipped," since the current parser doesn't consume it, but subsequent parsers won't be able to consume it either?

PR #583, opened 2017-09-20
err cast to errorkind robust against verbose-error feature state

kali:
This creates a `Err::into_error_kind()` that will work the same whether or not verbose error are activated.


Geal:
Thanks!

PR #585, opened 2017-09-28
use docs.rs for documentation

vadixidav:


Geal:
Hi!
as long as docs.rs does not allow adding markdown files to the documentation, there will still be a separate documentation

PR #586, opened 2017-10-01
Simplify InputIter for byte slices

target-san:
Current implementation uses Item and RawItem types in InputIter - because string slice iterators return chars by value directly, and byte slice iterators return slice bytes by reference. This PR makes IntoIter implementation for byte slice perform additional mapping under the hood, so that it also returns bytes by-value. RawItem is left for backwards compatibility.

Rationale for this change: if one wants to write some kind of generic code to work uniformly with both bytes and chars, he won't be stuck with value vs reference dichotomy.


Geal:
Thanks!

PR #594, opened 2017-10-14
introduce term "remainder" (and update nightly, because I fail at git)

### LINKED ISSUE #592: macro/function explanations are too alike for beginners
The differences between the variations of a macro seem to be really subtle from a beginners perspective. The wording could help significantly in understanding these differences.

An example from the docs:
1. take_until: consumes data until it finds the specified tag
2. take_until_and_consume: generates a parser consuming bytes until the specified byte sequence is found, and consumes it

Problems:
* (1) says _data_ and (2) says _bytes_
* (1) says _tag_ and (2) says _byte sequence_
* (1) and (2) appear to _consume_ the data

Fortunately I received help on IRC (thanks @s3bk!) and was explained that (1) doesn't consume the tag while (2) does, which makes a huge difference.
I'm not in a position to know and properly describe these subtleties yet so I can only file this issue about it ;-)
I'm surprised how complete the docs already are, now they just need some fine tuning 🙌

s3bk:
Attempt to resolve #592


Geal:
Thanks!

PR #595, opened 2017-10-16
Doc test `take_till!` and `take_till1!`

willmurphyscode:
This is a small contribution to https://github.com/Geal/nom/issues/522. It adds doc tests / examples to the `take_till!` and `take_till1!` macros, including assertions that illustrate the difference between these two macros. 

One piece of feedback I'd like is on the wording of the comments - I say "empty match is allowed" and "empty match is error" to illustrate the difference between `take_till!` and `take_till1!`, but I'm not sure I've used nom's usual terminology correctly.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -451,6 +451,8 @@ macro_rules! take_while1 (
 ///
 ///  let r = till_colon(&b"abcd:efgh"[..]);
 ///  assert_eq!(r, Done(&b":efgh"[..], &b"abcd"[..]));
+///  let r2 = till_colon(&b":abcdefgh"[..]); // empty match is allowed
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -442,6 +442,19 @@ macro_rules! take_while1 (
 /// returns the longest list of bytes until the provided function succeeds
 ///
 /// The argument is either a function `&[T] -> bool` or a macro returning a `bool
+///
+/// ```
+/// # #[macro_use] extern crate nom;
+/// # use nom::IResult::Done;
+/// # fn main() {
+///  named!( till_colon, take_till!(|ch| ch == b":"[0]) );
```
bluss: you can use `b':'` which is a literal for a single byte.

Geal:
great, thanks!

PR #596, opened 2017-10-17
Doc comment characters.rs

willmurphyscode:
As part of https://github.com/Geal/nom/issues/522, this PR adds examples in the doc comments for `src/characters.rs`.


Geal:
awesome, thanks!

PR #599, opened 2017-10-28
Request to add bnf to list of projects using nom

snewt:
This commit represents an addition to the list of known projects
using nom. I wrote a tool to parse BNF grammars using nom called bnf
available on Github [here](https://github.com/snewt/bnf) with the
corresponding crate documentation [here](https://crates.io/crates/bnf).
I really enjoyed learning to use nom! Thanks for making and maintaining
it! The project's documentation was awesome and it really helped me
through the process.


Geal:
oh, hey, that's great!
That means you could make a crate that parses a BNF, then generates the corresponding nom code? :D

snewt:
@Geal ah yeah, code generation is an ideal use case!

PR #602, opened 2017-11-03
ability to specify return type & input type for named args

vmchale:
This will allow me to use the `named_args!()` macro with custom input types and output types.


Geal:
Hello,

I added those alternatives in 54ee6ff8 and f2d658ed2f6 because it was difficult to merge correctly, but I credited you in the changelog (c3636fd).

PR #603, opened 2017-11-05
Allow trailing comma

HyeonuPark:
Allow trailing comma within macro call.
And some tiny fixes I found.

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -28,10 +28,10 @@
 /// # }
 #[macro_export]
 macro_rules! bits (
-  ($i:expr, $submac:ident!( $($args:tt)* )) => (
+  ($i:expr, $submac:ident!( $($args:tt)* ) $(,)*) => (
     bits_impl!($i, $submac!($($args)*));
   );
-  ($i:expr, $f:expr) => (
+  ($i:expr, $f:expr $(,)*) => (
     bits_impl!($i, call!($f));
   );
 );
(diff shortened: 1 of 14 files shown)


Geal:
hmm, that's a lot of changes for something I'm not really convinced is useful. Why do you need trailing commas in argument lists?

HyeonuPark:
Nom macros tend to be deeply nested, so it should usually be split into multiple lines.

And there's two major style rules for multiline function call: comma-first and comma-last

I saw you prefer comma-first style for your own code, but [many people in Rust community likes comma-last with trailing comma](https://github.com/rust-lang-nursery/fmt-rfcs/issues/42).

So it will be nice to allow trailing comma in nom macros!

perlun:
> I saw you prefer comma-first style for your own code, but many people in Rust community likes comma-last with trailing comma.

To give my 0,02€ of this, we really don't _know_ this. We just know that those who like trailing comma tend to raise their voice louder and share their opinion.

Until someone really does an exhaustive poll/examination of some sort, I think we should be careful to draw too far-reaching conclusions from that discussion.

(Disclaimer: I am the author of the linked issue, where I advocated that the default Rust style should not advocate trailing comma. 😃 Since writing that flamebait, I have given trailing comma a chance in a minor TypeScript project, and it's not _that_ bad once you get used to it.)

Geal:
closing this, it's not really useful for nom, sorry

PR #605, opened 2017-11-13
Fix typo in CHANGELOG

dbrgn:
I hope this change doesn't end up in the changelog, because if you accidentally mistype it again we'll end up IN AN ENDLESS RECURSIVE LOOP OF MISSPELLING, PULL REQUESTS, DEATH AND DEVASTATION 😱 

(As requested on Reddit, this fixes the typo in my hard-to-type username 🙃)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@
 ### Thanks
 
 - Chris Pick for some `Incomplete` related refactors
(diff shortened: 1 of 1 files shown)


Geal:
*makes a note to consistently mistype this username in every future changelog*
Thanks :)

PR #606, opened 2017-11-13
Add nom-sql to list of parsers in README

ms705:
[nom-sql](https://github.com/ms705/nom-sql) is an incomplete, but already useable SQL parser (e.g., used in [distributary](https://github.com/mit-pdos/distributary)).

It's currently written using very much outdated nom 1.x APIs, but will upgrade in the future.


Geal:
alright!
You should totally update to newer versions, nom 4 will be very interesting :)

PR #610, opened 2017-11-19
Use alloc instead of collections

iBelieve:
`collections` has been merged into `alloc` in the latest nightly.

See https://github.com/rust-lang/rust/pull/42648


Geal:
Thanks!

PR #612, opened 2017-11-19
Fix take_until_and_consume1!

dtrebbien:
- Using the macro would cause compilation errors.
- The generated parser combinator would return the incorrect needed amount.
- ErrorKind was missing a variant for take_until_and_consume1!.
- Ensure that the returned sequence is non-empty.


dtrebbien:
I rebased to resolve the conflict.

Geal:
thanks!

PR #573, opened 2017-08-30
Travis CI improvements

kamarkiewicz:
Hi! I've spent some time to improve Your CI.

- Removed duplications and fixed identation
- Fixed running in matrix for various FEATURES
- Removed usage of travis-cargo python script, as it was used only to check the currently running rust version

I'm open for any suggestions.

Related to #500

PR #634, opened 2017-12-04
Travis CI more improvements

### LINKED ISSUE #500: CI is lacking coverage of previous rustc versions and feature flag combinations
Nom 2.2.x requires rustc 1.9.0 due to a [deprecated annotation](https://github.com/Geal/nom/commit/b5e21c00e8a52b7dfec485245ba2dceafe5ecaa5).  Nom 2.0.0 passes the test suite with rustc 1.8.0.  The travis configuration appears to be targeting rustc 1.2 and passes, so there is almost certainly something wrong with the CI setup.

kamarkiewicz:
This pull request is a follow up to #573 
In https://github.com/Geal/nom/pull/573#issuecomment-326286088 I showed a rewritten version but didn't make a PR. Please check out this changes.

- The more flexible matrix configuration
- Split of .travis.yml into configuration part and execution script part (graphically it uses two newlines to mark this)
- Dropped Rust 1.2.0, as master doesn't compile on this version anymore. Maybe it is worthwhile to add some minimal version of Rust compiler? Are there branches that you support with 1.2.0?
- Bash's case used to show differences in scripts running on different versions. It looks much clearer this way
- Removed use of stream flag
- Compilation without std flag is now done only on nightly, as on other compiler versions it just won't work. It is OK, as it uses `feature(..)` mechanism. (maybe someday `alloc` will go stable and then we could test against other channels)
- better coverage reports; Coveralls shows percentages per job, which is now per a set of active features

I think this closes #500

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,7 @@ env:
     - FEATURES='' # default ones
     - FEATURES='--features "regexp"'
     - FEATURES='--features "regexp regexp_macros"'
-    - FEATURES='--features "stream regexp regexp_macros"'
-    - FEATURES='--features "stream regexp regexp_macros" --no-default-features'
+    - FEATURES='--features "regexp regexp_macros" --no-default-features'
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


Geal:
oh, this looks much better! I'm sorry, I had not seen this PR when I made some changes a few days ago. Thank you for your help!

Geal:
the tests don't pass right now because the regexp feature does not work without std, and the tests don't work without std either. I'll fix this right away

PR #638, opened 2017-12-10
Reduce CI times

kamarkiewicz:
Not forcing `cargo install` on every build can save CI a lot of time. Cargo, unfortunately, doesn't have some CI flag that would install or update the tools if needed. I applied the workaround described in issue https://github.com/roblabla/cargo-travis/issues/17

Latest master: https://travis-ci.org/Geal/nom/builds/314349125
Ran for 32 min 21 sec
Total time 2 hrs 10 min 14 sec

This PR: https://travis-ci.org/Geal/nom/builds/314369454
Ran for 13 min 4 sec
Total time 56 min 48 sec

These speedups come with some downside. If one of the tools stop working, you may need to clear Travis cache manually to install updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@ matrix:
       env: FEATURES='--features "nightly"'
     - rust: nightly
       env: FEATURES='--no-default-features'
-      before_script:
-        - export PATH=$HOME/.cargo/bin:$PATH
-        - which cargo-coveralls || cargo install cargo-travis
-        - which cargo-benchcmp  || cargo install cargo-benchcmp
     - rust: stable
       env: FEATURES=''
     - rust: stable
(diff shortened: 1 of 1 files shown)


Geal:
Hello,

I worked a bit on this earlier: c6c4d7c
Same result as you, build time is around 15mn

kamarkiewicz:
Yeah, your changes look better :) But I advise you to move following code:

      before_script:
        - export PATH=$HOME/.cargo/bin:$PATH
        - which cargo-coveralls || cargo install cargo-travis
        - which cargo-benchcmp  || cargo install cargo-benchcmp

outside `matrix::include` part, as this tools are required in all builds. Put those lines just between `addons` and `script` part, to run this section on all builds. In `matrix::include` you can put only overrides on the specific build.

Also, the following code shouldn't be requiring `eval`, as it is just a trick to inject env variables as program arguments:

        - eval cargo fmt -- --write-mode=diff

kamarkiewicz:
Looks like Your CI works and as this PR tries to do, it works fast. If something goes wrong, then we can fix it in different PR.

PR #645, opened 2017-12-29
Add basic calculator language parser example to README

balajisivaraman:
I created a very simple parser/evaluator for a basic calculator language using Nom. The goal was to learn the idioms Rust and Nom using a relatively simple example, something that is larger than just math operations, but smaller than a full fledged language or text file format.

It demonstrates the use of most of the fundamental Nom idioms (the primitive parser macros - `alpha!`, `digit!`, `ws!`, `char!` , `tag!`, and a bunch of other combinators; `do_parse!`; and combining smaller parsers into bigger ones). It also includes an example of parsing from a file using the older streaming API (`FileProducer` and a custom `Consumer`). I plan to keep this repository updated with Nom's API changes. So once the new generators API is finalised and released, I'll update that example as well.

My hope is that others new to both Rust and Nom can use this repository as a place to get started with both. (I put it under the Programming Languages category in the example because I couldn't find a better place for it.)

Thank you for the great library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ Here is a list of known projects using nom:
   * [torrc configuration file](https://github.com/dhuseby/torrc-rs)
   * [Web archive](https://github.com/sbeckeriv/warc_nom_parser)
 - Programming languages:
+  * [Basic Calculator](https://github.com/balajisivaraman/basic_calculator_rs)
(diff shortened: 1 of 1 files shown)


Geal:
thanks!

PR #646, opened 2017-12-29
Lift some restrictions in the floating-point number parser

### LINKED ISSUE #437: float, float_s, double and double_s can not recognize numbers without dots

veprbl:
Currently, the ```double``` parser will reject a floating point number presented as ```1243e5```.
Other examples of a rejected formats are ```1234.``` and ```.1234```, which are not a valid Rust floating-point literals [1], but they are widely used in text formats and are accepted by parser from the standard library.

I also drop ```complete!``` wrappers from the parsers. ```Nom``` claims to support streaming, but current parser with ```complete!``` may fail if exponent falls on the boundary of the input buffer. At the same time, it seems that, in most practical situations, we don't need to parse floats in the end of the output, so it should be fine to leave that case out of the scope. Overall, I don't have a strong opinion here, but I think that if ```complete!``` is to stay in any place, there should be a notice in the documentation.

[1] https://doc.rust-lang.org/reference/tokens.html#floating-point-literals

Fixes #437

### DIFF
--- a/src/nom.rs
+++ b/src/nom.rs
@@ -679,8 +679,8 @@ pub fn float(input: &[u8]) -> IResult<&[u8], f32> {
       tuple!(
         opt!(alt!(tag!("+") | tag!("-"))),
         alt!(
-          delimited!(digit, tag!("."), opt!(complete!(digit)))
-          | delimited!(opt!(digit), tag!("."), digit)
+          value!((), tuple!(digit, opt!(complete!(tag!("."))), opt!(complete!(digit))))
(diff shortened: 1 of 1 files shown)


Geal:
hey, so, I arrived at the same solution as you ( 99e4cea ), but didn't merge your PR, so I'll be sure to at least put you in the "thanks" part in the changelog :)

veprbl:
It works! Thank you @Geal!

Issue #676, opened 2018-02-04
no_std: feature guard for parsers that need alloc


kamarkiewicz:
Libraries that uses nom but don't have to make use of `Box` could be built on stable rust.
It is not possible, for now. The temporary solution is to fork nom, remove the following line and all parsers that make use of alloc.
```#![cfg_attr(not(feature = "std"), feature(alloc))]```
If the feature guard is possible, how it should look like? I'm thinking of a new feature called `alloc`.
I saw You have `nightly` feature already, so maybe I could reuse it in terms of the on-switch if `nightly` is used.

I can work on this and make a PR if You want.

Geal:
oh, that's a nice idea, and would be easily compatible with the new plan for no_std/std/core, etc.

glandium:
This was implemented in ba357743de04b2549776f5b712319c46c1ea57bd but the merge conflict resolution in 43ed4e99f39bc43f50178452def4e1797db8c95e broke it by adding a `#![cfg_attr(not(feature = "std"), feature(alloc))]` line to src/lib.rs.

Geal:
hello, this should be fixed in nom 4

PR #677, opened 2018-02-05
WIP Add `alloc` feature guard

kamarkiewicz:
Related to #676 
I started to work on proposed feature guard, but I encountered some issues.
Some core lib API is not stable yet which causes my troubles with selecting the right impl for `as_bytes` and some char methods.
You can see I just added `unimplemented!()` there just to make things compile on stable. Tests remain untouched. What do You think about it?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,8 @@ include = [
 ]
 
 [features]
-std = ["memchr/use_std"]
+alloc = []
+std = ["alloc", "memchr/use_std"]
 nightly = []
 default = ["std"]
 regexp = ["regex"]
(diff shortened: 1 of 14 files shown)


kamarkiewicz:
In `no_alloc` mode we should think about how to handle `format!` macro as it could not allocate String to pass error message. Should I just not to write any message at all or make a small array and `write!` to it?

Geal:
yup, `as _bytes`, is_alpha` and others will be a bit annoying to do. `ErrorStr` is not a big problem, it is just a type used in tests to check custom error codes, so let it stay behind the feature guard.
Could you add a build in the `.travis.yml` so it would test with the alloc feature, but without std?

Geal:
alright, I fixed the tests. I don't know what can be done for `is_alpha` and such, but the `unimplemented` error is clear enough for now

PR #684, opened 2018-02-11
Add new rest_len parser.

kurnevsky:
It can be useful in cases when we have to parse something with fixed size after part with dynamic size (so we just can calculate size of this part). What do you think about such parser?


Geal:
Hello,

I don't understand the use case. If there's a fixed size part after a dynamic size part, can't we just parse the dynamic sized one, then once it's done parse the fixed sized one? Or is it for a format with a terminator token?

kurnevsky:
Hi. We can't if we don't know the size of dynamic part. We only know that at the end we should take N bytes. Here is an example: https://github.com/tox-rs/tox/blob/master/src/toxcore/onion/packet.rs#L280-L285
In this case it's possible to parse using flat_map! with rest! and then take length of slice but it will be less convenient I think.

PR #693, opened 2018-02-18
Improve the many0 documentation

Phlosioneer:
The many0 macro will fail if and only if the embedded parser
does not consume any input. It took me a while to figure this out,
so it makes sense to document it.

Also, this commit adds periods to the existing documentation, and
removes the random "0 or more" below the example.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -216,9 +216,12 @@ macro_rules! separated_nonempty_list_complete {
 }
 
 /// `many0!(I -> IResult<I,O>) => I -> IResult<I, Vec<O>>`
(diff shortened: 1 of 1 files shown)


Geal:
which version of nom do you use?

Phlosioneer:
I use 3.2.1 normally, but I also checked that this behavior still exists with the current master branch. There's also a comment that explicitly states the error can happen.

Phlosioneer:
The checks are failing because rustfmt is failing. Rustfmt is failing because there are some lines in the repo longer than 140 characters, which causes it to return an error. It has nothing to do with my PR.

Geal:
thanks!

PR #696, opened 2018-02-20
Remove unused `use`s from take_until_either_1

CAD97:


CAD97:
I think the only reason CI is failing is #697

Geal:
yes, this is fine to merge, thanks :)

PR #699, opened 2018-02-22
torrc parser has moved

FauxFaux:
..although the parser is written with nom *0.3*, so maybe dropping it would be a better idea anyway.


Geal:
right. But maybe that's a good opportunity to update it to nom 3? it should not be too hard ;)

Geal:
remove in #698

PR #707, opened 2018-03-02
Minor doc fixes

mschmo:
Add syntax highlighting tag to rust snippets and fix relative
markdown URL links.


mschmo:
Whoops, never mind. I see now that those things are there for documentation compilation. Closing.

PR #711, opened 2018-03-05
CI: testing without std, but with alloc feature

kamarkiewicz:
Related to PR #677

Added a build in the `.travis.yml` so it would test with the `alloc` feature, but without `std`.
`alloc` feature is supported only on nightly channel.


Geal:
Thanks!

PR #712, opened 2018-03-08
allow multiple patterns in switch

jrakow:
This allows for `A | B` patterns in `switch`.
Before, this would fail to compile:
```rust
        switch!(take!(4),
        b"abcd" | b"xxxx" => take!(2) |
         b"efgh" => take!(4)
       ) 
```
I have added a small corresponding test case.


Geal:
Thank!s

PR #713, opened 2018-03-08
mark *_s macros as deprecated

### LINKED ISSUE #683: mark the *_s combinators as deprecated
Those combinators are now only alias of the equivalent ones (`tag_s` calls `tag` directly, etc).

jrakow:
Not much to see here.
Fixes #683 .


Geal:
Thanks!

PR #714, opened 2018-03-08
Clippy fixes

jrakow:
This fixes most of clippy's warnings.

The most debatable change is globally allowing `inline_always` and `type_complexity`.
Feel free to not include them.

There is also a small simplification of the float parsers.


jrakow:
Rebased and fixed build without default features.

PR #715, opened 2018-03-14
impl Deref for CompleteStr, CompleteByteSlice

lucasem:
With the impending increased use of `CompleteStr` and `CompleteByteSlice` in 4.0, this should give users (myself included) an easier time with low overhead.

Instead of:

```rust
named!(var<CompleteStr, Item>,
        do_parse!(
            name: ws!( alpha ) >>
            (Item::Var(name.0.to_string()))
        ));
```

We can avoid explicitly getting `name.0` thanks to Deref:

```rust
named!(var<CompleteStr, Item>,
        do_parse!(
            name: ws!( alpha ) >>
            (Item::Var(name.to_string()))
        ));
```


Geal:
Thanks!

PR #720, opened 2018-03-18
Fix nostd

### LINKED ISSUE #473: #[no_std] and cross compilation
hi!

I'm trying to get nom working on an embedded device.

Enabling the `core` feature works as expected natively, but fails unable to find a crate for `core` when cross targeting.

The toolchain and HAL I'm using for the embedded device use https://github.com/hackndev/rust-libcore to provide core, but evidently it has no collections crate.

I'm reaching out to you first to see if you have any intuitions about what's going on here, but at a glance it looks as though nom is consuming `std::collections` as an extern crate because it's in the search path for the compiler?

I can try to repackage `std::collections` in some way so I can pull it in, or build a full cross compiler stage instead of merely using the llvm backend, or have a look at what's involved in reducing the dependency on `collections` for a subset of nom (Which might be ideal for my embedded usecase, since allocating is expensive)

Restioson:
This *should* fix #473 (it has fixed it for me). I didn't add any tests since the issue was that the library was using `std` instead of `core` even in a `no_std` context. 

What I've done is what serde does -- made a `lib` module that is a facade over `std` and reexports various things from there or `core` depending. This had to be public to be used in macros. 

P.S: I apologise if I've done something wrong!

### DIFF
--- a/src/bits.rs
+++ b/src/bits.rs
@@ -45,7 +45,7 @@ macro_rules! bits (
 macro_rules! bits_impl (
   ($i:expr, $submac:ident!( $($args:tt)* )) => (
     {
-      use ::std::result::Result::*;
+      use $crate::lib::std::result::Result::*;
       use $crate::{Context,Err,Needed};
 
       let input = ($i, 0usize);
@@ -90,7 +90,7 @@ macro_rules! bits_impl (
 macro_rules! bits_impl (
   ($i:expr, $submac:ident!( $($args:tt)* )) => (
     {
-      use ::std::result::Result::*;
+      use $crate::lib::std::result::Result::*;
       use $crate::{Err,Needed,Context};
 
       let input = ($i, 0usize);
@@ -157,7 +157,7 @@ macro_rules! bytes (
 macro_rules! bytes_impl (
   ($macro_i:expr, $submac:ident!( $($args:tt)* )) => (
     {
-      use ::std::result::Result::*;
+      use $crate::lib::std::result::Result::*;
       use $crate::{Err,Needed};
 
       let inp;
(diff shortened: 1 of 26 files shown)


Restioson:
Appears build is failing on some combinations of features that I didn't try. I'll try fix this when I have a chance. Sorry!

Geal:
from what I see, it's just a matter of marking some tests as not working in no_std

Restioson:
I'll take a look at which tests are failing because of no std and see if they need to be marked. Otherwise, there could be another feature for enabling the alloc crate which is a subfeature of no std which uses the alloc crate to do heal instead of std.

Restioson:
Apparently travis didn't like my commits, but for whatever reason it's not running them...

Geal:
Hello,

I reworked a bit your PR to fix the doctests in db7f902, but unfortunately I messed the rebasing and your name was removed from the commit :/
I'll try to find a way to put it back, otherwise I'll make sure to credit you in the changelog.

THere were a few issues:
- the doctests: with your modification, they did not appear. Now, for the affected tests, I put the needed features right inside the code
- I tried various combination of compilation features and fixed it for regexp and verbose-errors
- some combinators are only built with alloc
- now `Cargo.toml` specifies which tests need which compilation feature

Restioson:
@Geal the combinators that only built with alloc/nostd were intentional, though? They all used Vec if I remember correctly. Same for doctests I think. I don't really mind if my name is on if or not as long as it works! ☺

Geal:
yes, it was necessary to have those combinators that allocate, but I have in mind alternative versions to which one would pass the output as argument, and avoid custom stuff like this:
https://github.com/Geal/parser_benchmarks/blob/master/http/nom-optimized/src/main.rs#L157-L175

PR #725, opened 2018-03-28
Add a conversion from &str to CompleteStr

myrrlyn:
This commit expands on the addition of `Deref<Target=&str>` and provides a means of downcasting an `&str` into a `CompleteStr`.

This should not be used arbitrarily, but is an excellent convenience for manipulating `CompleteStr`s using `&str` methods. Since the `Deref` work always returns bare `&str`s, even from string slices that are known to be `CompleteStr`, converting back into `CompleteStr` is safe and useful.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -5,6 +5,7 @@ use traits::{AsBytes, AtEof, Compare, CompareResult, ExtendInto, FindSubstring, 
              ParseTo, Slice};
 
 use std::str::{self, CharIndices, Chars, FromStr};
+use std::convert::From;
 use std::ops::{Deref, Range, RangeFrom, RangeFull, RangeTo};
 use std::iter::{Enumerate, Map};
 use std::slice::Iter;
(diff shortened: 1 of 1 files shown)


myrrlyn:
Despite the title, this PR operates on both `CompleteStr` and `CompleteByteSlice`.

src/types.rs
```diff
@@ -23,11 +23,17 @@ impl<'a> From<&'a str> for CompleteStr<'a> {
...
+}
+
 impl<'a> Deref for CompleteStr<'a> {
-  type Target = str;
+  type Target = &'a str;
 
-  fn deref(&self) -> &str {
-    self.0
+  fn deref(&self) -> &Self::Target {
+    &self.0
```
myrrlyn: The value returned from `Deref::deref` is now an `&'_ &'a str`, suitable for reconstruction into a `CompleteStr<'a>` via the `From` impl above. Formerly, the value returned was an `&'_ str` where the reference lifetime was tied to the *newtype wrapper*, not to the *referent str*, which was impossible to unify during `From`.

Geal:
thanks, this looks good :)

PR #726, opened 2018-03-28
Implement Display on CompleteStr

myrrlyn:
This is Rust's preferred way of creating a ToString::to_string impl on a type

Since `impl<'a> From<CompleteStr<'a>> for String` is an orphan impl and thus disallowed, we cannot create a `String::from` function that accepts `CompleteStr`. Furthermore, implementing `Into` directly is bad practice because of the blanket impl in std.

The correct way to serialize a value into a type is not `String::from` or `Into::<String>::into`, but `ToString::to_string` ... which has a blanket impl powered by `Display`.

So this commit implements `Display` on `CompleteStr` in order to create `CompleteStr::to_string`.


Geal:
Thanks for that one too!

Issue #706, opened 2018-03-02
Could not find `vec` in `std`


chris13524:
I'm building a no_std thingie and attempting to use `many1!`. I'm including nom like this: `nom = { version = "4.0.0-beta1", default-features = false, features = ["alloc"] }` so the `alloc` feature should be enabled. I see that `vec` is being added to `std` [here](https://docs.rs/nom/4.0.0-beta1/src/nom/lib.rs.html#368) But this doesn't seem to be working.

kamarkiewicz:
Does the same problem occur when you compile with current master?
(instead of `version` please try `git = "https://github.com/Geal/nom", branch = "master"`)

Geal:
hello, on current master, with the commit 9b12a2d it should work now, can you test?

chris13524:
Hmm, I can't recall how to reproduce this anymore... Probably worth closing and I'll re-open if I come across the issue again :)

PR #730, opened 2018-04-03
Some tests are constrained just on alloc

kamarkiewicz:
Hi! I checked one by one tests tagged by `std` feature to find out which one could be run also in `alloc` mode (as `std` enables `alloc`) to extend code coverage.

Initially, I just tried to investigate #706, which just looks like to work correctly...

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1289,6 +1289,8 @@ macro_rules! recognize (
 mod tests {
   use internal::{Err, IResult, Needed};
   use util::ErrorKind;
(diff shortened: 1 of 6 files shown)


Geal:
awesome, thanks! It looks like I've been overly broad with the `std` feature :)

Geal:
apparently I can move even more tests to depend only on `alloc` if I selectively import `alloc::String`. Like, the `HexDisplay` trait can work in `alloc` :)

PR #731, opened 2018-04-10
take_till1! macro does not bring ErrorKind into scope

myrrlyn:


myrrlyn:
I am very bad at Github :/

Anyway, this branch now pulls ErrorKind into scope and slaps an `#[allow(dead_code)]` on `unify_types` in `alt!` because I found a way to make that not get used somehow

PR #734, opened 2018-04-12
Fix stable build with --no-default-features

glandium:
ba357743 made feature(alloc) dependent on !std && alloc, but its merge
(43ed4e9) more or less undid that.


glandium:
The CI failure is due to rustfmt having changed some formatting, apparently.

Geal:
Thanks

PR #735, opened 2018-04-12
Add protobuf-parser in the list of README crates

tafia:
This crate, extracted from quick-protobuf, is new and is supposed to be used by both rust-protobuf and quick-protobuf.


Geal:
nice project :)

PR #736, opened 2018-04-15
Fix doctest formatting

murarth:
It seems that rustdoc had previously ignored doctests which were not immediately preceded by an empty line and (on 1.25.0 stable) I'm now getting warnings about all such doctests, indicating that a future release of doctest will enable them. This PR adds an empty line before all such doctests, which silences the warnings and enables them (unless they are marked `ignore`).


Geal:
Thanks

PR #737, opened 2018-04-16
Make `InputTakeAtPosition` methods generic over error type

### LINKED ISSUE #732: some combinators force the custom error type to u32
some combinators, specifically `complete` and `cond_reduce` in my case, force the custom error type to `u32` when they are used. i've had a quick look at the macro definitions and it seems that many other combinators also force the error type to `u32`, but not all of them.

demonstration of the problem:

```
fn err(i: &str) -> nom::IResult<&str, u32, bool> {
    Err(nom::Err::Error(error_position!(i, nom::ErrorKind::Custom(false))))
}

named!{test<&str, u32>, complete!(err) }
```

this fails during compilation:
```
error[E0308]: match arms have incompatible types
  --> src/main.rs:8:1
   |
8  |    named!{test<&str, u32>,
   |   _^
   |  |_|
   | ||
9  | ||     complete!(err)
10 | || }
   | || ^
   | ||_|
   | |__expected u32, found bool
   |    match arm with an incompatible type
   |
   = note: expected type `std::result::Result<(&str, u32), nom::Err<&str, u32>>`
              found type `std::result::Result<(&str, u32), nom::Err<&str, bool>>`

```

I'm using nom 4.0.0-beta2, but current master also forces the error type of some combinators.

murarth:


murarth:
More testing may need to be done to ensure that this truly makes all macros generic. I've already noticed (and fixed) a few overlooked macros.

Geal:
Hello,

you should have asked about this before changing all this code, because most of it is there for a very good reason and will not be removed.
The type inference issues were very common in user code, and that's why it took a lot of effort to remove them, through the `unify_type` trick, setting the error type to `u32` for basic combinators, using `Convert`, etc.
I could probably accept the changes to `InputtakeAtPosition`, but not the rest, sorry.

murarth:
I didn't think the whole PR would be accepted immediately. I was looking to start a conversation about balancing increased possibility of type inference failure with the flexibility of generic combinator macros. However, I have since decided not to use `nom` for my project, so I won't argue the point.

I've removed all changes except for the modification of the `InputTakeAtPosition` trait methods.

PR #753, opened 2018-05-13
impl AsRef<str> for CompleteStr

FauxFaux:
This allows a `CompleteStr` to inter-operate with more existing string-y methods, like ones accepting a..

```rust
fn foo<S: AsRef<str>>(string: S) -> ..
```
..which is pretty common, and [(kinda?) recommended by the stdlib](https://doc.rust-lang.org/std/convert/trait.AsRef.html#examples).

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -40,6 +40,12 @@ impl<'a> Display for CompleteStr<'a> {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Geal:
Thanks!

PR #755, opened 2018-05-14
Feature/warnings

myrrlyn:
Silence a missing import error and a dead code warning

It seems that alt! does not always use the unify_types function, though I haven't been able to isolate exactly why.


Geal:
thanks!

PR #756, opened 2018-05-15
Inline Deref for CompleteStr/CompleteByteSlice

xfix:
While upgrading to Nom 4.0.0, I noticed that in the compiled binary, there were plenty of calls Deref method of CompleteStr. Considering this function is pretty much a no-op, gets called automatically when you want to call a `str` method, having to call it confuses compiler optimizations there is no reason for it to compile into an actual call.

An alternative would be for an user to use LTO.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -49,6 +49,7 @@ impl<'a> AsRef<str> for CompleteStr<'a> {
 impl<'a> Deref for CompleteStr<'a> {
   type Target = &'a str;
 
+  #[inline]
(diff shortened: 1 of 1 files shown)


Geal:
hello, did you benchmark this fix? I'd really like to have the perf issues fixed for `CompleteByteSlice` and `CompleteStr`

xfix:
Yes, https://github.com/xfix/mvp

This isn't a synthetic benchmark however, it's a program I'm working on.

PR #757, opened 2018-05-15
Use match instead of comparisons in is_hex_digit example in README

xfix:
This makes it more readable for programmers who don't remember the ASCII table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,10 @@ fn from_hex(input: &str) -> Result<u8, std::num::ParseIntError> {
 }
 
 fn is_hex_digit(c: char) -> bool {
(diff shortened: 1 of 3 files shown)


jamwaffles:
Have you benchmarked this? I'd be interested to see if there's any
performance difference.

xfix:
@jamwaffles This isn't a performance improvement, rather a readability improvement. Additionally, the old version was wrong, as it considered `\u130` to be a hex digit due to cast to `u8`.

Geal:
what do you mean, there are people who don't know their ascii tables? :D

thanks for this!

PR #758, opened 2018-05-15
Reformat code with rustfmt

xfix:
This is stable, 1.26.0, installed using `rustup component add rustfmt-preview` and ran using `cargo +stable fmt`. `struct_field_align_threshold` option was removed from `rustfmt.toml`, as it's not supported on stable.

### DIFF
--- a/benches/arithmetic.rs
+++ b/benches/arithmetic.rs
@@ -4,8 +4,8 @@ extern crate test;
 #[macro_use]
 extern crate nom;
 
-use test::Bencher;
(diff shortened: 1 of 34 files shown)


xfix:
Coverage fell, but I don't think it's an issue for this pull request? It's simply what used to be multiple lines became one, causing coverage to fall.

Geal:
sorry, closing this, because formatting pull requests do not have much value and mess up git blame's output

PR #761, opened 2018-05-15
Disable Travis Cargo cache to improve performance

xfix:


xfix:
With this change: 52 min 22 sec
Without this change: Around 1 hour and 20 minutes

Geal:
hello,
disabling the cache is not really a good option because building all the tools takes a lot of time (that job takes 20mn in your PR). But you had the correct idea, I deleted all the caches, ran one full build, and now CI times are back to normal (10mn full build, 34mn accumulated build time https://travis-ci.org/Geal/nom/builds/381031897 ).
I'll try cargo-tarpaulin to see if it can reduce the build time further

PR #767, opened 2018-05-16
Added spaces to wrap wide columns

tbelaire:
This stops the comments field from requiring horizontal scrolling to read.
This cannot be zero width spaces since that breaks when the link is
hovered.


Geal:
this looks much better, thanks :)

PR #770, opened 2018-05-18
Remove deprecated duplicate macro exports

petrochenkov:
See https://github.com/rust-lang/rust/pull/50143 for more details.


Geal:
Thanks!

PR #774, opened 2018-05-19
Add pcap parser to the "network protocol formats" list

ithinuel:
As stated in the title, this simply add the pcap parser I made to the Readme file.

It may require to check to corresponding box on #14.


Geal:
Thanks!

PR #778, opened 2018-05-24
Allow trailing '|' for switch! macro

### LINKED ISSUE #476: switch! should allow a trailing |
Not sure how that's possible with macros, but supporting a trailing | would be more consistent with case which allows a traling comma.

joshleeb:
Fixes #476 - `switch! should allow a trailing |`

This PR modifies the `switch!` macro to support a trailing `|`. The doc tests have been updated to include the trailing `|`.

The change adds `$(|)*` as a suffix to the macro definition. Ideally we would actually want this to be `$(|)?` to support only 1 or 0 trailing `|` characters. However, this feature is only available in nightly. For reference see the [rust-lang/rust#48075](https://github.com/rust-lang/rust/issues/48075).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Added
 
+- Add support for trailing `|` in `switch!` macro
+
 ### Fixed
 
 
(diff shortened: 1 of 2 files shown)


ExpHP:
Worth noting: `$()?` is stable now.

As I wrote on the linked issue, however, the precedent set by `match` is to have a leading `|` rather than trailing.

joshleeb:
Closing due to inactivity.

PR #783, opened 2018-06-01
Switched to if and loop expression in (fold_)many0

tstorch:
Just a small switch to if expression


Geal:
the `break <return value>` syntax is only useful if the loop is considered as an expression and we store the result somewhere. This will break the combinator's behaviour.

tstorch:
Actually the loop is an expression. As the loop is the last thing inside the curly braces and is not followed by a semicolon it simply takes the role which was previouly held by `ret`...

PR #784, opened 2018-06-01
Minor refactoring of (fold_)many1

tstorch:
Small refatoring of the functions many1 and fold_many1 to make
them easier to comprehend

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -317,42 +317,44 @@ macro_rules! many1(
       use $crate::InputLength;
       let i_ = $i.clone();
       match $submac!(i_, $($args)*) {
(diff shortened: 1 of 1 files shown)


Geal:
same issue as #783, this use of `break <return value>` will break the combinator.
I'm closing these two pull requests, for the following reason: I know you're meaning well, but proposing refactorings of the most complex combinators does not help me much and takes me too much time to review. The changes you propose are not minor, and while you find them easier to read, they transform code that evolved slowly and carefully, to a completely new piece of code that I, as maintainer of the library, am not familiar with.
I would be much happier about pull requests fixing bugs, and in the case of new rust syntax that make the combinators easier to write, at least discussing them in an issue beforehand would be better.

PR #789, opened 2018-06-04
Fix typo in doc comment for take_until

WesleyAC:
Just a small fix.


Geal:
Hello, this was fixed in #788, thanks for the help though :)

PR #791, opened 2018-06-06
Add missing import to bytes macro

Nemo157:
Fixes `` Use of undeclared type or module `Context` `` errors in users of the macro


Geal:
thanks :)

PR #799, opened 2018-06-18
Fix warnings in return_error!().

ProgVal:


Geal:
Hello,
unfortunately, the `unify_types` function is there to trick the compiler's type inference into guessing the right types, so it should stay

ProgVal:
Could we use `#[allow(unused)]`?

ProgVal:
Er, no, I don't think does anything to the compiler if it is not called

Geal:
I just took a look, the `unify_types` trick is not used in that pattern, so it's fine to merge, sorry for the delay

PR #808, opened 2018-07-01
Point iso8601 article to updated docu in repository

badboy:
Hey,

the linked article was kinda outdated and so was iso8601. Thanks to @ccouzens there's now an updated article available, which I imported to the repository, so it's easier to update in the future as well. It's best to point there now.


Geal:
great, thanks :)

PR #812, opened 2018-07-06
Remove weird heading lines in migration guides.

phaazon:
Not sure about the meaning of those lines. If you refuse the PR because
of them being useful, please explain to me; I’m curious!


Geal:
they were there because I was generating documentation manually for separated files, and hoping rustdoc and docs.rs would finally support it, but I'm giving up on that now

PR #821, opened 2018-07-23
Add map_res_err!()

YaLTeR:
This combinator preserves the error returned by the function. Here's my use case (also reflected in the new unit test case):
```rust
#[macro_use]
extern crate nom;
use nom::*;

// Valid values that I'm trying to parse.
#[derive(Debug, Eq, PartialEq)]
enum ValidValue {
    One,
    Two,
    Three,
}

// My custom parse error type.
#[derive(Debug, Eq, PartialEq)]
enum ParseError {
    InvalidValue(u8),
}

// Required for custom errors.
impl From<u32> for ParseError {
    fn from(_: u32) -> Self {
        // Not used.
        unreachable!()
    }
}

// Converts a raw u8 into an enum provided it's within range.
fn parse_valid_value(value: u8) -> Result<ValidValue, ParseError> {
    match value {
        b'1' => Ok(ValidValue::One),
        b'2' => Ok(ValidValue::Two),
        b'3' => Ok(ValidValue::Three),
        x => Err(ParseError::InvalidValue(x)),
    }
}

// Parses one ValidValue.
named!(parse_value<&[u8], ValidValue, ParseError>,
    // map_res!() drops the error returned by parse_valid_value().
    map_res_err!(fix_error!(ParseError, be_u8), parse_valid_value)
);

fn main() {
    // Valid input parses successfully.
    let input = b"1";
    let result = parse_value(&input[..]);
    assert_eq!(result, Ok((&input[1..], ValidValue::One)));

    // Invalid input should contain the root cause.
    let input = b"5";
    let result = parse_value(&input[..]);

    // When using map_res!(), the resulting error contains just MapRes itself and not a list of
    // MapRes followed by the underlying ParseError.
    println!("result = {:#?}", result);
    assert_eq!(
        result,
        Err(
            Err::Error(
                Context::List(
                    vec![
                        (
                            &input[..],
                            ErrorKind::Custom(ParseError::InvalidValue(b'5')),
                        ),
                        (
                            &input[..],
                            ErrorKind::MapRes
                        ),
                    ]
                )
            )
        )
    );
}
```


z2oh:
https://github.com/Geal/nom/pull/821/files#diff-3fece039ec70cda79f2b4e1f4c33562fR97

I think there's a typo here; the line should start with `map_res_err!` instead of `map_res!`.

YaLTeR:
@z2oh Oh, right. Fixed.

YaLTeR:
:heavy_check_mark:

PR #830, opened 2018-08-30
Use docs.rs for reference docs

ciphergoth:
Current reference docs link is a 404; use docs.rs instead.


Geal:
done in b49b68dc7370

PR #835, opened 2018-09-24
Simplify is_hex_digit() in example.

smarnach:
This simplifies the `is_hex_digit()` implementation in the example in the README file by referring to the core method `char::is_digit()`.  It's not really an important change, but I think it puts more focus on the bits of the code that are specific to nom.  (I'm not sure the function is needed at all, since there's also an `is_hex_digit()` built into nom now, but I wanted to keep the change minimal – I'm not really familiar with nom.)


Geal:
it looks nicer, thanks :)

Issue #841, opened 2018-10-08
compilation error for add_return_error in verbose-errors mode


Geal:
cf https://travis-ci.org/Geal/nom/jobs/438574247

Geal:
fixed in 5edc06b

PR #842, opened 2018-10-12
Provide stability guarantees for older Rust versions

hcpl:
## Prerequisites

- related issue number: #841 (fixed along the way, but not the main issue)

## Motivation

It is a good idea to cover a large userbase which include those who don't want/can't update their Rust toolchain (e.g. if they stuck on an old OS and use Rust provided by a system-wide package manager).

Seeing that `nom` is a popular crate, I think the project needs to establish an explicitly stated policy regarding this issue.

## Overview

Pinning the minimum supported Rust version (MSRV) for CI builds is a stability guarantee mechanism used by many popular crates, like [`serde`](https://github.com/serde-rs/serde/blob/master/.travis.yml#L36), [`rand`](https://github.com/rust-random/rand/blob/master/.travis.yml#L66) or [`regex`](https://github.com/rust-lang/regex/blob/master/.travis.yml#L5), so there are positive precedents for adopting this approach.

### Pros

- At least some releases older than current stable have a guarantee to work properly with `nom`.
- Maintainers of libraries that have `nom` in their dependency chain wouldn't need to actively chase down MSRV that can actually build their libraries --- just reading README/.travis.yml is enough (app maintainers, on the other hand, are generally less affected by all this MSRV-hunting because they tend to use relatively new Rusts).

### Middle ground

- Increase of build time due to compilation of build script and processing of additional `#[cfg]` attributes, which should be negligible thanks to simplicity of the build script and `#[cfg]`-processing being fast. Local tests show that the increase fall within the margin of error introduced by OS, so this should not be a problem at all.
- Increase of maintenance burden (can't use new features). Though `nom` 4.1.0 doesn't use anything that requires MSRV newer than 1.20 aside of 128-bit integers, so this may not be an issue too.

### Cons

- Increase of maintenance burden (have to watch out for dependencies that do update their MSRV). This one can become very messy if actual MSRV != documented MSRV for those dependencies (real case: https://github.com/rust-lang-nursery/lazy-static.rs/issues/114).
- +2 builds on Travis CI. From what I can see, the total time increase amounts to 5-6 minutes and the run time increase --- to 2-3 minutes (timing data here are for cached builds).

## Unresolved questions

- Is Rust 1.20 too restrictive for the needs of `nom` (most importantly, in long term)?
- There are different variations on versioning policy regarding MSRV employed by different projects:
  * Most conservative: increase MSRV between major versions --- as a result, `cargo update` will not break builds with caret-versioned `nom`.
  * Conservatively increase MSRV between minor releases --- `cargo update` will not break builds with tilde-versioned `nom`, but can break those with caret ones. "Conservatively" here means updating only to the oldest release that can build, but no more.
  * Track only latest stable minus N version. It is still advised to increase Rust version only between minor releases, as patch versions are, well, for bug-fixing patches.
  * Something else, which I am not aware of or I don't think of as suitable for `nom`.

  Which one `nom` should adopt? As a library developer, I prefer the first one (as written in pro number 2).
- Does the Rust version requirement need to be mentioned anywhere else from README?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rust:
   - nightly
   - beta
   - stable
+  - 1.20.0
 
 env:
   matrix:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,9 @@ default-features = false
 #git = "https://github.com/carllerche/bytes"
 #rev = "a7d38e29"
 
+[build-dependencies]
+version_check = "^0.1"
+
 [package.metadata.docs.rs]
 features = [ "alloc", "std", "regexp", "regexp_macros", "verbose-errors" ]
 all-features = true
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Build Status](https://travis-ci.org/Geal/nom.svg?branch=master)](https://travis-ci.org/Geal/nom)
 [![Coverage Status](https://coveralls.io/repos/Geal/nom/badge.svg?branch=master)](https://coveralls.io/r/Geal/nom?branch=master)
 [![Crates.io Version](https://img.shields.io/crates/v/nom.svg)](https://crates.io/crates/nom)
+[![Minimum rustc version](https://img.shields.io/badge/rustc-1.20+-lightgray.svg)](#rust-version-requirements)
 
 nom is a parser combinators library written in Rust. Its goal is to provide tools to build safe parsers without compromising the speed or memory consumption. To that end, it uses extensively Rust's *strong typing* and *memory safety* to produce fast and correct parsers, and provides macros and traits to abstract most of the error prone plumbing.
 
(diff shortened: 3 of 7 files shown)


hcpl:
Hi!

> this is a good thing to consider, but I'm not sure yet about what we should do here.
> [...]
> I'm not saying no to this, but I have to think a bit about it

Sure, please take your time! It is a pretty tough problem as something has to be sacrificed in any case, so I am willing to wait.

Essentially, the whole issue boils down to shifting the toolchain management responsibility --- either to the `nom` maintainers (by restricting yourselves to a compatible subset of Rust and/or `#[cfg]`ing parts of code) or to its users (by having to update their Rusts).

> cfg tags are another way but this will significantly increase the complexity of building and testing nom.

Yes they will, but only if they are there to begin with. If they are not present --- then the only maintenance burden left is to refrain from using new language/standard library features. And from my observation, nearly all of `nom` is already Rust 1.20-compatible, so this shouldn't be too hard to follow, I think.

Strictly speaking, I will probably fine either way, be it switching to Rust 1.26 to use `nom` 4.1.0 or requiring `=4.0.0` in my `Cargo.toml`. It's just that builds on Travis CI [started](https://travis-ci.org/hcpl/rust-tl-lang/jobs/440636034) to [fail](https://travis-ci.org/hcpl/rust-tl-lang/jobs/440636035) after `nom` 4.1.0 has been released and that was quite an unfortunate surprise for me. And even then, the `.travis.yml` for `nom` only tracks the current stable as its oldest supported version, so I would have to track only the latest stable as well, which is not a good idea for libraries in my opinion.

Geal:
alright, I fixed the conflicts and merged this, it will probably be useful to have an idea of which rust versions should be used, even if it moves between major versions

PR #849, opened 2018-11-13
Fix minor Typo

sombrastudios:
The Generic Type parameters `O` was written in Lowercase by accident


Geal:
here `Some(o)` is meant to be a value, not a type

PR #853, opened 2018-12-01
Improved json test

JoshMcguigan:
This PR improves the readability of the json test by using multi-line string formatting. It also asserts on the output rather than using `println!`. 

One thing I'd like to fix but I'm not sure how, is the input string must end in a null byte '\0` or else the parsing fails. This behavior was carried over from the previous implementation.


Geal:
the solution would probably require the code at https://github.com/Geal/parser_benchmarks/blob/master/json/nom/src/main.rs#L130-L140
ie not using `ws!` for the root parser

PR #854, opened 2018-12-06
count! should reserve Vec size.

archseer:
This is my first rust PR, so I hope this change actually makes sense, but I figured since we know how many elements we're parsing, it would be good to pre-reserve the capacity upfront.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -583,6 +583,7 @@ macro_rules! count(
       let ret;
       let mut input = $i.clone();
       let mut res   = $crate::lib::std::vec::Vec::new();
+      res.reserve($count as usize);
 
       loop {
         if res.len() == $count {


archSeer:
There's also probably more cases where this can be applied.

Geal:
is there any performance improvement? One issue I see here would be that if parsing failed, we allocated the vector for nothing

archSeer:
I haven't benchmarked it, but on larger counts it should probably show a difference since it allocates once instead of reallocating a few times (I don't think it reallocates on each push though, it has some spare capacity).

My use-case is parsing a file where it starts with a count, then followed by N=count elements; and since it's a file and not a stream, I know that it's very likely to succeed.

That said, I need to make an adjustment to this PR, instead of `::new + reserve` I should be using [`::with_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity).

archSeer:
So in my case, I'm writing a VM that's [parsing binary files containing erlang machine code](https://github.com/archSeer/enigma/blob/f834d4862ba5b94dc382d07cb4eb191cbe2727a1/src/loader.rs#L368-L548), so I have to trust the file anyway. But reducing allocations would help speed up parsing large files at runtime.

Geal:
it would be a better idea to have your own version of the combinator for this then

PR #859, opened 2018-12-19
Added a more generic converter for errors

wackywendell:
## Reasoning

`nom` errors can be `Convert`ed already if the `Err` part is convertible, but its nice to be able to convert e.g. `Err<&str, u32>` to `Err<String, u32>` more ergonomically. This makes that possible.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -119,7 +119,7 @@ where
 
 use util::Convert;
 
-impl<I, F, E: From<F>> Convert<Err<I, F>> for Err<I, E> {
+impl<I, H: From<I>, F, E: From<F>> Convert<Err<I, F>> for Err<H, E> {
   fn convert(e: Err<I, F>) -> Self {
     match e {
       Err::Incomplete(n) => Err::Incomplete(n),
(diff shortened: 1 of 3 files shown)


eddyb:
Note that this regressed code like this (between `4.1.1` and `4.2.0`):
```rust
pub fn foobar(input: &[u8]) -> IResult<&[u8], Vec<u32>> {
    bits!(input, length_count!(
        // using 0 because this was reduced from a more complex testcase
        value!(0u8),
        take_bits!(u32, 0usize)
    ))
}
```
That is, it now produces a "cannot infer type for `I`" from inside `length_count!`.

The workaround I found was to move the `length_count!(...)` into a separate function, to force inference (which `bits!` doesn't propagate, for some reason I don't yet understand).

Geal:
@eddyb maybe it's missing a `From` implementation somewhere, between bit parsing base type `(&[u8], usize)` and `&[u8]`. This should get better in nom 5, with functions replacing the underlying macro implementation

eddyb:
It doesn't have enough information to infer, so I don't think more impls will help.
Glad to hear that about nom 5 though!

PR #865, opened 2019-02-16
clean macros with vis identifier

### LINKED ISSUE #838: simplify `named!` macros with the :vis macro arg type
[issue introducing :vis](https://github.com/rust-lang/rust/pull/41012) for visibility identifiers
suggested in [#566](https://github.com/Geal/nom/pull/566#issuecomment-427576300) by @derekdreery

vaffeine:
Hi,

This (hopefuly) closes #838. That issue is only about named! macro, but I also fixed the same issue in named_attr! and named_args! (by the way, 2 last variants in named_args! didn't support pub(crate)). I didn't touch macros in methods.rs as they are marked as deprecated.

I updated documentation a little bit to run doc-tests on different forms of named! and named_args! instead of just ignoring. Also I compiled and ran test in 2 crates that uses nom 4: der_parser and rusticata-macros. They work fine.

As a drawback - minimal rustc version grows up to 1.30, which makes pipeline jobs with rustc 1.24 fail. But I guess I could rebase this PR after the minimal rustc version for nom would be updated.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -80,102 +80,64 @@ macro_rules! closure (
 /// The type can be set up if the compiler needs
 /// more information
 ///
-/// ```ignore
+/// Function-like declaration:
+/// ```
+/// # use nom::{named, tag};
 /// named!(my_function( &[u8] ) -> &[u8], tag!("abcd"));
-/// // first type parameter is input, second is output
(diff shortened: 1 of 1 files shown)


Geal:
finally merged, thank you for this :)

Issue #836, opened 2018-10-01
Allow tuple destructuring in `do_parse!`


z2oh:
I have a lot of parsers that return tuples and when using these parsers in `do_parse` I find myself always wanting to destructure them something like this:

```
do_parse!(
    (x, y): parse_point >>
    ( Point { x, y } )
)
```

I'd be happy to give this a shot myself if there is interest from the maintainers.

Geal:
hello,
I tried a while ago and did not succeed, but do not remember why it did not work. If you want to try it out, feel free :)
If you implement it, there are two more issues we can fix: #459 and #560

z2oh:
Ah, I should've checked for pre-existing issues!

I'll give it a try, but I imagine your difficulties implementing it will be mine as well. We will see! :)

z2oh:
@Geal I have been thinking on this and I see a few potential ways of going forward. Tuple destructuring is performed with a pattern as the left hand side and a expression evaluating to a tuple on the right hand side. In a macro match arm, a `pat` cannot be followend by a colon, which breaks with `do_parse`'s syntax. (I'm guessing this is the issue that you ran into when implementing this yourself).

```
do_parse!(
    (x, y): returns_tup >>
    ^ pattern
    (...)
);
```

However, all hope is not lost: I have three potential solutions to this problem. They are all demo'd in [this playground link.](https://play.rust-lang.org/?gist=f38bb8636f180fa0b3fbf320cf0cee3b&version=stable&mode=debug&edition=2015)

---

**Solution 1**: Emulate tuple destructuring by manually parsing out `ident`s from between a set of parentheses. This is demonstrated by the `multiple_arity_tuples` macro in the above playground link.

Pros:
* current syntax does not change
* user experience is identical to tuple destructuring

Cons:
* can only support tuples up to some arbitrary arity (this is common in std lib, arity 12 seems like a good stopping point: https://doc.rust-lang.org/std/primitive.tuple.html )
* lots of boilerplate code to maintain in the `do_parse` macro

**Solution 2**: Use a symbol other than `:` to assign `do_parse` fields. I propose the `=` symbol; this is demonstrated by the `real_destructuring` macro in the playground link.

Pros:
* allows for arbitrary patterns on the left hand side; real tuple destructuring
* ability to apply other forms of destructuring (struct destructuring) for free

Cons:
* syntax change (both could be supported, but that's a con in its own right)
* the `:` token feels ergonomic, as it is the token used for fields of a struct

[...]

PR #869, opened 2019-02-27
 Adding tuple and struct destruction in do_parse 

adamAndMath:
Fixes: [ #836](https://github.com/Geal/nom/issues/836)

This PR modifies `do_parse` to allow tuple destruction, struct destruction as well as nested destructors.

This change wraps fields and struct destructors in parentheses that can then enable full pattern destructions. This disallows `field: type` as well as `some::path(args)` but allows all other patterns. To achieve a disallowed pattern, simply wrap it in parenthesis.


src/sequence.rs
```diff
@@ -950,6 +998,38 @@ mod tests {
     assert_eq!(length_value(&b[..]), Err(Err::Incomplete(Needed::Size(5))));
   }
 
+  #[cfg_attr(rustfmt, rustfmt_skip)]
+  //#[test]
```
z2oh: Is this supposed to be commented out?

z2oh:
I'm looking forward to testing this out tomorrow! This is a feature I've been wanting for a long time 🥇 

Hopefully @Geal will get some time to review this soon!

src/sequence.rs
adamAndMath: Ah. I commented it out to get more compile errors while debugging. Didn't mean to commit it

PR #870, opened 2019-03-01
Clean up documentation

cjbassi:
Hi, I did some various documentation cleanups, mainly with the `choosing_a_combinator` document. There's still more to do but this is a start.

Also, is there a particular reason the documentation was moved from the wiki to the `doc` folder? Having them in the wiki might help more people get involved with the docs. Or maybe there's a way to integrate the docs with the documentation hosted at `docs.rs/nom`? I'm not too familiar with rust documentation tooling so idk if that would work, but it would be a good way to consolidate the docs.


src/lib.rs
```diff
@@ -1,7 +1,7 @@
 //! # nom, eating data byte by byte
 //!
 //! nom is a parser combinator library with a focus on safe parsing,
-//! streaming patterns, and as much as possible zero copy.
+//! streaming patterns, and as minimal copying as possible.
```
Geal: "zero copy" is actually an important word here, so it should not be removed

Geal:
documentation was moved from the wiki because people cannot contribute to the wiki if they're not part of the project. It's easier with pull requests that can be reviewed.
There's a way to integrate external doc files in rust documentation, but I'll have to create dummy modules for that

src/lib.rs
cjbassi: Reverted

Geal:
I merged the docs change, thanks

PR #871, opened 2019-03-02
fix regression in 4.2.1

lucasem:
The following compiles with 4.2.0 but not 4.2.1.

```rust
use nom::{alpha, types::CompleteStr};
use nom::{do_parse, named};

named!(
    foo<CompleteStr<'_>, CompleteStr<'_>>,
    do_parse!(x: alpha >> (x))
);
```

This is because do_parse in 4.2.1 expects the `call!` macro to be defined.

### DIFF
--- a/src/sequence.rs
+++ b/src/sequence.rs
@@ -301,7 +301,7 @@ macro_rules! delimited(
 /// # }
 /// ```
 ///
-#[macro_export]
+#[macro_export(local_inner_macros)]
 macro_rules! do_parse (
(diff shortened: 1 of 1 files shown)


lucasem:
I'd add tests, but this only works with the 2018 edition

Geal:
hello, thank you for the report. The fix was not complete though: I had removed the `local_inner_macros` argument because `do_parse` uses the `compile_error` macro from std, which is not a local one in nom. So I had to add a local wrapper to it.
macro import in edition 2018 continues to be a mess.

The fix is released in nom 4.2.2

PR #889, opened 2019-03-31
Set minimal Rust version to 1.26.2 to compile 'criterion.rs'

LexLuengas:
The test runs on Travis CI didn't compile anymore because _criterion.rs_ only compiles with a Rust version greater than 1.26.2.


Geal:
the minimal version will be 1.28 in 5.0

Issue #914, opened 2019-04-22
nom 5 code level documentation


Geal:
this new version introduces a lot more combinators and parsers:
- the existing macros combinators
- the streaming related function combinators (usually in `src/<module>/streaming.rs`)
- the complete input related function combinators (usually in `src/<module>/complete.rs`)
- the more general combinators (usually in `src/<module>/mod.rs`)

So in a lot of cases, we have three versions: macro, streaming and complete. Users should be able to understand quickly which one they need, without being overwhelmed by their number (I'm in the process of removing some of them).

Here is the format I'd like for most documentation entries:

```
/// Recognizes one character  <-- short, one line description for the function or macros list
///
/// the input data will be compared to the char combinator's argument, and it will return a `char` if successful <-- longer description if the function is confusing
///
/// *streaming version*: will return `Err(nom::Err::Incomplete(_))` if there's not enough input data <-- additional information for streaming or complete version. For complete: *complete version*: will return an error if there's not enough input data
///
/// ```rust
//  # <-- use this syntax to hide the boilerplate
/// # #[macro_use] extern crate nom;
/// # use nom::{Err,error::ErrorKind};
/// use nom::character::streaming::char;
/// # fn main() {
/// let parser = |s: &str| {
///   char('a')(s);
/// };
///
/// short examples with assertions for main cases:
/// - valid input
/// - invalid input
/// - what happens if the parser reaches the end of input (different behaviour for streaming/complete)
/// 
/// assert_eq!(match_letter_a("abc"), Ok(("bc", 'a')));
/// assert_eq!(match_letter_a(b"123cdef"), Err(Err::Error(error_position!(&b"123cdef"[..], ErrorKind::Char))));
/// assert_eq!(match_letter_a(""), Err(Err::Incomplete(Needed::Size(1)));
/// # }
/// ```
```

A lot of macros are already documented, but not all of them, and their docs are often confusing, so they need a fresh eye.

I added `#![warn(missing_docs)]` and `#![warn(missing_doc_code_examples)] `, so to know where to look, clone the repo, launch a compilation and look for the warnings :)

lu-zero:
Could you please make a project with small issues and tick-boxes ? ^^

Geal:
I felt the itch to do it but did not have the time to write it
unfortunately :(

lu-zero:
I guess I can help on that then :P

GuillaumeGomez:
[...]

PR #916, opened 2019-04-22
Add documentation examples for character/complete

GuillaumeGomez:
Part of #914.

### DIFF
--- a/src/character/complete.rs
+++ b/src/character/complete.rs
@@ -1,6 +1,6 @@
-//! character specific parsers and combinators, complete input version
+//! Character specific parsers and combinators, complete input version.
 //!
-//! functions recognizing specific characters
+//! Functions recognizing specific characters.
 
 use internal::{Err, IResult, Needed};
 use error::ParseError;
@@ -9,9 +9,21 @@ use traits::{AsChar, AtEof, FindToken, InputIter, InputLength, InputTakeAtPositi
 use traits::{Compare, CompareResult};
 use error::ErrorKind;
 
-/// recognizes one character
+/// Recognizes one character.
 ///
-/// *complete version*: will return an error if there's not enough input data
+/// *complete version*: Will return an error if there's not enough input data.
+///
+/// # Example
+///
+/// ```
+/// # use nom::{Err, error::ErrorKind};
+/// # use nom::character::complete::char;
+/// # fn main() {
+/// assert_eq!(char::<_, (&str, ErrorKind)>('a')("abc"), Ok(("bc", 'a')));
+/// assert_eq!(char::<_, (&str, ErrorKind)>('a')("bc"), Err(Err::Error(("bc", ErrorKind::Char))));
+/// assert_eq!(char::<_, (&str, ErrorKind)>('a')(""), Err(Err::Error(("", ErrorKind::Char))));
+/// # }
+/// ```
 pub fn char<I, Error: ParseError<I>>(c: char) -> impl Fn(I) -> IResult<I, char, Error>
 where
   I: Slice<RangeFrom<usize>> + InputIter,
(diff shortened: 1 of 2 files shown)


src/character/complete.rs
```diff
@@ -9,9 +9,21 @@ use traits::{AsChar, AtEof, FindToken, InputIter, InputLength, InputTakeAtPositi
...
-/// *complete version*: will return an error if there's not enough input data
+/// *complete version*: Will return an error if there's not enough input data.
+///
+/// # Example
+///
+/// ```
+/// # use nom::{Err, error::ErrorKind};
+/// # use nom::character::complete::char;
+/// # fn main() {
+/// assert_eq!(char::<_, (&str, ErrorKind)>('a')("abc"), Ok(("bc", 'a')));
```
Geal: It's not a good idea to write `char::<_, (&str, ErrorKind)>` everywhere, it makes the example hard to read. If type inference on the error type is an issue, it should be better to write:
```rust
fn parser(input: &str) -> IResult<&str, &str> {
  char('a')
}
```

GuillaumeGomez:
Updated.

GuillaumeGomez:
So after some investigations, I'm wondering if this isn't coming from rustdoc... For records, the error output is:

```
---- src/character/complete.rs - character::complete::alphanumeric1 (line 770) stdout ----
thread 'src/character/complete.rs - character::complete::alphanumeric1 (line 770)' panicked at 'test executable failed:

thread 'main' panicked at 'not yet implemented: error[E0658]: use of unstable library feature 'core_char_ext': the stable interface is `impl char` in later crate (see issue #32110)', src/traits.rs:258:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: <char as nom::traits::AsChar>::is_alpha
   9: <char as nom::traits::AsChar>::is_alphanum
  10: nom::character::complete::alphanumeric1::{{closure}}
  11: <&str as nom::traits::InputTakeAtPosition>::split_at_position1_complete::{{closure}}
  12: core::iter::traits::iterator::Iterator::find::{{closure}}
  13: core::iter::traits::iterator::Iterator::try_for_each::{{closure}}
  14: core::iter::traits::iterator::Iterator::try_fold
  15: core::iter::traits::iterator::Iterator::try_for_each
  16: core::iter::traits::iterator::Iterator::find
  17: <&str as nom::traits::InputTakeAtPosition>::split_at_position1_complete
  18: nom::character::complete::alphanumeric1
  19: rust_out::main::parser
  20: rust_out::main
  21: std::rt::lang_start::{{closure}}
  22: std::panicking::try::do_call
  23: __rust_maybe_catch_panic
  24: std::rt::lang_start_internal
  25: std::rt::lang_start
  26: main

', src/librustdoc/test.rs:332:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: <rustdoc::test::Collector as rustdoc::test::Tester>::add_test::{{closure}}
   8: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
   9: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
  10: __rust_maybe_catch_panic
  11: test::run_test::run_test_inner::{{closure}}
```

Funny thing is that it works perfectly fine when doing the same thing in a binary alongside. I recall you, @QuietMisdreavus working on rustdoc's test handling recently. Do you think this is coming from the same part?

QuietMisdreavus:
Looking at it a bit deeper, it looks like that builder is building `nom` without the `alloc` feature, which causes the `is_alpha` function called by those doctests to trip this assertion:

https://github.com/Geal/nom/blob/c003a2102bb2a4d18d80f529d86bba6e214eccc6/src/traits.rs#L258-L260

Should those functions be gated on the `alloc` feature?

GuillaumeGomez:
As I understood, @Geal said that I wasn't doing something correctly on my side, but maybe you're right. I'll go for that for the moment and then we'll see. Thanks!

Geal:
merged in 2ad650940545

PR #918, opened 2019-04-23
[nom5] Add some more combinators

Keruspe:
Add the flat_map and map_opt combinators which are currently missing


Keruspe:
I'll add some tests and/or doc if you point me to the correct place to do so for nom5 :)

Note that *this* flat_map is an actual fmap (functor map), and is not a port of the former flat_map macro which wasn't.

Keruspe:
Ok, ported the old `flat_map!` under a new name, tell me what you think of it.
More details from the commit message, ftr:

```
This is a port of the old flat_map! macro.
This macro was ill-named, this should better describe what it does.
In case we do not consume all the output of the first parser,
we now error out instead of silently discarding the rest
```

PR #928, opened 2019-04-28
Added documentation for nom 5.0 bytes/mods.rs

Waelwindows:
This pull request ticks the bytes section of #915. Please tell me if there's anything that needs changing/fixing

Issue #915, opened 2019-04-22
Undocumented nom 5.0 functions


lu-zero:
Here the list of all  the functions missing documentation helpfully split per-file.

(I let @Geal format/split it further)

 - src/branch/mod.rs
   - [x] `src/branch/mod.rs:pub fn or<I: Clone, O, E: ParseError<I>, F, G>(f: F, g: G) -> impl Fn(I) -> IResult<I, O, E>`
   - [x] `src/branch/mod.rs:pub fn alt<I: Clone, O, E: ParseError<I>, List: Alt<I,O,E>>(l: List)  -> impl Fn(I) -> IResult<I, O, E> {`
- src/bytes/complete.rs
   - [x] `src/bytes/complete.rs:pub fn tag<'a, T: 'a, Input:'a, Error: ParseError<Input>>(tag: T) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn tag_no_case<T, Input, Error: ParseError<Input>>(tag: T) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn is_not<T, Input, Error: ParseError<Input>>(arr: T) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn is_a<T, Input, Error: ParseError<Input>>(arr: T) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn take_while<F, Input, Error: ParseError<Input>>(cond: F) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn take_while1<F, Input, Error: ParseError<Input>>(cond: F) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn take_while_m_n<F, Input, Error: ParseError<Input>>(m: usize, n: usize, cond: F) -> impl Fn(Input) -> IResult<Input, Input, Error>`
   - [x] `src/bytes/complete.rs:pub fn take_till<F, Input, Error: ParseError<Input>>(cond: F) -> impl Fn(Input) -> IResult<Input, Input, Error>`
[...]

PR #937, opened 2019-05-02
Adding documentation for bytes

Waelwindows:
This pull request is a continuation of #928 which now adds documentation for both the streaming, and complete version of bytes (#915), with appropriate documentation when there's a difference. There's also been a few stylistic differences  from #928

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -16,9 +16,12 @@ use traits::{Compare, CompareResult, FindSubstring, FindToken, InputIter, InputL
 /// # Example
 /// ```rust
 /// # #[macro_use] extern crate nom;
-/// # use nom::{Err, error::ErrorKind, Needed};
+/// # use nom::{Err, error::ErrorKind, Needed, IResult};
 /// use nom::bytes::complete::tag;
-/// let parser = |s: &'static str| tag::<_, _, (_, ErrorKind)>("Hello")(s);
+///
+/// fn parser(s: &str) -> IResult<&str, &str> {
+///   tag("Hello")(s)
+/// }
 ///
 /// assert_eq!(parser("Hello, World!"), Ok((", World!", "Hello")));
 /// assert_eq!(parser("Something"), Err(Err::Error(("Something", ErrorKind::Tag))));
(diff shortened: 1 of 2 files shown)


Waelwindows:
I think I should rewrite the test for `escaped_transform` because it keeps tripping up the testing, and it uses the macro instead. However, I'm not sure on how to rewrite it though

src/bytes/complete.rs
```diff
@@ -46,6 +82,25 @@ where
   }
 }
 
+/// Parse till certain characters are met
+///
+/// The parser will return the longest slice till one of the characters of the combinator's argument are met.
+///
+/// It doesn't consume the matched character,
+///
+/// It will return a `Err::Incomplete(Needed::Size(1))` if the pattern wasn't met
```
Geal: since we're in the complete version, it should return an error, not `Incomplete`

src/bytes/complete.rs
```diff
@@ -183,6 +390,27 @@ where
...
+///
+/// # Example
+/// ```
+/// # #[macro_use] extern crate nom;
+/// # use nom::{Err, error::ErrorKind, Needed};
+/// # use nom::character::complete::digit1;
+/// use nom::bytes::complete::escaped;
+/// use nom::character::complete::one_of;
+/// let esc = |s: &'static str| {
+///   escaped::<&str, (_, ErrorKind), _, _, _, _>(digit1, '\\', one_of("\"n\\"))(s)
```
Geal: maybe write the function like this to avoid all those generic parameters:
```rust
fn esc(s: &str) -> IResult<&str, &str> {
  escaped(digit1, '\\', one_of("\"n\\"))(s)
}
```

Geal:
remove the test for `escaped_transform`, it can go in another PR. I'll take a look at it, I'm pretty sure I introduced a regression when I rewrote it (it looks like an infinite loop triggered by the test)

Geal:
thank you! There's still one CI variant that does not pass, but I'll fix that one

Geal:
I found the issue, the examples should use `alpha1` (that requires at least one valid char) instead of `alpha0`. I added them back in 5d7e5dd

PR #938, opened 2019-05-03
[5.0] Add doc for sequence/mod.rs

NamsooCho:
Add documentation for sequence/mod.rs
It is related with #915 .

### DIFF
--- a/src/sequence/mod.rs
+++ b/src/sequence/mod.rs
@@ -153,6 +153,39 @@ where
   terminated(first, second)(input)
 }
 
+/// Gets an object from the first parser,
+/// then matches an object from the sep_parser and discards it,
+/// then gets another object from the second parser.
+/// # Arguments
+/// * `first` The first parser to apply.
+/// * `sep` The separator parser to apply.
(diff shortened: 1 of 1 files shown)


src/sequence/mod.rs
```diff
@@ -6,6 +6,34 @@ mod macros;
...
+/// use nom::bytes::complete::tag;
+/// # fn main() {
+/// let first_parser = |s: &'static str| {
+///   tag::<_, _, (_, ErrorKind)>("abc")(s)
+/// };
+/// let second_parser = |s: &'static str| {
+///   tag::<_, _, (_, ErrorKind)>("efg")(s)
+/// };
+/// let parser = |s: &'static str| {
+///   pair::<_, _, _, (_, ErrorKind), _, _>(first_parser, second_parser)(s)
```
Geal: maybe write the function like this instead:
```rust
fn parser(s: &str) -> IResult<&str, (&str, &str)> {
  pair(first_parser, second_parser)(s)
}
```
All those generic parameters will scare users a bit

Geal:
thanks

Issue #945, opened 2019-05-19
number::complete::recognize_float fails for "12.34"


frehberg:
rustc 1.36.0-nightly (963184bbb 2019-05-18)
nom version = "5.0.0-beta1"

the function recognize_float fails to parse simple '12.34'

I expected it would recognize '12.34'

Reproduce:
 
I added the following doc-test to latest clone of nom-repo

```
diff --git a/src/number/complete.rs b/src/number/complete.rs
index c8b5c64..b702eb8 100644
--- a/src/number/complete.rs
+++ b/src/number/complete.rs
@@ -738,6 +738,7 @@ pub fn hex_u32<'a, E: ParseError<&'a [u8]>>(input: &'a[u8]) -> IResult<&'a[u8],
 ///
 /// assert_eq!(parser("11e-1"), Ok(("", "11e-1")));
 /// assert_eq!(parser("123E-02"), Ok(("", "123E-02")));
+/// assert_eq!(parser("12.34"), Ok(("", "12.34")));
 /// assert_eq!(parser("123K-01"), Ok(("K-01", "123")));
 /// assert_eq!(parser("abc"), Err(Err::Error(("abc", ErrorKind::Alt))));
```

Executing in nom directory "cargo test", it yields with error

```
thread 'src/number/complete.rs - number::complete::recognize_float (line 730)' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Err(Incomplete(Size(1)))`,
 right: `Ok(("", "12.34"))`', src/number/complete.rs:14:1
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

', src/librustdoc/test.rs:341:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test src/number/complete.rs - number::complete::recognize_float (line 730) ... FAILED

```

frehberg:
AFAICS, related to #910

Geal:
hello, this should work correctly with f891704, I'll soon publish a beta2 including it

frehberg:
Thanks for porting the function number::complete::recognize_float to the new function-style, latest patches seem to solve this issue, please see the regression test as PR

Geal:
great :)

PR #947, opened 2019-05-25
#945 regression test

frehberg:
This is the regression test for issue #945, testing primitive floats such as 12.34 are parsed correctly for complete and streaming case


Geal:
those are documentation tests, the regression tests should be in the `tests` submodule. And I already added them when I fixed the issue :)
https://github.com/Geal/nom/blob/master/src/number/complete.rs#L1184
https://github.com/Geal/nom/blob/master/src/number/streaming.rs#L1185

PR #955, opened 2019-06-05
Add S-expression parsing example

### LINKED ISSUE #952: Request - simple lisp parsing example
This would be helpful.

MarkMcCaskey:
resolves #952

Hello!

Thanks for taking the time to review this!  Let me know if there are any changes I can make to improve this example -- I'm happy to make any changes.

### DIFF
--- a/examples/s_expression.rs
+++ b/examples/s_expression.rs
@@ -1,0 +1,223 @@
+extern crate nom;
+
+use nom::{
+  branch::alt,
+  bytes::complete::tag,
+  character::complete::{alpha1, char, digit1, multispace0},
+  combinator::{map, opt},
+  error::context,
+  multi::many0,
+  sequence::{delimited, preceded, tuple},
+  IResult,
+};
+use std::str;
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum BuiltIn {
+  Plus,
+  Minus,
+  Times,
+  Divide,
+  Equal,
+  Not,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum Atom {
+  Num(i32),
+  Keyword(String),
+  Boolean(bool),
+  BuiltIn(BuiltIn),
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum Expr {
+  Constant(Atom),
+  Application(Box<Expr>, Vec<Expr>),
+  If(Box<Expr>, Box<Expr>),
+  IfElse(Box<Expr>, Box<Expr>, Box<Expr>),
+  Quote(Vec<Expr>),
+}
+
+fn parse_builtin<'a>(i: &'a str) -> IResult<&'a str, BuiltIn> {
+  alt((
+    map(tag("+"), |_| BuiltIn::Plus),
+    map(tag("-"), |_| BuiltIn::Minus),
+    map(tag("*"), |_| BuiltIn::Times),
+    map(tag("/"), |_| BuiltIn::Divide),
+    map(tag("="), |_| BuiltIn::Equal),
+    map(tag("not"), |_| BuiltIn::Not),
+  ))(i)
+}
+
+fn parse_keyword<'a>(i: &'a str) -> IResult<&'a str, Atom> {
+  map(preceded(tag(":"), alpha1), |sym_str: &str| Atom::Keyword(sym_str.to_string()))(i)
+}
+
+fn parse_bool<'a>(i: &'a str) -> IResult<&'a str, Atom> {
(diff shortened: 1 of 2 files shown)


examples/s_expression.rs
```diff
@@ -0,0 +1,331 @@
...
+  /// '(3 (if (+ 3 3) 4 5) 7)
+  Quote(Vec<Expr>),
+}
+
+/// Continuing the trend of starting from the simplest piece and building up,
+/// we start by creating a parser for the built-in functions.
+fn parse_builtin<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, BuiltIn, E> {
+  // alt gives us the result of first parser that succeeds, of the series of
+  // parsers we give it
+  alt((
```
Geal: you could write it in a simpler way without the "not" case:
```rust
let (i, t) = one_of("+-*/=")(i)?;

match t {
  '+' => Ok((i,  BuiltIn::Plus)),
  '-' => Ok((i,  BuiltIn::Minus)),
  '*' => Ok((i,  BuiltIn::Times)),
  '/' => Ok((i,  BuiltIn::Divide)),
  '=' => Ok((i,  BuiltIn::Equal)),
   _ => { unreachable!() }
}

}
```
That can be isolated in another function, and then you do `alt((builtin_operators, map(tag("not"), |_| BuiltIn::Not)))`.
Not sure that will be the cleaner for this one because of the "no" case, but that could be nicer for the `parse_bool` function

examples/s_expression.rs
```diff
@@ -0,0 +1,331 @@
...
+  map(context("keyword", preceded(tag(":"), cut(alpha1))), |sym_str: &str| {
+    Atom::Keyword(sym_str.to_string())
+  })(i)
+}
+
+/// Next up is number parsing. We're keeping it simple here by accepting any number (> 1)
+/// of digits but ending the program if it doesn't fit into an i32.
+fn parse_num<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, Atom, E> {
+  alt((
+    map(digit1, |digit_str: &str| Atom::Num(digit_str.parse::<i32>().unwrap())),
```
Geal: you can use `map_res` to avoir that unwrap: `map_res(digit1, |digit_str: &str| digit_str.parse::<i32>().map(Atom::Num))`

examples/s_expression.rs
```diff
@@ -0,0 +1,331 @@
...
+    map(preceded(tag("-"), digit1), |digit_str: &str| {
+      Atom::Num(-1 * digit_str.parse::<i32>().unwrap())
+    }),
+  ))(i)
+}
+
+/// Now we take all these simple parsers and connect them.
+/// We can now parse half of our language!
+fn parse_atom<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, Atom, E> {
+  alt((parse_num, parse_bool, map(parse_builtin, |bi| Atom::BuiltIn(bi)), parse_keyword))(i)
```
Geal: `map(parse_builtin, Atom::BuiltIn)` ;)

examples/s_expression.rs
```diff
@@ -0,0 +1,331 @@
...
+  If(Box<Expr>, Box<Expr>),
+  /// (if predicate do-this otherwise-do-this)
+  IfElse(Box<Expr>, Box<Expr>, Box<Expr>),
+  /// '(3 (if (+ 3 3) 4 5) 7)
+  Quote(Vec<Expr>),
+}
+
+/// Continuing the trend of starting from the simplest piece and building up,
+/// we start by creating a parser for the built-in functions.
+fn parse_builtin<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, BuiltIn, E> {
```
Geal: you can avoid the generic `E` parameter, and directly bring the `nom::error::VerboseError` type, since only this one will be used (I made the JSOn example generic to demonstrate how various error types work but it's not necessary everywhere)

MarkMcCaskey:
Great!  Thank you for your feedback!  I'm very surprised I had never seen `map(Constructor)` before -- that's very useful.

I've updated the example to reflect the changes you suggested!

Geal:
awesome! One last thing: you should add `alloc` as feature to the example like this: https://github.com/Geal/nom/blob/master/Cargo.toml#L154
Otherwise the `no_std` CI variant won't pass: https://travis-ci.org/Geal/nom/jobs/545411894

MarkMcCaskey:
Okay!  I updated it to be like the JSON example and squashed the extra commits.  Thanks for the guidance!

MarkMcCaskey:
It looks like the reason CI is failing now is unrelated to my changes

Geal:
thank you for your help :)

Issue #953, opened 2019-06-02
bytes::complete::escaped and escaped_transform return Ok(..) even if the parser given in the first argument fails.


webholik:
```
use nom::{
    bytes::complete::escaped,
    character::complete::{alphanumeric1, char},
    IResult,
};

fn test(i: &str) -> IResult<&str, &str> {
    escaped(alphanumeric1, '\\', char('n'))(i)
}
fn main() {
    println!("{:?}", test("abcd\\nefgh")); // Ok(("", "abcd\\nefgh"))
    println!("{:?}", test("____")); // Ok(("____", ""))
}
```

Giving a non-alphanumeric to `test` in the above code should return an `Err` but it succeeds by matching an empty string. Same thing happens with `escaped_transfrom`.

```
nom = "5.0.0-beta2"
```

Geal:
there's currently no way in `escaped` and `escaped_transform` to enforce the number of parsed characters, and I'm not sure it should be possible have that feature depending on the first argument.
That would probably implies duplicating those combinators

Geal:
in the meantime you can use the `verify` combinator to check the length after parsing

webholik:
> That would probably implies duplicating those combinators

I don't understand why it would. Something like this should work: https://github.com/webholik/nom/commit/48a49335c434b7947aafdb718894465a1e3779a3.

Geal:
oh, right, can you send a PR then? :)

liamdawson:
Gah, this was a breaking change for me, as I deliberately need to accept empty strings in my parsing code. Is the correct alternative to use a branch to detect an empty string?

PR #958, opened 2019-06-13
Fix escaped not returning err on failure (issue #953)

webholik:
If none of the input parsers match, then `escaped` and `escaped_transform` should return an `Err` instead of `Ok("")`


Geal:
thanks!

PR #959, opened 2019-06-13
escaped: Use raw strings in example

Havvy:
I think raw strings read easier than trying to count the number of backslashes and that figuring out the Rust escaping vs. the escape output is confusing currently.


Geal:
thanks

Issue #967, opened 2019-06-22
Panic in convert_error on reaching EOF


Dr-Emann:
## Prerequisites

- Rust version : rustc 1.35.0 (3c235d560 2019-05-20)
- nom version : 5.0.0-beta3
- nom compilation features used: default

## Test case

```rust
use nom::character::complete::char;
use nom::error::{convert_error, VerboseError};
use nom::IResult;

fn main() {
    let input = "";

    let result: IResult<_, _, VerboseError<&str>> = char('x')(input);
    match result.unwrap_err() {
        nom::Err::Error(e) | nom::Err::Failure(e) => {
            eprintln!("{:?}", e);
            eprintln!("{}", convert_error(input, e));
        }
        nom::Err::Incomplete(_) => {
            unreachable!();
        }
    }
}
```

Result:
```
VerboseError { errors: [("", Char('x'))] }
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libcore/slice/mod.rs:2686:10
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: rust_begin_unwind
             at src/libstd/panicking.rs:308
   7: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   8: core::panicking::panic_bounds_check
             at src/libcore/panicking.rs:61
   9: <usize as core::slice::SliceIndex<[T]>>::index
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libcore/slice/mod.rs:2686
  10: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libcore/slice/mod.rs:2543
[...]

PR #970, opened 2019-06-25
Changes to convert_error

Dr-Emann:
This fixes the case where the input is non-empty, but still requires a char at EOF.

This also does some cleanup to speed up `convert_error`
* counting lines directly, rather than collecting into strings
* formatting into the result string, rather than into temporary strings first

#### This introduces two changes in the output of `convert_error`

In case of empty input (or any early EOF), outputs:
```
a b 
    ^
expected 'c' , found EOF
```

Chars are formatted with debug. This does not change the literal line, just the expected line:
```
a b 
    ^
expected 'c' found '\u{0}'
```

## Related Issue
#967

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -140,73 +140,56 @@ where
 #[cfg(feature="alloc")]
 pub fn convert_error(input: &str, e: VerboseError<&str>) -> crate::lib::std::string::String {
   use crate::{
-    lib::std:: iter::repeat,
+    lib::std:: fmt::Write,
     traits::Offset
   };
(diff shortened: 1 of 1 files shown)


Geal:
hello, are some of those changes still applicable after the merge of 653f6f32?

Geal:
I'll close this for now, because it is in conflict with the changes in 653f6f32c5435d74de331 (from PR #1032), but if you want to rebase the changes using memchr (I find them interesting), feel free to reopen

PR #980, opened 2019-07-01
Fix incorrect types in documentation of named! and named_args!.

phaazon:
I didn’t check all occurrences of the `&[u8]` pattern but at least those are fixed.


phaazon:
Current situation [here](https://docs.rs/nom/5.0.0/nom/macro.named.html).

Screenshot 2019-07-01 at 14 02 23

phaazon:
Fixed.

Screenshot 2019-07-01 at 14 03 24

Geal:
thanks

PR #984, opened 2019-07-04
many: make progress when possible for *0 parsers

### LINKED ISSUE #982: many0 doesn't return results for alphanumeric0
- Rust version : rustc 1.35.0 (3c235d560 2019-05-20)
- nom version : 5.0.0
- nom compilation features used: none (default)

I'm having trouble understanding why the test using alphanumeric1 passes but fails when using alphanumeric0

```
extern crate nom;
use nom::{
    character::complete::{alphanumeric0, alphanumeric1},
    multi::many0,
    IResult,
};

fn parser0(s: &str) -> IResult<&str, Vec<&str>> {
    many0(alphanumeric0)(s)
}

fn parser1(s: &str) -> IResult<&str, Vec<&str>> {
    many0(alphanumeric1)(s)
}

#[test]
fn test_parser0() {
    assert_eq!(parser0("abc"), Ok(("", vec!["abc"])));
}

#[test]
fn test_parser1() {
    assert_eq!(parser1("abc"), Ok(("", vec!["abc"])));
}
```

The issue seems to be at src/multi/mod.rs:47 (version 5.0.0) where once all of the valid input has been pushed into acc i1 == i and many0 returns an error.  I would expect to get the acc vector that has one element.  Shouldn't it always return acc since no results is valid?  It looks like many1 uses a similar pattern after finding the first result.

tones111:
This change attempts to resolve #982.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -44,7 +44,11 @@ where
         Err(e) => return Err(e),
         Ok((i1, o)) => {
           if i1 == i {
-            return Err(Err::Error(E::from_error_kind(i, ErrorKind::Many0)));
+            if acc.is_empty() {
+              return Err(Err::Error(E::from_error_kind(i, ErrorKind::Many0)));
+            } else {
+              return Ok((i, acc));
+            }
           }
 
           i = i1;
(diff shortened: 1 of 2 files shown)


Geal:
accepting this behaviour would hide potential bugs, I'd rather document clearly that many0 and many1 have this protection against infinite loops

PR #986, opened 2019-07-07
Fix panic on `convert_error` on EOF

CouleeApps:
`convert_error` tries to print the next char when printing char errors, but if the error is at EOF, the next char does not exist.

Couldn't find an issue for this, and since the fix was trivial, I'm just submitting it directly.


Dr-Emann:
Related to #967 and #970

CouleeApps:
Any update on this? It's a pretty minor change that fixes a panic I get pretty regularly.

CouleeApps:
Alternatively accept #970 which would obsolete this bug

sthiele:
related to #1027

Geal:
closing this, as it was fixed in  653f6f3 (from PR #1032)

Issue #976, opened 2019-06-28
nom 5 error in docs/complete::take_bits missing


leshow:
- Rust version : 1.35
- nom version : 5
- nom compilation features used: none
 

I submitted an issue long ago https://github.com/Geal/nom/issues/768 that was closed, so I'm revisiting it to see if I can write it in nom 5. Essentially it's turning a Vec<u16> into a parser that produces 11 bits at a time. I'm following the docs and came across an example https://docs.rs/nom/5.0.0/nom/bits/fn.bits.html

That contains code that doesn't compile. `nom::bits::complete::take_bits` doesn't exist. I'm happy to fix this and submit a PR, but I don't see the function anywhere-- as a result I can't really see how things fit together.

I assumed it meant `bits::complete::take`, but that doesn't work either:

```rust
use nom::{IResult, bits::{complete::take, bits}};

fn take_11_bits(input: &[u16]) -> IResult<&[u16], u64> {
    bits(take(11usize))(input) // <--- error on take
}
```
error:
```bash
error[E0277]: the trait bound `&[u16]: nom::InputIter` is not satisfied
  --> src/mnemonic.rs:30:10
   |
30 |     bits(take(11usize))(input)
   |          ^^^^ the trait `nom::InputIter` is not implemented for `&[u16]`
   |
   = help: the following implementations were found:
             <&'a [u8] as nom::InputIter>
   = note: required by `nom::complete::take`
```

It looks like it's not possible to make a bit parser that takes more bits than 8. Is that correct?

In my original issue, nom 2 could produce a parser that took a Vec<u8> and produced a parser that spit out a Vec<u16> containing each 11-bit chunk.

PR #992, opened 2019-07-10
Enabled and updated bits module doc tests.

michaelgrahamevans:
This PR addresses issue #976. The examples in the docs for the `bits` and `bytes` functions were outdated and set to `ignore`. I enabled the tests and updated them.

The compiler (1.36.0) was not able to fully infer the types for the `bits` and `bytes` functions so I had to add the turbofish syntax `::<_, _, (_, _), _, _>`. In particular it needed to be told that the `E1` type is a tuple with two values. It doesn't look pretty but it worked. Please let me know if there is a better way to write this!

### DIFF
--- a/src/bits/mod.rs
+++ b/src/bits/mod.rs
@@ -19,14 +19,14 @@ use crate::traits::{Slice, ErrorConvert};
 /// away.
 ///
 /// # Example
-/// ```ignore
+/// ```
 /// # #[macro_use] extern crate nom;
 /// # use nom::IResult;
 /// use nom::bits::bits;
-/// use nom::bits::complete::take_bits;
+/// use nom::bits::complete::take;
 ///
 /// fn take_4_bits(input: &[u8]) -> IResult<&[u8], u64> {
-///   bits( take_bits(4usize) )(input)
+///   bits::<_, _, (_, _), _, _>( take(4usize) )(input)
 /// }
 ///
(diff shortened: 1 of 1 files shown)


kamek-pf:
Oh wow, just stumbled in this.
In my case I have a bunch of things to extract, it ends up looking like this:
```rust
let bits = bits::<&[u8], usize, (_, _), _, _>(take_bits(16usize))(input)?;
let field1 = take_bits::<&[u8], u8, _, (_, _)>(1usize)(bits)?; // <- ? doesn't work here 
let field2 = take_bits::<&[u8], u8, _, (_, _)>(3usize)(bits);
    // ...
```

Using `?` on `take_bits` (which is just an aliased `bits::complete::take`), I get this :
```
let field1 = take_bits::<&[u8], u8, _, (_, _)>(1usize)(bits)?;
   |                                                        ^ the trait `std::convert::From<nom::internal::Err<((&[u8], usize), nom::error::ErrorKind)>>` is not implemented for `nom::internal::Err<(&[u8], nom::error::ErrorKind)>`
```

I hope there's a more ergonomic way to use this

michaelgrahamevans:
You can use `nom::sequence::tuple` like this:
```rust
let (rest, (field1, field2)) = bits::<_, _, (_, _), _, _>(
    tuple((
        take_bits(1usize),
        take_bits(3usize)
    ))
)(input)?;
```

PR #995, opened 2019-07-16
fix many_m_n for n=0

kali:
I don't think other function in multi are susceptible to the same bug (fold_many_m_n uses a for loop).


Geal:
thanks

PR #997, opened 2019-07-17
Add delimited_list parser.

kaj:
Alternates between two parsers delimited by two outer parsers to produce a list of elements.

The parser `delimited_list(pre, item, sem, term)` is similar to `delimited(pre, separated_list(sep, item), term)`, but handles errors differently: Since this parser knows what is supposed to terminate the list, it can report errors from the `item` parser rather than assuming that the list has ended when item parsing fails.

The `pre` parser is mainly added for symmetry, but the `term` parser is needed for differentiating between an error in `item` parsing and the end of the list.  If you prefer a `terminated_list` parser without the `pre`, I could rewrite the PR to do that instead.

## Prerequisites

Documentation and testing example provided.  (The [combinator list](https://github.com/Geal/nom/blob/master/doc/choosing_a_combinator.md) seems to be mainly about macro combinators, so nothing added there.)

Example code repeated here:

```rust
let parser = delimited_list(tag("("), is_a("abcde"), tag(","), tag(")"));

assert_eq!(parser("(a,b,c)"), Ok(("", vec!["a", "b", "c"])));
assert_eq!(parser("(a,b,c,)"), Ok(("", vec!["a", "b", "c"])));
assert_eq!(parser("()"), Ok(("", vec![])));

// This call returns the error from the terminator parser:
assert_eq!(parser("(a!)"), Err(Err::Error(("!)", ErrorKind::Tag))));
// This call returns the error from the item parser:
assert_eq!(parser("(a,!)"), Err(Err::Error(("!)", ErrorKind::IsA))));
```

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -980,3 +980,77 @@ where
 {
   length_value(f, g)(i)
 }
+
+/// Alternates between two parsers delimited by two outer parsers to
+/// produce a list of elements.
+///
+/// Similar to `delimited(pre, separated_list(sep, item), term)`, but
+/// handles errors differently: Since this parser knows what is
+/// supposed to terminate the list, it can report errors from the item
+/// parser rather than assuming that the list has ended when item
+/// parsing fails.
+///
+/// The list is allowed to be empty, and a trailing separator (before
+/// the terminator) is allowed but not required.
+///
+/// # Arguments
+///
+/// * `pre` The opening parser.
+/// * `item` Parses the elements of the list.
+/// * `sep` Parses the separator between list elements.
+/// * `term` The list-terminating parser.
+///
+/// # Example
+///
+/// ```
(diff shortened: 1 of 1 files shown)


kaj:
The CI failures with nightly compiler regarding `cfg(doctest)` seems unrelated as far as I can tell.

kaj:
If the `pre` parser argument is removed, the resulting function should probably be named `separated_list_till` for consistency with existing parser combinators `separated_list` and `many_till`.

Geal:
this looks interesting, but this looks like a case where I would instead the `cut` combinator inside the `item` parser?

kaj:
https://github.com/Geal/nom/commit/03f8eef6770c2bc1f87d75f89f09a010b9499919 removes the need for this.  I'll close the PR.

PR #999, opened 2019-07-18
[nom5] Add new combinator: take_up_to

c0d3d:
Works like `take_until`, except it consumes the pattern, and doesn't include the final pattern in the result.

Story behind this is, I was working on parsing something (duh ...), and found that I was writing this as a combination of other sub-parsers. Could be useful to others as well.

Couldn't think of a better name, so suggestions are welcome :)

Added docs for the new functions, as well as doc tests which are passing.

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -419,6 +419,46 @@ where
   }
 }
 
+/// Returns the longest input slice until it matches the pattern, pattern is consumed.
+///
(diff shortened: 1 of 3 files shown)


Geal:
hello, it is indeed useful, and there was a combinator like that in previous versions that was called `take_until_and_consume`, but I have been reducing the number of special case parsers for nom 5, especially when they can easily be done as a combination of others.
I might make another library to hold that kind of combinators though, I'll ping you if I start making it

c0d3d:
Thanks for the update, I'd be happy to help with that :)

Issue #1000, opened 2019-07-19
flat_map is to strict


adamAndMath:
In nom 5.0 [flat_map](https://docs.rs/nom/5.0.0/nom/combinator/fn.flat_map.html)'s second parameter returns a `Fn`, but it could just as well return a `FnOnce`

Geal:
why do you need it to be a `FnOnce` ?

adamAndMath:
Needed to forward non copy information

Geal:
Could you give me more details so I can understand ? Code example, etc

adamAndMath:
I have a range-like syntax that accepts `x-y` and `x+`, cosponsoring to `x..=y` and `x..` in rust. I don't want to re-pass `x` if the first case fails, so I want to do something like
```rust
flat_map(pair(usize::parse, anychar), |(from, c)|move|i|match c {
    '-' => map(usize::parse, |to|LevelRange::Closed(from, to))(i),
    '+' => Ok((i, LevelRange::Open(from))),
    _ => Err(nom::Err::Error(make_error(i, ErrorKind::Alt))),
})
```
In this case I could copy the value, but I have other similar cases where that isn't possible and there are more variants.

koutheir:
The `*map()` operation itself does not call its given lambda more than once, which makes it enough to use the `FnOnce` bound (i.e., the `Fn` bound is too strict in this context). However, if the `*map()` operation itself is called multiple times (for example, in a sequence), then this might result in the given lambda being called multiple times, which would violate `FnOnce`, and require `Fn`.

adamAndMath:
The difference is that the closure produced in `flat_map` is produced and consumed per call to the parser. So for `flat_map` you could still call it multiple times, unlike `map`

Geal:
with 84d1ac025ba2 the trait bound is `FnMut`, this should be testable soon

Geal:
nom 6 is released now :partying_face:

PR #1001, opened 2019-07-19
Relaxing bounds on flat_map's second input

adamAndMath:
Fixing #1000.

Relaxes bound on `H` from `Fn` to `FnOnce`


Geal:
moving the discussion about changing from `Fn` to `FnOnce` or `FnMut` to the 6.0 milestone as that would imply a breaking change in the API. Please note that that I cannot work on a new major release before at least a few months, but the discussion on this change and the valuation of its impact can continue meanwhile

Geal:
with 84d1ac025ba2 the trait bound is now `FnMut`

Issue #949, opened 2019-05-26
map! prevents side effects in nom 5


ankoh:
Hey all,

I ran into a quite far-reaching issue while migrating to the nom 5 beta.
If I got the diff correctly, you introduced a tiny function call for the map! macro here:
https://github.com/Geal/nom/commit/e6f3c6fcbe23bd456bf5474ead6c249a2b75188e

The argument, however, is a `Fn` closure which prevents mutable borrows.
Is this intended?
I started to use map! for side effects quite often.
I'm regularly scanning over a buffer to update an existing state in-flight via these side effects, effectively leaving the remaining state untouched.

That turned out to be a quite elegant pattern that is no longer working without interior mutability.
Can this argument be lifted to `FnMut`?

- Rust version : rustc 1.35.0 (3c235d560 2019-05-20)
- nom version : 5.0.0-beta1
- nom compilation features used: default

## Test case

Example test case:

```
#[macro_use]
extern crate nom;

use nom::number::complete::{le_u8, le_u32, le_f64};
use nom::IResult;

fn foo() {
  let buffer = &b"a"[..];
  let some_side_effect: f64 = 0.0;
  
  do_parse!(
    buffer,
    fields: le_u32 >>
    count!(|input: &[u8]| -> IResult<&[u8], ()> {
        switch!(input, le_u8,
            // No longer allowed in nom 5 due to Fn closure
            0 => map!(le_f64, |v| { some_side_effect = v })
        )},
        fields as usize
    ) >>
    (())
  );
}
```

ankoh:
After skimming the issues I understand now that this function call is, in fact, a crucial aspect of nom 5. And immutable captures are probably sufficient most of the time.
But I found nom to be a very appealing tool to scan quite primitive wire formats.
And there I oftentimes do not really want to pay for a Vec<UmbrellaEnum> if a side effect per match does the same.

Geal:
yes, I'm not sure switching to `FnMut` will be possible. It's still possible to get the side effect you want in a `Fn` with interior mutability, though. cf https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4eacbb3f86ff67698e6b7596b42b7623

ankoh:
I understand. Yes, that's exactly what I did first but it spread far to many Cells. I ended up folding state copies which unfortunately puts a fair amount of trust in the compiler.

I'd argue that side effects are probably not too uncommon especially in more performance sensitive scenarios.
[...]

PR #1002, opened 2019-07-22
[DISCUSSION] Reducing closure requirement from Fn to FnMut

petrosagg:
I created this PR after having a situation where mutable state while parsing was a good fit for my usecase. I found two previous occurrences of this need in the issue tracker:

* https://github.com/Geal/nom/issues/949

Here you mentioned that switching to `FnMut` would not be possible. I'm wasn't sure if you meant that the current design would break which is how I started changing everything in the project to `FnMut` just to see where it will break. However I reached a point where everything is `FnMut` and all the tests pass. Did I miss something in the process?

Of course using a `Cell` is definitely possible, but it makes the code a messier.

* https://github.com/Geal/nom/issues/898

Here you mentioned that in `fold_many0` we'd have to make `init` `Clone`. But after replacing everything in the project with FnMut an issue didn't seem to come up. Again I'm unsure if I missed something.

Also, I was wondering the same question as @z2oh. Do restricted versions of the combinators have a place in the main project, either in this or a future major release? Or would you rather have this be a separate crate (e.g `nom-sideeffect`)?

As the title suggests, this PR is more to discuss the issue of mutability and let people run tests using the branch rather than changing everything in the project.

### DIFF
--- a/examples/s_expression.rs
+++ b/examples/s_expression.rs
@@ -147,9 +147,9 @@ fn parse_constant<'a>(i: &'a str) -> IResult<&'a str, Expr, VerboseError<&'a str
 ///
 /// Unlike the previous functions, this function doesn't take or consume input, instead it
 /// takes a parsing function and returns a new parsing function.
-fn s_exp<'a, O1, F>(inner: F) -> impl Fn(&'a str) -> IResult<&'a str, O1, VerboseError<&'a str>>
+fn s_exp<'a, O1, F>(inner: F) -> impl FnMut(&'a str) -> IResult<&'a str, O1, VerboseError<&'a str>>
 where
-  F: Fn(&'a str) -> IResult<&'a str, O1, VerboseError<&'a str>>,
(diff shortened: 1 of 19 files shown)


Geal:
that's interesting, I'll have to look for the issues I encountered before. If it works it could make a good change for nom 6 (that's still a complete API change, so major version change).

ebkalderon:
Mentioning #1008 to tie the discussion between that issue and this pull request together.

PaperPlaneSoftware:
I was just writing a new issue requesting this exact feature. Rather than pollute the issue board I decided to give this issue a +1

PeytonT:
I also request this feature. Perhaps the FnMut combinators could be added alongside using a different naming scheme?

Geal:
with 84d1ac025ba2 the trait bound is `FnMut`, this should be testable soon

Geal:
closing this as the move to `FnMut` is already done

PR #1004, opened 2019-07-24
Fix custom_input_types.md

dalance:
I tried to fix custom_input_types.md.
This PR includes two changes.

* Change the sample parser code to function-style of nom5.
* Separate the traits list into "for MyInput" and "for MyItem". I think `AsChar` is a trait for `Item` type.


Geal:
thanks!

PR #1016, opened 2019-08-11
Add missing `use` statement to "Code organization" parser example

JDepooter:
In the "Making a new parser from scratch" document, the sample code in the "Code organization" section does not compile as written. I've added the missing `use` statement required to make the code compile.


Geal:
Thanks

PR #1023, opened 2019-08-26
Update documentation links to point to latest

ottosson:
Links in choosing_a_combinator.md was pointing to an old version of the docs. I've updated them to 5.0.1 and removed the link to `count_fixed` which was removed since 5.0.0.


Hywan:
Actually, we should replace the version number of `latest`, e.g. https://docs.rs/nom/latest/nom/macro.count.html, so that URL are alwasy pointing the last version.

ottosson:
Yes, that's probably a better idea 👍

Geal:
thanks!

PR #1026, opened 2019-09-02
Feature: Parsing using Tendril

dignifiedquire:
I would like to use [tendril](https://github.com/servo/tendril/) instead of `&str` and `&[u8]` for my parsers. This has some nice benefits of reducing allocations, especially when processing more complex things and wanting to support both streaming and complete input.

(Because of the orphan rule (I think that is the one) this needs to be implemented here, or in the tendril repo afaict. )

I have started to implement the necessary traits behind a feature flag, but ran into some issues around lifetimes and the various `Input*` traits. 

When you try to run the tests (`cargo test --features tendrils`) you can see the issue

```
the trait `traits::InputIter` is not implemented for `tendril::tendril::Tendril<tendril::fmt::UTF8>`
```

Which I am unable to fix, due to unconstrained lifetimes on that trait, as `Tendril` does not contain a lifetime, whereas `&str` and `&[u8]` the types for which this is implemented constrain the lifetime. 

I would appreciate input both on if this is something you could imagine merging and how to possibly fix this. I would be also happy to maintain this in a different place if there is a good solution we can find that the compiler accepts.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ default = ["std", "lexical"]
 regexp = ["regex"]
 regexp_macros = ["regexp", "lazy_static"]
 lexical = ["lexical-core"]
-tendrils = ["tendril", "std"]
+tendrils = ["tendril", "std", "rental"]
 
 [dependencies.regex]
 version = "^1.0"
@@ -52,6 +52,10 @@ optional = true
 
 [dependencies.tendril]
 version = "^0.4.1"
+optional = true
+
+[dependencies.rental]
+version = "^0.5.4"
 optional = true
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


Havvy:
Shouldn't the feature name be "tendril" that way just having the tendril dependency enables the feature-set on nom's side?

dignifiedquire:
> Shouldn't the feature name be "tendril" that way just having the tendril dependency enables the feature-set on nom's side?

Unfortunately features that are introduced through optional dependencies, can't specify dependencies, and I need it to trigger the `std` feature.

Geal:
hello! this looks very good and useful, but I don't think it would be a good fit to add this to nom's repository. Orphan rules are an issue, yes, but you could wrap in a new type, ie: `struct MyTendril(Tendril<F, A>)` and implement the traits on it. That way it could live in its own crate.

PR #1034, opened 2019-09-19
Add SystemVerilog parser to parsers list

dalance:
I wrote a SystemVerilog parser with nom.
So I want to add it to parsers list of README.


Geal:
nice!

PR #1041, opened 2019-10-03
Add Err::<(&T, ErrorKind)>::to_owned for T=str and [u8]

kpp:
Hi.

Almost all combinators return `Err<(&str, ErrorKind)>` or `Err<(&[u8], ErrorKind)>`. I added an implementation of Err::to_owned for `&str` and `&u8` to cast `Err<(&str, ErrorKind)>` to `Err<(String, ErrorKind)>` and `Err<(&[u8], ErrorKind)>` to `Err<(Vec<u8>, ErrorKind)>`


Geal:
that will be useful, thanks!

Issue #591, opened 2017-10-06
nom4: the trait `std::error::Error` is not implemented for `nom::internal::Err<&[u8]>`


Keruspe:
It would be great to have that implemented so that it can be used with `try!` easily

Geal:
from what I see in https://doc.rust-lang.org/std/error/trait.Error.html it would not be terribly useful. The `description` method must return a `&str` :/

Keruspe:
Well, in nom3, it was implemented, so we could bubble up some nom::Error alongside other Errors easily.

CAD97:
What would you think about (optionally?) supporting [`failure`](https://crates.io/crates/failure) for `nom::Err`? `Fail` then offers a `compat` fn for converting to `std::Error` if necessary.

On the other hand, there seems to be support to [forget about `Error::description`](https://github.com/rust-lang/rfcs/pull/2230) and just use the `Display` impl because of the limitations of returning `&'static str`.

Even if you don't want to impl Error or Fail for `nom::Err`, due to whatever reasons, it would be beneficial to provide a compat struct with a `From` impl in order to easily stuff it into a "I don't care; something went wrong" so the option is there.

Geal:
this would be useful for nom 5, and a bit easier to implement

whatisaphone:
Update: The `Error::description` soft-deprecation has landed on stable (quite some time ago actually).

https://doc.rust-lang.org/std/error/trait.Error.html#method.description

>Although using it won’t cause compilation warning, new code should use `Display` instead and new `impl`s can omit it.

The `?` operator is great and it would be nice to be able to use it with `nom`.

Stargateur:
the fact that the error type of nom doesn't implement error is REALLY annoying, you can't use `Box<dyn Error>`. Is there a technical problem to that ?

Geal:
it's now implemented with #1043

PR #1043, opened 2019-10-05
Implemented std::error::Error and std::fmt::Display traits for nom::Err

Jibbow:
Related issue: #591 

I came across this issue when I worked with nom. I thought having the `Error` and `Display` trait on `nom::Err` might be very helpful.
When I poked around the code I realized that there already was an old implementation that was commented out. Referring to the mentioned issue it might have been because of `Error::description`. But because this is now deprecated anyway I considered opening a PR.

If the removal of the old `Error` and `Display` implementations had other reasons, feel free to close this PR.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -84,7 +84,6 @@ impl<E> Err<E> {
   }
 }
 
-/*
 #[cfg(feature = "std")]
 use std::fmt;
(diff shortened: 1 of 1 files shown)


Geal:
thanks!

PR #1047, opened 2019-10-10
fix imap repo url

ngortheone:
Hey @Geal 

This is really small PR, just readme update.

Reason:
IMAP repo has moved to https://github.com/djc/tokio-imap, old repo is now marked as archived (see https://github.com/djc/imap-proto)


Geal:
thanks!

PR #1050, opened 2019-10-15
Add AsStr trait for convert_error

dalance:
`convert_error` function takes `&str` only. So if I use an input type except `&str` like `nom_locate::LocatedSpanEx<&str, ()>`, `convert_error` can't be used.

This PR add `AsStr` trait to provide `as_str()`.
Hereby any input type implementing `AsStr` can be used at `convert_error`.


Geal:
hello, I'll close this in favor of b2ed755 which uses a more standard trait. Thanks for the push anyway!

PR #1056, opened 2019-10-29
Fix Needed::Size behaviour in streaming number parsers

### LINKED ISSUE #1055: Streaming non-u8 number parsers request too many bytes
- Rust version: `rustc 1.40.0-nightly (7979016af 2019-10-20)`
- nom version : `5.0.1`
- nom compilation features used: Default features

The primitive streaming parsers in `nom::number::streaming` return `Incomplete(Needed(n))`, where `n` is their full size in bytes, even in cases where fewer bytes are needed.

All `le_*`/`be_*` streaming parsers other than the `u8` and `i8` parsers are affected.

There are many doc tests to check this strange behaviour, so I'm not sure if this is for some reason intended.

## Test case

```
extern crate nom;

fn main() {
  assert_eq!(
    nom::number::streaming::be_i16(&[0x12]),
    Err(nom::Err::<()>::Incomplete(nom::Needed::Size(1)))
  );
}
```

Palladinium:
Closes #1055

### DIFF
--- a/src/number/macros.rs
+++ b/src/number/macros.rs
@@ -12,12 +12,12 @@
 /// named!(be<u16>, u16!(Endianness::Big));
 ///
 /// assert_eq!(be(b"\x00\x01abcd"), Ok((&b"abcd"[..], 0x0001)));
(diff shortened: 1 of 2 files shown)


Geal:
thanks!

PR #1058, opened 2019-10-31
Use once_cell instead of lazy_static

light4:
`once_cell` provides a neat way of initializing lazy singletons without macro.

# references

* RFC: https://github.com/matklad/rfcs/blob/std-lazy/text/0000-standard-lazy-types.md
* crate: https://crates.io/crates/once_cell
* async_std: https://github.com/async-rs/async-std/issues/406


Geal:
with #1042 lazy_static usage has been completely removed, I'll see if it needs to come back

PR #1065, opened 2019-11-09
Updating doc reference to util::dbg_dmp

nyarly:
util isn't exported, but dbg_dmp is `pub use`d, so it can be imported directly from nom.


Geal:
thanks!

PR #1076, opened 2019-11-27
Fix take_while_m_n macro example

guedou:
Thanks for your work on nom!

This PR fixes the take_while_m_n macro example.


Geal:
thanks!

PR #1077, opened 2019-11-27
Add inline attribute to the to_usize functions

adytzu2007:
The nom take parser automatically converts it's argument to usize
using to_usize.

However, if the argument is already usize, this should be a no-op and
the compiler should just skip the call. Surprisingly, during my tests,
I noticed that a large part of the cycles were wasted in
to_usize(usize).

Adding the [inline] attribute fixes it for me.

Example:

Before the change, the parser would take 29,606,796,156 cycles.
After the change, the parser only takes 9,245,008,172 cycles.
This translates in runtime decreasing from 7.6 seconds to 2.8 seconds.

Thanks for nom! It's an awesome library.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1109,22 +1109,26 @@ impl ExtendInto for char {
 /// with 32 and 64 bits pointer platforms
 pub trait ToUsize {
   /// converts self to usize
+  #[inline]
   fn to_usize(&self) -> usize;
 }
 
 impl ToUsize for u8 {
+  #[inline]
   fn to_usize(&self) -> usize {
     *self as usize
   }
 }
 
 impl ToUsize for u16 {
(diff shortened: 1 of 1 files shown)


Geal:
that overhead is surprising, but that would not be the first time we see something like this

Razican:
I seem to be getting a bunch of these warnings when compiling in Rust 1.40.0:
```
warning: `#[inline]` is ignored on function prototypes
    --> src/traits.rs:1112:3
     |
1112 |   #[inline]
     |   ^^^^^^^^^
```

adytzu2007:
Yeah, it seems that #[inline] is not relevant for function prototypes. Will fix it.

adytzu2007:
https://github.com/Geal/nom/pull/1107

PR #1082, opened 2019-12-02
Escaped parsers empty input fix

AnderEnder:
## Issue:
During the migration from nom 4 to nom 5 we found the following regression:
```
many!(escaped_transform(..)) 
```
Produced the `Many1` error, which is produced by indefinite cycle, which was related to `escaped_transform` behaviour, when empty input produces empty output.

## Test cases:
- integration regression test in `issues.rs`
- 2 unit tests in `escaped.rs`

## Solution:
Check the input length  before and return error in **complete** versions.

## Questions:
- Should **escaped** return error only if inner parser return error on empty output?
- Should **stream** versions be fixed in the similar way?

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -452,6 +452,10 @@ where
   use crate::traits::AsChar;
 
   move |input: Input| {
+    if input.input_len() == 0 {
+      return Err(Err::Error(Error::from_error_kind(input, ErrorKind::Escaped)));
+    }
+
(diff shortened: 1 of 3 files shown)


Geal:
escaped parsers should accept an empty input, it is not an error. Why are you using this parser inside a `many1`? Isn't there a separating token somewhere?

Geal:
this has been fixed in d31a752

PR #1088, opened 2019-12-17
Fix length_data docs

courajs:
Looks like it was copy/pasted without changes from `length_value` at some point


Geal:
it was fixed in 47306c4

PR #1091, opened 2019-12-21
Substantially optimized convert_error

Lucretiel:
I've identified and fixed the following issues in the implementation of `convert_error`. These changes do not alter the behavior of this function, only its implementation.

- Substantially overallocates. Virtually every append to `result` involves allocating a new string; this is virtually always unnecessary, since `String` implements `fmt::Write` and can be written to directly with `write!`.
- Too many appends: `String` implements `fmt::Write`, which means that rather than doing many tiny appends, we can simply `write` once directly to it. This simplifies the code and makes it easier to see the expected output in the format spec.
- line/column number calculation is inefficient: by counting newline bytes in the original input string,  we can more quickly find the line number, column number, and line substring.
- Incorrect line numbers: virtually all editors and tools use 1-indexed line numbers. Fix the output to conform to this standard, to prevent confusion.
- Rustfmt: This change rustfmts the error.rs file.

# Correctness notes:

- There are no existing test cases for this function. However, I've confirmed that the examples that use this function produce identical output (except for. the corrected line numbers as mentioned above).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,7 +11,6 @@
 /// It provides methods to create an error from some combinators,
 /// and combine existing errors in combinators like `alt`
 pub trait ParseError<I>: Sized {
-
   /// creates an error from the input position and an [ErrorKind]
   fn from_error_kind(input: I, kind: ErrorKind) -> Self;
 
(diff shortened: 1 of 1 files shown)


Geal:
awesome, thanks :)

PR #1093, opened 2019-12-21
Simplified alt() implementation

Lucretiel:
Simplified alt() to use a match, rather than `if err.is_some() { err.unwrap()...}`. Also deduped the assignments.


Geal:
nice catch, thanks!

PR #1094, opened 2019-12-21
Added example for an escaped string parser

Lucretiel:
Added a comprehensive example for how to parse a more complex escaped string. This parser supports regular escape sequences (`\n`, `\r`, `\t`) as well as rust-style escaped code points (`\u{1F602}`) and escaped whitespace (`\     `). Thoroughly commented.


Geal:
wow that's an awesome example, thanks!

Issue #913, opened 2019-04-21
take_while_m_n fails on multi-byte characters


Recmo:
## Prerequisites

- Rust version : `rustc 1.34.0 (91856ed52 2019-04-10)`
- nom version : `4.2.2`
- nom compilation features used: default

## Test case

Please provide a short, complete (with crate import, etc) test case for
the issue, showing clearly the expected and obtained results.

Example test case:

```rust
use nom::*;
named!(parser<&str, &str>, take_while_m_n!(1, 1, |c| c == 'A' || c == '😃'));

fn main() {
    assert_eq!(parser("A!"), Ok(("!", "A")));
    assert_eq!(parser("😃!"), Ok(("!", "😃")));
}
```

```
thread 'main' panicked at 'byte index 1 is not a char boundary; it is inside '😃' (bytes 0..4) of `😃!`', src/libcore/str/mod.rs:2027:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:70
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:200
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:215
   4: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
             at src/libstd/panicking.rs:478
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:385
   6: std::panicking::try::do_call
             at src/libstd/panicking.rs:312
   7: <T as core::any::Any>::type_id
             at src/libcore/panicking.rs:85
   8: core::str::traits::<impl core::slice::SliceIndex<str> for core::ops::range::Range<usize>>::index::{{closure}}
             at src/libcore/str/mod.rs:2027
   9: core::str::<impl str>::as_ptr
             at /rustc/91856ed52c58aa5ba66a015354d1cc69e9779bdf/src/libcore/str/mod.rs:1877
  10: <core::result::Result<T, E> as core::cmp::PartialEq>::eq
             at /rustc/91856ed52c58aa5ba66a015354d1cc69e9779bdf/src/libcore/option.rs:386
[...]

PR #1097, opened 2020-01-02
Fix take_while_m_n utf8 where all input chars are matching

audunhalland:
## Prerequisites

Related issue: #913 (take_while_m_n fails on multi-byte characters)

`take_while_m_n` panics when all of the input characters match the predicate, and `n` is smaller than the input (string is split in the middle of a multibyte char).

On the other hand: `take_while_m_n` seems to be inefficient when it cannot find a predicate complement. It appears to scan the complete string in that case, even if a string length equal to `n` is found before the end of string. I have not fixed this, just the panic.

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -276,8 +276,10 @@ where
       None => {
         let len = input.input_len();
         if len >= n {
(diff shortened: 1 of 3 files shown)


Geal:
nice catch, thanks!

Issue #1100, opened 2020-01-07
Update Lexical Version


Alexhuszagh:
Hi, and thanks for all the hard work on Nom.

## Prerequisites

Nom uses a dependency I maintain, lexical-core, which used to define a C-API and Rust API within a single crate. Due to symbol issues and to remove unnecessary functionality from lexical-core, the C-API was separated into a new crate. Other bug fixes have since been implemented, for example, in ensuring full standard compliance in parsing weird edge-cases like ". ", among others.

**Issues**

Other uses have reported issues with older versions of lexical:

[Undefined Symbols when building on Windows](https://github.com/Alexhuszagh/rust-lexical/issues/29)
[Symbol Issues when used in a proc macro](https://github.com/Alexhuszagh/rust-lexical/issues/23)
[Undefined Symbols when building on Windows](https://github.com/Geal/nom/issues/1066)
[Incorrect result with parsing ". "](https://github.com/Geal/nom/issues/1080)

**Versions**

To support Rustc versions 1.31+, as Nom currently does, please use `lexical-core = "0.6"` in Cargo.toml, which supports any version of Rustc >= 1.24.0.

**F64 Complete**

The patch for the complete, f64 parser would be:

```rust
#[cfg(feature = "lexical")]
pub fn double<T, E:ParseError<T>>(input: T) -> IResult<T, f64, E>
where
  T: crate::traits::AsBytes + InputLength + Slice<RangeFrom<usize>>,
{
  match ::lexical_core::parse_partial(input.as_bytes()) {
    Ok((value, processed)) => Ok((input.slice(processed..), value)),
    Err(_) => Err(Err::Error(E::from_error_kind(input, ErrorKind::Float)))
  }
}
```

**F64 Streaming**

The patch for the streaming, f64 parser would be:

```rust
#[cfg(feature = "lexical")]
pub fn double<T, E:ParseError<T>>(input: T) -> IResult<T, f64, E>
where
  T: crate::traits::AsBytes + InputLength + Slice<RangeFrom<usize>>,
{
  match ::lexical_core::parse_partial(input.as_bytes()) {
    Ok((value, processed)) => {
      if (processed == input.input_len()) {
        Err(Err::Incomplete(Needed::Unknown))
      } else {
        Ok((input.slice(processed..), value))
      }
    },
    Err(_) => Err(Err::Error(E::from_error_kind(input, ErrorKind::Float)))
  }
}
```

Issue #1080, opened 2019-11-29
double parser returns unexpected result for the input ". " when the lexical feature is enabled


evant:
## Prerequisites

- Rust version : rustc 1.39.0 (4560ea788 2019-11-04)
- nom version : 5.0.1/master
- nom compilation features used: lexical

## Test case

The `number::complete::double` parser returns an unexpected result when passed the input `". "`. I would expect it to fail, but it returns `Ok((" ", 0.0))`. If the lexical feature is disabled, then the expected error of: `Err(Error((" ", Digit))` is returned.

Sample repo: https://github.com/evant/nom-double-parse-issue

```rust
use nom::number::complete::double;
use nom::IResult;

fn main() {
    println!("nom double parse: {:?}", nom_double(". "));
}

fn nom_double(input: &str) -> IResult<&str, f64> {
    double(input)
}
```

Alexhuszagh:
The specific issue is here, where the complete `nom` parser uses lexical (which I maintain) to provide a streaming-like feature detection, which I believe is an issue here. Older versions of lexical detected the first invalid digit, which would raise an InvalidDigit if only a `'.'` character was found (or a '.' character follow by any non-valid digit), which was fairly amibiguous.

My new fork should correct this issue with a more modern version of lexical:
https://github.com/Alexhuszagh/nom-double-parse-issue

Newer versions, which I will shortly submit a PR for, will correctly fail with this input.

``` bash
nom double parse: Ok((" ", 0.0))
lexical_043 double parse: Result { value: 0.0, error: Error { code: InvalidDigit, index: 1 } }
lexical_070 double parse: Err(Error { code: EmptyFraction, index: 0 })
```

Geal:
#1101 is merged, new version coming up

Issue #1066, opened 2019-11-10
Fail to build on windows when used in a dylib


breard-r:
## Environment

rustc 1.39.0
cargo 1.39.0
Windows 10
Visual Studio 2019

## Bug description

If nom is used within a library which uses the "dylib" crate-type (even in conjunction with other types), then every binary having this lib as a dependency will fail to compile on windows (error log below).

### The library (named `what`)

The library's `Cargo.toml` must contain:

```
[lib]
crate-type = ["rlib", "staticlib", "dylib"]

[dependencies]
nom = "^5.0"
```

Also, `lib.rs`, must contain any call to nom:

```
use nom::IResult;
```

There is no need to actually use `IResult`.

### The binary (named `nom_test`)

The `Cargo.toml` must include a dependency to the above library:

```
what = { path = "../what" }
```

## Full compilation error log

```
   Compiling what v0.1.0 (C:\Users\IEUser\Documents\what)
warning: unused import: `nom::IResult`
 --> C:\Users\IEUser\Documents\what\src\lib.rs:1:5
  |
1 | use nom::IResult;
  |     ^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

error: linking with `link.exe` failed: exit code: 1120
  |
[...]

PR #1101, opened 2020-01-07
Update lexical-core version to "^0.6".

Alexhuszagh:
Patches:
- #1100
- #1080
- #1066
- And a few upstream issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ version = "^2.0"
 default-features = false
 
 [dependencies.lexical-core]
-version = "^0.4.0"
+version = "^0.6.0"
 optional = true
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


Razican:
Now that 0.7 has been released, it would be a nice addition to this :)

Alexhuszagh:
@Razican, eventually, yes, however, v0.7 only has support for Rustc >= 1.37, while v0.6 has support for Rustc >= 1.24 (this is due to dependency restrictions). I would like to avoid any breaking changes in Nom currently, but ideally yes, long-term, that would be ideal :D.

ErichDonGubler:
Looks like upgrading to 0.7 is just a bump -- tests still pass. :smile:

Geal:
thank you for this! I'll probably raise the minimum Rust version for nom 6, so it could switch to v0.7

PR #1073, opened 2019-11-26
chore: Update version_check to 0.9

ignatenkobrain:
Also simplify build.rs.

Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

PR #950, opened 2019-05-28
update version_check to 0.9

SergioBenitez:
A simple update to the dependency.

PR #1106, opened 2020-01-14
Upgraded jemallocator dependency

Razican:
This PR just upgrades jemalloc and criterion to the latest supported versions. It does not require any code change, and tests seem to pass. It updates the two missing dependencies after #1101 and #1073 (or #950) get merged.

The API does not change, so this can be released as a minor version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ optional = true
 
 [dev-dependencies]
 criterion = "0.2"
-jemallocator = "^0.1"
+jemallocator = "^0.3"
 doc-comment = "0.3"
 
 [build-dependencies]


Razican:
Issues are related to a new minimum rust version.

ErichDonGubler:
@Geal, ping!

Geal:
since this would rais the minimum rust version, the PR needs to go in nom 6 (although criterion could be updated separately)

Razican:
I rebased and modified the commits to upgrade only the "jemallocator" dependency, that is compatible with Rust 1.31.0, so this could be added to nom 0.5, @Geal.

Geal:
thanks

PR #1107, opened 2020-01-14
Remove inline attribute from prototypes

adytzu2007:
Inline attribute is ignored for prototypes so this change removes it. See https://rust-lang.github.io/rust-clippy/master/#inline_fn_without_body


Geal:
Thanks!

PR #1111, opened 2020-02-05
Make `error::context()` available without `alloc` feature

lo48576:
Currently [`nom::error::context()`](https://docs.rs/nom/5.1.0/nom/error/fn.context.html) is available only when `alloc` feature is enabled.
However, [`Clone` trait is available in nostd build](https://doc.rust-lang.org/stable/core/clone/trait.Clone.html), and `nom::error::context()` would be still useful for no-`alloc` environment (especially when the input type `I` is `&str`).

This pull request makes `context()` available even when `alloc` feature is not enabled.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -114,13 +114,11 @@ impl<I> ParseError<I> for VerboseError<I> {
   }
 }
 
-#[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


Geal:
nice, thanks

Issue #1061, opened 2019-11-05
Function alternative to the `eof!` macro


shnewto:
While updating the [bnf](https://github.com/shnewto/bnf) crate from macros (nom 3.2) to nom 5.1 functions recently, I came up against some logic that used the `eof!` macro. I couldn't find a 1 to 1 function to replace it with and ended up implementing my own. 

Here's what I ended up with:
```rust
pub fn eof<I: Copy + InputLength, E: ParseError<I>>(input: I) -> IResult<I, I, E> {
    if input.input_len() == 0 {
        Ok((input, input))
    } else {
        Err(nom::Err::Error(E::from_error_kind(input, ErrorKind::Eof)))
    }
}

```

I'd be happy to make a PR to add something like this but there may have been a good reason to omit a `pub fn eof` and so I thought I'd check before jumping into anything.

So, I suppose. If there's not going to be a 1 to 1 `pub fn eof` function, what would be a good alternative?

If there could be a `pub fn eof`, would you be open to a PR?

mdorman:
Since I just borrowed this for a project, I think you should just open the PR for it; it seems like an oversight that it wasn't converted.

Geal:
fixed in #1112

PR #1112, opened 2020-02-06
Add an `eof` function

shnewto:
The changes in this PR set out to address issue #1061, and adds a function alternative to the `eof!` macro (as well as tests for that function).

Hopefully this addition will help people who relied on `eof!` transition away from their macros and away from writing their own special case eof.

I am certainly happy to make revisions to these changes, address things you spot that I've missed, and really, would welcome any feedback, if there is some.

Thanks for taking a look! :smile:


shnewto:
Haha, idk why coverage went down because in the file I touched, coverage went up 🙃 

image

If you'd like me to fill in some coverage gaps so this PR doesn't decrease coverage I'd be willing, though that seems more appropriate as separate PR

Lucretiel:
You should make it clear that this does not work on streaming input

shnewto:
@Lucretiel you'll have to pardon me if I'm being naive here :sweat_smile: but I don't know how `eof` would even work w/o complete data, either you're at eof or you're not right? It seems like the convention atm is to put functions that _can_ be either streaming or compete in either `src/bytes/complete.rs`, `src/character/complete.rs`, `src/bytes/streaming.rs`, or `src/character/streaming.rs`. Otherwise they live elsewhere... At least, the only places I've noticed any explicit mentions of whether the function is for streaming v complete data lives in those places. 

If I'm totally off base, I'd love it if you were willing to elaborate :smile:

g-plane:
Are there any updates on this?

calebsander:
Do you have an example of a use case for `eof()` that wouldn't make more sense to implement using `all_consuming()`? It seems like you can always replace `eof()` with `all_consuming(rest)` anyways.

shnewto:
@calebsander for me in particular,`eof` is just one of the possible delimiters for the patterns I'm parsing for. And in those cases, I don't actually want to consume it if it's there (so I use `peek`).

Geal:
Thanks!

PR #1115, opened 2020-02-29
Allow lexical-core 0.7

shepmaster:
[No APIs changed between 0.6 and 0.7](https://github.com/Alexhuszagh/rust-lexical/blob/master/lexical-core/CHANGELOG). The minimum Rust version was increased, but as this PR only *adds* a new allowed version, the previous version can still be used by downstream users of nom that need to maintain compatibility.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ version = "^2.0"
 default-features = false
 
 [dependencies.lexical-core]
-version = "^0.6.0"
+version = ">= 0.6, < 0.8"
 optional = true
 
 [dev-dependencies]


Geal:
oh, I did not know that syntax was available, it's nice, thanks

jonhoo:
Is this actually a breaking change? If it's not, may be worth backporting to `nom 5`, since `lexical 0.6` [now pins `cfg-if 0.1.9`](https://github.com/Alexhuszagh/rust-lexical/blob/8c75e297689591fea903176bcaeb03c27325dccf/lexical-core/CHANGELOG#L47), which fails to compile all sorts of things on newer Rust versions (this is fixed in `cfg-if 0.1.10`). I'll file an issue with `lexical` as well.

jonhoo:
See also https://github.com/Alexhuszagh/rust-lexical/issues/38. One other option (if they don't backport) is to release a new `nom 5` that requires `lexical-core < 0.6.3`.

Issue #1120, opened 2020-03-13
[wishlist] Option to limit recursion depth


kandykoma:
Issue
---------
Deep recursions trigger a stack overflow, that canot be handled.

Consequence
------
Cannot use recursive nom parser on untrusted input, as cannot protect against DoS.

Demo
---
A minimal repro case:
```
fn cp(i: &str) -> nom::IResult<&str, &str> {
    use nom::branch::alt;
    use nom::bytes::complete::tag;
    use nom::sequence::delimited;

    delimited(tag("("), alt((tag("x"), cp)), tag(")"))(i)
}

fn main() {
    let mut buffer = String::new();
    let stdin = std::io::stdin();
    let mut handle = stdin.lock();

    {
        use std::io::Read;
        handle.read_to_string(&mut buffer).expect("read failed");
    }
    println!("{:?}", cp(&buffer));
}
```
With input:
```
perl -e 'print("(" x 10000, "x", ")" x 10000, "\n")' | ./target/debug/main
```

reports:
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted
```
Improvement Option
-----
Intruduce a maximal depth limit, which when reached gracefuly terminates the parsing.

Keruspe:
Why don’t you just have something like cp which calls my_cp(0) and my_cp(n) which calls my_cp(n+1) in the alt?

This way you can throw an error when n is too big without any modification in nom

kandykoma:
Thanks @Keruspe, it works!

Two follow ups:
- is there a better way to trigger an error?
- is there a way to not only signal a matching failure, but trigger an abort of the parsing process?

```
fn cp(n: u32) -> impl Fn(&str) -> nom::IResult<&str, &str> {
    use nom::branch::alt;
    use nom::bytes::complete::tag;
    use nom::sequence::delimited;
    move |i| {
        if n > 10 {
            tag("never matching string")(i)
        } else {
            delimited(tag("("), alt((tag("x"), cp(n+1))), tag(")"))(i)
        }
    }
}

fn main() {
    let mut buffer = String::new();
    let stdin = std::io::stdin();
    let mut handle = stdin.lock();

    {
        use std::io::Read;
        handle.read_to_string(&mut buffer).expect("read failed");
    }
    println!("{:?}", cp(0)(&buffer));
}
```

Geal:
[...]

PR #1125, opened 2020-03-21
Add InputAux wrapper.

kandykoma:
This is a non-invasive extension to give support for a limited form of the feature described in #1120.

The solution itself is a new input type, that is initialized with arbitrary extra data in addition to the input itself. This extra data is then readable by the users during parsing.

This way the administration of recursion depth of the parsers is left to the library users, which adds flexibitlity with the customary built-in footgun.


kandykoma:
crates.io/crate/nom_input_aux

PR #1128, opened 2020-03-27
Update parser readme

chifflier:
Hi @Geal ,

This (trivial) PR does some maintenance on the README file:
- remove Redis, not using nom
- add more references to various parsers

Note that I added my pcap(and pcapng) parser even if there are others, those are not maintained, and do not cover the entire format specifications.


Geal:
great, more parsers :)

PR #1129, opened 2020-04-04
Add `success` combinator returning the given value without consuming

sphynx:
Hi Geoffroy,

This pull request adds a new `success` parser which is basically like `pure` parser in Haskell Parsec or other parser combinator libraries. It always succeeds, consumes nothing and returns a  given value. I know that it can be achieved with `value!` macro, but I thought since version 5 using macros are discouraged.

My use case for this macro is included as the doc test, it is a parser for optional sign which I wanted to express like this:

```
alt((value(1, char('+')), value(-1, char('-')), success(1))) 
```

Using `success` here allows to specify a default for `alt` which might be quite useful. It is of course possible to achieve it using just a closure `|i| Ok((i, 1))`, but I think having a special combinator is more idiomatic.


sphynx:
I've rebased this change against recent 6.0 alpha.

Geal:
added in dbaa2e5, thanks!

PR #1127, opened 2020-03-27
Add with_input_consumed function very similar to recognize

Alfriadox:
Also ran rustfmt.

Still to be added: 
- [x] implementation
- [ ] edit `choosing_a_combinator.md`
- [ ] more testing 
- [x] some documentation

Most of the 53 files changed here are from running rustfmt. The only significant changes are in combinator/mod.rs.

PR #1146, opened 2020-04-25
consumed (previously with_input) parser combinator

Alfriadox:
This replaces #1127 

This pull request adds a new parser combinator, similar to recognize. This combinator, called "consumed," returns the consumed input along with the parser result. 

TODO:
- [x] Implement combinator
- [x] Write documentation
- [x] Modify [combinator list](https://github.com/Geal/nom/blob/master/doc/choosing_a_combinator.md)
- [ ] Macro version.

Comments and feedback appreciated!

### DIFF
--- a/doc/choosing_a_combinator.md
+++ b/doc/choosing_a_combinator.md
@@ -85,6 +85,7 @@ Parsing integers from binary formats can be done in two ways: With parser functi
 - `parse_to!`: Uses the parse method from `std::str::FromStr` to convert the current input to the specified type
 - `peek!`: Returns a result without consuming the input
 - `recognize!`: If the child parser was successful, return the consumed input as the produced value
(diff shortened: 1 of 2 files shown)


Alfriadox:
@Geal I cannot create a proper macro version of this combinator similar to the one for `recognize!` because the output type conflicts with the one specified in `named!`. Otherwise this PR should be ready for merge. What are your thoughts?

Stargateur:
I'm not a native English speaker. `consumed` is not very self describing in my opinion.

Alfriadox:
@Stargateur That's understandable! What would you suggest as a name -- I don't want to make it too verbose.

Alfriadox:
When this PR is merged I think it should close #1183. I don't have the permissions to link the issue right now though I think.

Geal:
thank you for this new combinator!  Don't worry about the macro version, they're not used as much these days

PR #1150, opened 2020-05-04
Add ANT FIT parser to README

stadelmanma:
Add a link in README to my [ANT FIT](https://www.thisisant.com/developer/ant/ant-fs-and-fit1/) file parser implemented using nom.

Thank you for providing this awesome parsing library!


Geal:
thanks!

PR #1151, opened 2020-05-05
Add custom errors example

AntoineCezar:
Hello,

First, thanks for this awesome crate that make parsing more accessible to non-specialists in my opinion.

I'm fairly new to rust and had a hard time figuring out how to have my own errors. It's the only things I had to battle with during man short nom usage. So now (I think) I've found how, I would like to add an example to avoid the same hassles for future users.

I wish you the best for your personal problems.

### DIFF
--- a/doc/error_management.md
+++ b/doc/error_management.md
@@ -87,6 +87,9 @@ expected '}', found 1
   ^
 ```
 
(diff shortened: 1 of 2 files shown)


Geal:
Thank you! It's good to have an example of custom errors

PR #1152, opened 2020-05-07
Replace "cfg(test)" with "cfg(doctest)" for readme testing

GuillaumeGomez:
Rustdoc now passes "doctest" when running in test mode.


Geal:
awesome, thanks!

PR #1158, opened 2020-05-29
Fix a few typos in the documentation

Eijebong:


kali:
I've heard it both ways

Geal:
pushed as 74bdbed (conflicts with another PR)

PR #1161, opened 2020-06-11
Add a functional version of length_count.

resistor:
This is an existing macro combinator that was missing from the functional side.  Testcases are included in the doc comment.


Geal:
Thanks!

Issue #1063, opened 2019-11-08
Nested separated_list for empty input fails?


zbraniecki:
- Rust version : 1.39
- nom version : 5.0.1

## Test case

```rust
fn and_parser(i: &str) -> IResult<&str, Vec<&str>> {
    separated_list(tag(" and "), tag("foo"))(i)
}

fn or_parser(i: &str) -> IResult<&str, Vec<Vec<&str>>> {
    separated_list(tag(" or "), and_parser)(i)
}

assert_eq!(parser("foo and foo or foo"), Ok(("", vec![vec!["foo", "foo"], vec!["foo"]])));

// FAILS with `Err(Error(("", SeparatedList)))`
assert_eq!(or_parser(""), Ok(("", vec![vec![]])));
```

Why wouldn't that work?

zbraniecki:
According to @pcpthm analysis in https://users.rust-lang.org/t/nom-5-and-nested-separated-list/34590/2?u=zbraniecki it seems like a bug in nom

Geal:
now that ##1170 is merged, this should be more consistent

PR #1170, opened 2020-07-05
Allow empty elements in separated_list

ericseppanen:
There's no particular reason why empty elements shouldn't be allowed in the separated_list[0|1] macros and functions. It's useful to be able to parse things like CSV `",,,foo"`.

An empty separator is still not allowed.

This change includes test coverage of both cases (where the previous tests were, in `macros.rs`; not sure if more is expected in `issues.rs`.

This is intended to address #1063. (There seem to be older reports as well, but that's the one that's still open.)

An example of code that wouldn't work before, and does after this change:
```rust
use nom::IResult;
use nom::multi::separated_list0;
use nom::character::complete::alphanumeric0;

fn comma_separated_text(input: &str) -> IResult<&str, Vec<&str>> {
    separated_list0(tag(","), alphanumeric0)(input)
}

#[test]
fn test_comma_separated() {
    assert_eq!(comma_separated_text("a,b,c"), Ok(("", vec!["a", "b", "c"])));
    assert_eq!(comma_separated_text(",,c"), Ok(("", vec!["", "", "c"])));
}

```

### DIFF
--- a/src/multi/macros.rs
+++ b/src/multi/macros.rs
@@ -583,6 +583,7 @@ mod tests {
   fn separated_list0() {
     named!(multi<&[u8],Vec<&[u8]> >, separated_list0!(tag!(","), tag!("abcd")));
     named!(multi_empty<&[u8],Vec<&[u8]> >, separated_list0!(tag!(","), tag!("")));
+    named!(empty_sep<&[u8],Vec<&[u8]> >, separated_list0!(tag!(""), tag!("abc")));
     named!(multi_longsep<&[u8],Vec<&[u8]> >, separated_list0!(tag!(".."), tag!("abcd")));
 
     let a = &b"abcdef"[..];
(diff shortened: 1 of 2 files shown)


ericseppanen:
There is one corner case here where I'm not certain the new behavior is correct.

Same example as above:
```rust
fn comma_separated_text1(input: &str) -> IResult<&str, Vec<&str>> {
    separated_list0(tag(","), alphanumeric0)(input)
}
```

What should the result be if the input doesn't contain any alphanumeric characters or commas?

```rust
    assert_eq!(comma_separated_text(""), Ok(("", vec![""])));
    assert_eq!(comma_separated_text("$"), Ok(("$", vec![""])));
```

The result is always `Ok((__, [""]))`.  This kind of makes sense... I mean, I did tell it I wanted to allow empty elements.  But it does mean that `separated_list0` and `separated_list1` behave exactly the same when the element can be empty, since both will always return a single empty element in the face of unmatched input.

Geal:
it would make sense that a parser allowing empty element would return something on empty inputs, that's fine with me

Issue #1169, opened 2020-06-29
Enhancement to Docs: A "recipes" section


rljacobson:
Hi! Long time listener, first time caller. Love nom. 

Would you be interested in a PR for a "recipes" section of the docs that contains relatively short snippets to do common tasks? I'm thinking of stuff that's less than a complete example but more than a single combinator:

- EOL comments
- C-style comments, unnested and nested
- floats (an arbitrary but named syntax that a user could modify to their language)
- combinators that have been asked for but for which there is no built-in
- Python-style indentation
- identifiers
- common integer forms (`0x3f5a`, `34u`, `0b11010011`, etc.)

Also, are there more detailed docs or examples for verbose errors with row/col info? If not, and if you are agreeable to a PR along those lines, maybe you could point me to a nom parser that uses it, and I can extract bits to make a minimal but thorough tutorial.

nilsmartel:
Long time lurker here as well, id love and cherish this idea!

phorward:
sounds great!

nilsmartel:
I'd like to add an interesting example:

- Javascript style string interpolation (note: js is just an example here)

E.g.
```js
     `Hello ${user.name}. How are you today?`
```

nilsmartel:
While this is usually something one would put in the wiki, I'd like to entertain the idea to create a separate `Recipes.md` in the root directory of the project.

People like me tend to clone and explore repositories and these recipes would be nice to have alongside the code.

Also nice: This way everyone can have a simple cheatsheet that is easiest to adapt once the project gets cloned

rljacobson:
@nilsmartel Maybe we collectively author it as a wiki page first, as there may be many authors interested in contributing recipes.

nilsmartel:
I just took a look at the wiki to start right in, but they all are moved to the source code.
Non the less, I think it might be a good catalyst for this

rljacobson:
I set up a publicly editable wiki page on my fork here: https://github.com/rljacobson/nom/wiki/Nom-Recipes. When we're ready, I'll fold it into the docs and make a PR. Future additions would then just make a PR on the existing recipes page in the docs.

nilsmartel:
@rljacobson you're doing beautiful work here.
I'd like to have little discussion about it here:

[...]

PR #1180, opened 2020-08-01
Add a "recipes" section to the docs (issue #1169).

rljacobson:
This PR implements issue #1169. While I am listed as the author, @nilsmartel also made significant contributions.

- [ ] Review to see if the new `nom_recipes.md` should be linked to from other places in the docs. (I.e. should the README link to it?)
- [ ] I have all examples in a compilable rs file. It's not clear if the code should be in a test. (I argue yes.) If so, it's not clear to me how best to do this.


doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+### Wrapper combinators that eat whitespace before and after a parser
+
+```rust
+/// A combinator that takes a parser `inner` and produces a parser that also consumes both leading and 
+/// trailing whitespace, returning the output of `inner`.
+fn ws<'a, F: 'a, O, E: ParseError<&'a str>>(inner: F) -> impl Fn(&'a str) -> IResult<&'a str, O, E>
+  where
+  F: Fn(&'a str) -> IResult<&'a str, O, E>,
```
calebsander, line 32: This can be `F: Parser<&'a str, O, E>` and the return type can be `impl Parser<&'a str, O, E>`

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+{
+  move |i| {
+    delimited(
+      multispace0,
+      &inner,
+      multispace0
+    )(i)
+  }
```
calebsander, line 40: Why not just return `delimited(...)` directly?

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+### `// C++/EOL-style comments`
+
+This version does not consume the newline character and returns an output of `()`.
+
+```rust
+pub fn peol_comment<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, (), E>
+{
+  map(pair(char('%'), is_not("\n\r")),
+      |_| () // Output is thrown away.
+  )(i)
```
calebsander, line 57: Could use `value()` instead of `map()` (below as well)

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+## Comments
+
+### `// C++/EOL-style comments`
+
+This version does not consume the newline character and returns an output of `()`.
+
+```rust
+pub fn peol_comment<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, (), E>
+{
+  map(pair(char('%'), is_not("\n\r")),
```
calebsander, line 55: Mention that this uses `%` as the comment start sequence?

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+    pair(
+      alt((alpha1, tag("_"))),
+      many0(alt((alphanumeric1, tag("_"))))
+    )
+  )(input)
+}
+```
+
+Let's say we apply this to the identifier `hello_world123abc`.
+The first `alt` parser would recognize `hello_world`. The `pair` combinator ensures that `123abc` will be piped to the next `alphanumeric0` parser, which recognizes every remaining character.
```
calebsander, line 96: The first `alt` parser actually recognizes only `hello` since it delegates to `alpha1`

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+
+This is [one of the examples](https://github.com/Geal/nom/blob/master/examples/string.rs) in the examples directory.
+
+### Integers
+
+The following recipes all return string slices rather than integer values. How to obtain an integer value instead is demonstrated for hexadecimal integers. The others are similar.
+
+The parsers allow the grouping character `_`, which allows one to group the digits by byte, for example: `0xA4_3F_11_28`. If you prefer to exclude the `_` character, the lambda to convert from a string slice to an integer value is slightly simpler. You can also strip the `_` from the string slice that is returned, which is demonstrated in the second hexdecimal number parser.
+
+If you wish to limit the number of digits in a valid integer literal, replace `many1` with `many_m_n` in the recipes: `many1(one_of("_01"))` becomes `many_m_n(1, 16, one_of("_01"))` for binary numbers having at most 16 bits.
```
calebsander, line 111: This will parse numbers with at most 16 *characters*, including `_`, not 16 digits. You could replace `one_of("_01")` with `terminated(one_of("01"), many0(char('_')))`.

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+fn hexadecimal_value<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str, i64, E> {
+  map(preceded(
+    alt((tag("0x"), tag("0X"))),
+      recognize(
+        many1(
+          one_of("_0123456789abcdefABCDEF")
+        )
+      )
+    ),
+    |out: &str| i64::from_str_radix(&str::replace(&out, "_", ""), 16).unwrap()
```
calebsander, line 142: `unwrap()` can fail if the literal has too many digits. It would be better to use `map_res()` to handle this case rather than panicking.

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+The parsers allow the grouping character `_`, which allows one to group the digits by byte, for example: `0xA4_3F_11_28`. If you prefer to exclude the `_` character, the lambda to convert from a string slice to an integer value is slightly simpler. You can also strip the `_` from the string slice that is returned, which is demonstrated in the second hexdecimal number parser.
+
+If you wish to limit the number of digits in a valid integer literal, replace `many1` with `many_m_n` in the recipes: `many1(one_of("_01"))` becomes `many_m_n(1, 16, one_of("_01"))` for binary numbers having at most 16 bits.
+
+#### Hexadecimal
+
+The parser outputs the string slice of the digits without the leading `0x`/`0X`.
+
+```rust
+fn hexadecimal<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str,&'a str, E> {
```
calebsander, line 118: Nit, but a bunch of these function signatures are missing a space after `IResult<&'a str,`

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Decimal
+
+```rust
+fn decimal<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str,&'a str, E> {
+  recognize(
+    tuple((
+      one_of("0123456789"),
+      many0(one_of("_0123456789")
+      )
```
calebsander, line 185: Nit: this parenthesis should be on the previous line

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+    )
+    , // Case two: 42e42 and 42.42e42
+    recognize(
+      tuple((
+        decimal,
+        opt(preceded(
+          char('.'),
+          decimal,
+        )),
+        tuple((
```
calebsander, line 218: The nested `tuple()` seems unnecessary

doc/nom_recipes.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+### Floating Point Numbers
+
+The following is adapted from [the Python parser by Valentin Lorentz (ProgVal)](https://github.com/ProgVal/rust-python-parser/blob/master/src/numbers.rs).
+
+```rust
+fn float<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str,&'a str, E> {
+  alt((
+    // Case one: 42.
```
calebsander, line 198: This actually recognizes `.42`

doc/nom_recipes.md
rljacobson: The `Parser` trait is not publicly exported in v5.1.2.

doc/nom_recipes.md
rljacobson: Vestigial code, I think. :)

doc/nom_recipes.md
rljacobson: No, the first `alt` only recognizes the `'h'`. But yeah, looks like I forgot to update the text with the code.

doc/nom_recipes.md
rljacobson: Nice catch! And elegant solution, too.

doc/nom_recipes.md
rljacobson: I debated about how much error handling I should do in these. I think I agree with you, `map_res()` is the right thing to do here.

...but now rustc can't infer the error type.

```
189 |   let x = hexadecimal_value("0xdeadbeef");
    |       -   ^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `E` declared on the function `hexadecimal_value`
    |       |
    |       consider giving `x` the explicit type `std::result::Result<(&str, i64), nom::Err<E>>`, where the type parameter `E` is specified
```

What is the concrete error type in this case? 

Debugging "cannot infer type" errors should be a recipe, actually.

doc/nom_recipes.md
rljacobson: Ach! Makes me twitch.

doc/nom_recipes.md
rljacobson: Ah, I remember now. I was emulating how Nom defines its combinators. To take a random example, look at [the implementation of char](https://github.com/Geal/nom/blob/master/src/character/complete.rs#L31).

As for why Nom uses a closure... I don't know.

rljacobson:
Awesome peer review by @calebsander. Thanks!

doc/nom_recipes.md
nilsmartel: I think we should through in a comment here, that the unwrap might fail.
reason:
- The error handling distracts from the goal to communicate just *parser* design
- Rust users are trained to have an alert going on in their heads once they read `unwrap` anyway. They will use this with caution

nilsmartel:
I agree and learned a lot because of you @calebsander.

doc/nom_recipes.md
calebsander: That's a good point; I didn't know that. You can remove the `'a` restriction on `F` if you don't pass `inner` by reference into `delimited()`. `F` can also be `FnMut(...) -> ...`.

doc/nom_recipes.md
calebsander: The reason Nom uses closures is that this is the only way to define primitive parsers and combinators. The Nom parser and combinator functions return "parser functions", which need custom logic, and that logic is encoded in the closure. Some of Nom's combinators could be implemented in terms of other combinators, but I believe this is avoided to improve performance. For example, the [`value()` combinator](https://github.com/Geal/nom/blob/1c34db743864455568ca9d30b2a90b26d94cca93/src/combinator/mod.rs#L535) could be written in terms of `map()`:
```rust
pub fn value<I, O1: Clone, O2, E: ParseError<I>, F>(
  val: O1,
  parser: F,
) -> impl FnMut(I) -> IResult<I, O1, E>
where
  F: Parser<I, O2, E>,
{
  map(parser, move |_| val.clone())
}
```

So it basically comes down to whether are implementing a new primitive combinator (which is rare, unless you are contributing to Nom), or using Nom's existing combinators to define a new combinator (which is usually the case, since a rich set of combinators are already provided).

calebsander: Here is how I would write this combinator (and I think it's a great example of a custom combinator):
```rust
fn ws<'a, O, E, F>(inner: F) -> impl FnMut(&'a str) -> IResult<&'a str, O, E>
where
  E: ParseError<&'a str>,
  F: FnMut(&'a str) -> IResult<&'a str, O, E>,
{
  delimited(multispace0, inner, multispace0)
}
```

doc/nom_recipes.md
calebsander: Here is a test that shows the first `alt` matches `hello`. Try it for yourself. This is because `alpha1` greedily matches 1 or more alphabetic characters, so it will match `h`, `e`, `l`, `l`, and `o` before failing to parse `_`.
```rust
use nom::branch::*;
use nom::bytes::complete::*;
use nom::character::complete::*;
use nom::multi::*;
use nom::sequence::*;
use nom::*;

pub fn identifier(input: &str) -> IResult<&str, (&str, Vec<&str>)> {
  pair(
    alt((alpha1, tag("_"))),
    many0(alt((alphanumeric1, tag("_"))))
  )(input)
}

assert_eq!(identifier("hello_world123abc"), Ok(("", ("hello", vec!["_", "world123abc"]))));
```

doc/nom_recipes.md
calebsander: > ...but now rustc can't infer the error type.
> 
> ```
> 189 |   let x = hexadecimal_value("0xdeadbeef");
>     |       -   ^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `E` declared on the function `hexadecimal_value`
>     |       |
>     |       consider giving `x` the explicit type `std::result::Result<(&str, i64), nom::Err<E>>`, where the type parameter `E` is specified
> ```

This doesn't have anything to do with `map_res()`; you would see the same behavior with the original version. It's because the parser is generic over error types `E`, so Nom can't tell what error type you want to use. (Among other options, you can choose between `()` (which stores no information about how the parse failed), `(I, ErrorKind)` (which stores the root cause of the parse failure), and `VerboseError<I>`(which tracks the full context of an error, e.g. `char` failed, inside an `alt`, inside a `many0`).
I agree this is a somewhat awkward feature of Nom and probably deserves a recipe. There are two main solutions: specify `E` explicitly when you call the parser (which is more general), or choose a particular type for `E`.
Here is what specifying a type for `E` looks like:
```rust
let x = hexadecimal_value::<VerboseError<_>>("0xdead_beef");
let y = hexadecimal_value::<(_, ErrorKind)>("0xdead_beef");
```
And fixing a particular `E` just requires changing the function signature. You can use the fact that `IResult` defaults to `E = (I, ErrorKind)`:
```rust
fn hexadecimal_value(input: &str) -> IResult<&str, i64>
```

calebsander: > I think we should through in a comment here, that the unwrap might fail.
> reason:
> 
> * The error handling distracts from the goal to communicate just _parser_ design
> * Rust users are trained to have an alert going on in their heads once they read `unwrap` anyway. They will use this with caution

I agree it feels un-Rusty to not handle the error here. If the parser says it parses an `i64`, then it should either successfully parse an `i64`, or it should fail to parse, but it shouldn't panic. I do think it would be valuable to use `map_res()` here. It actually results in less code, and it is a very idiomatic way to build error handling into the parsing. And it doesn't seem like any of the other examples illustrate `map_res()`. Here's the `map_res()` version, with some indentation fixes:
```rust
map_res(
  preceded(
    alt((tag("0x"), tag("0X"))),
    recognize(many1(one_of("_0123456789abcdefABCDEF"))),
  ),
  |out: &str| i64::from_str_radix(&out.replace("_", ""), 16)
)(input)
```

doc/nom_recipes.md
rljacobson: > You can remove the 'a restriction on F if you don't pass inner by reference into delimited(). 

There's no need to pass `inner` by reference, but if I elide the explicit lifetime `'a` I get a cryptic type error. [Here's a gist of my test dirty code](https://gist.github.com/rljacobson/840e99239eadc27b91d99feabf40a50a#file-nom-recipes-testing-rs-L25) if you want to take a look. (I wish the rust playground had the nom crate.)

> `F` can also be `FnMut(...) -> ...`.

Another version difference: `nom::sequence::delimited` requires an `Fn` in [v5.1.2](https://docs.rs/nom/5.1.2/nom/sequence/fn.delimited.html).

But since this is the second time it's come up, maybe the recipes *should* target v6.0.0. What do you think?

doc/nom_recipes.md
rljacobson: Thanks for the explanation!

Stylistically, having `E` in the `where` clause improves readability. I'll change that throughout. (See above for `FnMut` and `'a`.)

doc/nom_recipes.md
rljacobson: Oh, no test necessary, as it's completely obvious! \**facepalm*\* I had edited my version of the function. *Doh!*

I appreciate the excellent example, though!

doc/nom_recipes.md
rljacobson: @nilsmartel I would normally agree with you. The reason @calebsander changed my mind in this case is because this is exactly the use case for nom's built-in [`map_res()`](https://docs.rs/nom/5.1.2/nom/combinator/fn.map_res.html). I agree that the reader will be alerted to the need for additional error handling by the presence of `unwrap()`, but if we leave it there, we will be sending the reader down a blind alley, so to speak. They will ultimately want to know about `map_res()` anyway. Including it here tips them off to that need without necessarily cluttering the exposition.

doc/nom_recipes.md
calebsander: Hmm, I just started using Nom, so I didn't know there were so many breaking changes. I don't think it matters so much which version the examples are for, but it would be good to point out any changes that need to be made to work with the other version.
As far as the lifetime is concerned, I don't see any type errors on your gist version. The lifetime is only required if you pass `inner` by reference, since `inner` will go out of scope at the end of the function (before any parsing is done).

doc/nom_recipes.md
rljacobson: I don't think it matters for the purpose of the recipe, but Rust allows prefixed numbers (`0x`-, `0o`-, `0b`-) to start with an underscore: `0x_dead`. Having *only* an `_`, as in `0x_`, is an error, though.

calebsander: Yeah I noticed that, but wasn't sure it was worth complicating the example. You could always slap a `preceded(opt(char('_'), ...)` on there to fix that.

PR #1196, opened 2020-09-13
Fix undefined behavior in JSON test/bench (#1195)

### LINKED ISSUE #1195: Undefined behavior in JSON test and benchmark
The current parser UBs on [this line](https://github.com/Geal/nom/blob/13e047e085e96e82ad147e4efab7d1d453dcd8c8/tests/json.rs#L51) if the input string contains a UTF-16 surrogate (see [JSON](https://en.wikipedia.org/wiki/JSON#Data_portability_issues) and [UTF-16](https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF) for details).

## Prerequisites

- Rust version : `rustc 1.46.0 (04488afe3 2020-08-24)`
- nom version : master
- nom compilation features used: default

I have a PR ready to both fix this and implement actual parsing of surrogates.

FallenWarrior2k:
This fixes the undefined behavior caused by unsafely turning a u32 into a char without checking the necessary preconditions.
It also implements decoding surrogates if a Unicode escape sequence decodes to a high surrogate and a corresponding low surrogate is found in a directly adjacent second Unicode escape.

Fixes #1195.

### DIFF
--- a/benches/json.rs
+++ b/benches/json.rs
@@ -11,7 +11,7 @@ use nom::{
   branch::alt,
   bytes::complete::{tag, take},
   character::complete::{anychar, char, multispace0, none_of},
-  combinator::{map, map_res, value},
+  combinator::{map, map_opt, map_res, value},
   error::{ErrorKind, ParseError},
   multi::{fold_many0, separated_list0},
   number::complete::{double, recognize_float},
@@ -50,11 +50,11 @@ fn character(input: &str) -> IResult<&str, char> {
           _ => return Err(()),
         })
       }),
-      map(
+      map_opt(
         map_res(preceded(char('u'), take(4usize)), |s| {
           u16::from_str_radix(s, 16)
         }),
-        |c| unsafe { std::char::from_u32_unchecked(c as u32) },
+        |c| std::char::from_u32(c as u32),
       ),
     ))(input)
   } else {
--- a/benches/json.rs
+++ b/benches/json.rs
@@ -11,7 +11,7 @@ use nom::{
   branch::alt,
   bytes::complete::{tag, take},
   character::complete::{anychar, char, multispace0, none_of},
-  combinator::{map, map_opt, map_res, value},
+  combinator::{map, map_opt, map_res, value, verify},
   error::{ErrorKind, ParseError},
   multi::{fold_many0, separated_list0},
   number::complete::{double, recognize_float},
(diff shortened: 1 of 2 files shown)


benches/json.rs
```diff
@@ -118,7 +142,7 @@ fn json(input: &str) -> IResult<&str, JsonValue> {
 
 fn json_bench(c: &mut Criterion) {
   let data = "  { \"a\"\t: 42,
-  \"b\": [ \"x\", \"y\", 12 ] ,
+  \"b\": [ \"x\", \"y\", 12 ,\"\\u2014\", \"\\uD83D\\uDE10\"] ,
```
calebsander, line 145: Not sure I would include this in the benchmark, as I don't think it's a very common case.

benches/json.rs
```diff
@@ -35,6 +35,35 @@ fn boolean(input: &str) -> IResult<&str, bool> {
   alt((value(false, tag("false")), value(true, tag("true"))))(input)
 }
 
+fn u16_hex(input: &str) -> IResult<&str, u16> {
+  map_res(take(4usize), |s| u16::from_str_radix(s, 16))(input)
+}
+
+fn unicode_escape(input: &str) -> IResult<&str, char> {
```
calebsander, line 42: The standard library already implements decoding of UTF-16 sequences to characters. Is it possible to use [`char::decode_utf16()`](https://doc.rust-lang.org/std/char/fn.decode_utf16.html) instead of hand-rolling this code? I guess it would be a little awkward to construct an iterator over `u16`s decoded from the JSON. A simpler method might be something like:
```rust
// Assume unicode_escape() parses e.g. "\u12AB"

fn unicode_char(input: &str) -> IResult<&str, char> {
  let (input, cp) = unicode_escape(input)?;
  if let Some(c) = char::from_u32(cp) {
    return Ok((input, c));
  }

  map_res(unicode_escape, |cp2| {
    char::decode_utf16([cp, cp2].iter().copied()).next().unwrap()
  })(input)
}
```

calebsander (commented):
I wrote this code, and this is a good point that I glossed over. Really the JSON specification doesn't say anything about handling UTF-16 surrogate pairs; these characters can be written directly without escape characters. So I think it would be sufficient (and faster) to just use `map_opt(u16_hex, char::from_u32)`, treating surrogate pairs as invalid JSON.

benches/json.rs
FallenWarrior2k: Fair. I was just thinking to cover all the paths.

FallenWarrior2k:
>So I think it would be sufficient (and faster) to just use map_opt(u16_hex, char::from_u32), treating surrogate pairs as invalid JSON.

A valid choice and Python's native `json` module does that as well. `serde_json` however, does decode them transparently (see [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=691032b762cc259599c0211923dd476d)).
My main reason for implementing it is that I already had the code from a previous experiment project, but in an entirely imperative form, and thought this a nice idea to experiment with nom's functional style a bit, since the rest of that file was almost fully written in that style as well.

I'll leave it up to you. I'll also leave the other discussion around for time being, depending on that.

calebsander:
Sure, if `serde_json` handles surrogate pairs, then the test/bench probably should as well. Here's the version I suggested above, which handles surrogate pairs without reimplementing `std::char::decode_utf16()` from the standard library: https://github.com/calebsander/nom/compare/master...fix/json-surrogate-pairs.

Geal:
Thanks! It looks like despite my best attempts, that parser will end up spec compliant :smile:

PR #1199, opened 2020-09-20
Add vimwiki parser and associated macro library to list of parsers

chipsenkbeil:
Both the library and macro library are functional and part of an effort to standardize the vimwiki language (see https://github.com/vimwiki/vimwiki/issues/989) as well as provide the foundation for tooling on top.


Geal:
Thanks!

PR #1200, opened 2020-09-26
clarify description of take_until

fanf2:
Greetings!

This is a small doc change because I wasn't sure I understood the description of take_until. I have included doctests to illustrate the case I wasn't sure about - does `nom` eat up to the first tag or the last one?

I have submitted this patch against the 5.0 branch because the current master doesn't build for me with latest stable rust. I'm happy to resubmit against master if you would prefer!

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -414,7 +414,7 @@ where
   }
 }
 
-/// Returns the longest input slice till it matches the pattern.
+/// Returns the input slice up to the first occurrence of the pattern.
 ///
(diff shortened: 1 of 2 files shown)


Geal:
hello! this should go on the master branch and it should compile on stable now, I had not pushed some dependencies change around bitvec

fanf2:
Yes, master builds for me now, so I have rebased this PR and `cargo t` passes OK :-)

Geal:
Thanks!

PR #1202, opened 2020-10-07
Make map, map_res and map_opt accept FnMut instead of Fn

ruza-net:
Hello!

I've run into an issue while writing a parser, which would be resolved by making `map` (and friends) accept `FnMut` instead of just `Fn` as the 2nd argument.

So I changed that. The `cargo test` still runs without failures, and there isn't any obvious unintended behaviour.

Since `map` and friends are used inside combinators, the captured context will at most modify the locals and arguments of the surrounding parser-function.

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -70,10 +70,10 @@ where
 /// assert_eq!(parser.parse("abc"), Err(Err::Error(("abc", ErrorKind::Digit))));
 /// # }
 /// ```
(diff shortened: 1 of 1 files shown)


Geal:
hello, what was the issue you encountered?

ruza-net:
I got some parsed data from a call, which I needed to capture and move in the closure. That requires `FnMut` instead of `Fn`. Otherwise, I would have to `clone` the data.

Obviously, I can map the closure manually, but I don't see why `map` shouldn't support this.

Geal:
Thanks!

ruza-net:
Oh, I thank *you* :)

PR #1203, opened 2020-10-10
Correct Incomplete errors from length_data/value

jameysharp:
I believe the doc-tests were wrong for these macros/functions. Any example from this patch should illustrate the issue, such as:

```rust
named!(x, length_data!(le_u8));
assert_eq!(x(b"\x02."), Err(Err::Incomplete(Needed::new(2))));
```

That parser should see the `\x02` and try to read two bytes, but there is only one available after that. That means it needs one more, but it currently reports that it needs all two.

I also wrote a couple other miscellaneous patches for things I spotted while fixing this: one to remove some unnecessary `Copy` constraints, and another to add some doc-tests to the `tag` parser for behavior that wasn't obvious to me from the documentation.


Geal:
thanks! this will follow the fixes in #1055

Issue #1205, opened 2020-10-18
Why is there no nom::character::is_newline?


jmg-duarte:
As the title states I don't understand why there is no `nom::character::is_newline`.

To provide more context, I'm currently developing a Markdown parser as a learning experiment.
To parse code blocks I currently do the following:

```rust
pub fn parse_code_block(input: &str) -> IResult<&str, Element> {
    sequence::delimited(complete::tag("```"), parse_code, complete::tag("```"))(input)
}

fn parse_code(input: &str) -> IResult<&str, Element> {
    let (input, (language, code)) = sequence::pair(
        nom::combinator::opt(
            complete::take_till1(|c| c == '\n') // <-- Notice here
        ),
        complete::take_until("`"),
    )(input)?;
    Ok((input, Element::Code(Code::new(code, language))))
}
```

In the marked line of the code, I am required to do `|c| c == '\n'` instead of being able to use a simple provided function.

Is there some underlying reason I am missing, or should I implement the function and submit a PR?

Geal:
there's no underlying reason, it could be added

jmg-duarte:
I'll submit a PR asap then!

PR #1209, opened 2020-10-25
Add nom::character::is_newline function

jmg-duarte:
The function is present in nom::character and tests if a byte is equal
to "\n".

This PR is related to https://github.com/Geal/nom/issues/1205

I also noticed that there are several char traits that use similar functions.
If this function is required to be included in such traits I'll be happy to add such changes.

If anything else is required I'm open to improve this PR


Geal:
Thanks

PR #1210, opened 2020-10-26
Add another example to character::complete::char

hbina:
I had to check the implementation to know this was the case.


Geal:
thanks

PR #1211, opened 2020-10-26
Update PR template to be a comment.

hbina:
Removes the need to delete them.


Geal:
ah nice, I didn't know it was possible

PR #1212, opened 2020-10-28
Enable Display impl of Err on no-std configuration

HeroicKatora:
The impl was previously hidden behind `cfg(feature = "std")`. This is uneccesary as all relevant traits and even the macro to write a formatted output appear in `core`.


Geal:
nice catch, thanks

PR #1213, opened 2020-10-29
Add GitHub Actions workflow for CI

nickelc:
The PR add a workflow based on the Travis config, except that `fmt/doc/MSRV/bench/coverage` are separate build job definitions and not part of the main build matrix.

The different build jobs are now more visible with the better UI integration and run faster with the parallel execution.

I kept the coverage & bench build jobs simple because i have no experience with coveralls.io
and I'm not sure what the current bench part is trying to achieve and if it is working as intended.

[Result](https://github.com/nickelc/nom/actions/runs/339088734)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,185 @@
+name: CI
+
+on: [push, pull_request]
+
+env:
+  RUST_MINVERSION: 1.37.0
+
+jobs:
+  test:
+    name: Test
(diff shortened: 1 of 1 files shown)


Geal:
that's great! Thanks! Travis was getting too slow for nom builds

nickelc:
> that's great! Thanks! Travis was getting too slow for nom builds

yeah, i noticed that the first bulk of build jobs run parallel but the rest is executed sequentially

PR #1214, opened 2020-10-29
Added a helper function to check for vertical tab.

hbina:
Encoding hurts my brain, I hope this is correct. Apparently Rust does not support using `\v` directly.

Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>


Geal:
hello, it's missing the streaming version, I think

PR #1217, opened 2020-10-30
Closes #1215

### LINKED ISSUE #1215: BitVec take_split impl returns the splits in the wrong order
- Rust version : 1.49.0-nightly
- nom version : HEAD
- nom compilation features used: default

## Test case

This behavior can actually be observed in an existing test case at https://github.com/Geal/nom/blob/master/tests/bitstream.rs

```
#[test]
fn parse_bitstream() {
  let data = [0xA5u8, 0x69, 0xF0, 0xC3];
  let bits = data.view_bits::<Msb0>();

  fn parser(bits: &BitSlice<Msb0, u8>) -> IResult<&BitSlice<Msb0, u8>, &BitSlice<Msb0, u8>> {
    tag(bits![1, 0, 1, 0])(bits)
  }

  assert_eq!(parser(bits), Ok((&bits[..4], &bits[4..])));
}
```

In the assertion, it's asserting the value matched by  the `tag` parser in the first position of the tuple, and the remaining bits in the second position of the tuple. The following (currently failing) test case makes this issue more apparent:

```
#[test]
fn parse_bitstream_with_map() {
  let data = [0b1000_0000];
  let bits = data.view_bits::<Msb0>();

  fn parser(bits: &BitSlice<Msb0, u8>) -> IResult<&BitSlice<Msb0, u8>, bool> {
    map(take(1_u8), |val: &BitSlice<Msb0, u8>| {
      // You can observe here that the result of the take parser is supplying the remaining bits to
      // the mapping function, not the bits that were taken.
      println!("Expected val to be [1] but is {}", val);
      val[0]
    })(bits)
  }

  assert_eq!(parser(bits), Ok((&bits[1..], true)));
}
```

cc @myrrlyn

myrrlyn:


Geal:
closing this as it was fixed in #1216

PR #1218, opened 2020-10-31
Re-export bitvec

bgourlie:
Re-export bitvec crate so that types necessary to create bit parsers are exposed.

I had considered a couple different ways of doing this:

- Re-export bitvec from `nom::bits`
- Create a `nom::bits::prelude` module that re-exports `bitvec::prelude::*`

Ultimately, the way it's being done in this PR seems straightforward and is similar to what's currently being done with the regex crate.


Geal:
Thanks!

PR #1233, opened 2020-11-06
Remove #[deprecated] attribute from trait impl block

nickelc:
The PR fixes the [nightly builds](https://github.com/nickelc/nom/runs/1363679867?check_suite_focus=true#step:4:40)

Deprecation attributes have no effect on trait implementations. See rust-lang/rust#78626


Geal:
huh, that one was annoying, thanks for fixing it!

PR #1234, opened 2020-11-07
docs: Remove wrong call from a whitespace combinator example

jlkiri:
It tries to apply the returned function to source immediately when it should just return the combinator.

### DIFF
--- a/doc/nom_recipes.md
+++ b/doc/nom_recipes.md
@@ -27,7 +27,7 @@ These are short recipes for accomplishing common tasks with nom.
 ```rust
 /// A combinator that takes a parser `inner` and produces a parser that also consumes both leading and 
 /// trailing whitespace, returning the output of `inner`.
-fn ws<'a, F: 'a, O, E: ParseError<&'a str>>(inner: F) -> impl Fn(&'a str) -> IResult<&'a str, O, E>
+fn ws<'a, F: 'a, O, E: ParseError<&'a str>>(inner: F) -> impl FnMut(&'a str) -> IResult<&'a str, O, E>
   where
   F: Fn(&'a str) -> IResult<&'a str, O, E>,
 {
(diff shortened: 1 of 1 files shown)


SeanMcLoughlin:
I was trying to make sense of this documentation example, but even with this change, the documentation is still incorrect and I hit compilation errors:

```rust
error[E0277]: expected a `std::ops::Fn<(&'a str,)>` closure, found `std::result::Result<(&str, O), nom::Err<E>>`
  --> src/functions.rs:18:6
   |
18 | ) -> impl Fn(&'a str) -> IResult<&'a str, O, E>
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an `Fn<(&'a str,)>` closure, found `std::result::Result<(&str, O), nom::Err<E>>`
...
22 |     delimited(multispace0, inner, multispace0)(i)
   |     --------------------------------------------- this returned value is of type `std::result::Result<(&'a str, O), nom::Err<E>>`
   |
   = help: the trait `std::ops::Fn<(&'a str,)>` is not implemented for `std::result::Result<(&str, O), nom::Err<E>>`
   = note: the return type of a function must have a statically known size
```

From what I can tell, the file `tests/json.rs` has an actual working implementation, and the docs should be replaced with this. But I'm also not sure why this isn't even built in to Nom to begin with.

```rust
fn ws<'a, O, E: ParseError<&'a str>, F: Parser<&'a str, O, E>>(f: F) -> impl Parser<&'a str, O, E> {
  delimited(multispace0, f, multispace0)
}
```

Geal:
the function was returning a `Fn` instead of a `FnMut`, it will compile fine now.
@SeanMcLoughlin it is not built into nom because if used in multiple places in a parser, you end up calling the space parser for nothing in a lot of places, and then I get suport request where people wonder about performance issues :neutral_face: 
A better way would be to parse whitespace once at the beginning of the main parser, then use a combinator that ad whitespace parsing as suffix only. Or even write a lexer (I should probably make an example about that one)

PR #1238, opened 2020-11-11
Add D-Bus parser to README

toshokan:
I recently completed a parser for the D-Bus wire format using nom. 
I added it to the README in case it's a useful reference to someone.


Geal:
nice work!

PR #1242, opened 2020-11-12
Docs: fix typo in `ws` recipe

mayabyte:
There is no `i` defined in this function; it seems like a leftover from some of the other recipes on this page.
Copy-pasting the function as it currently is results in a compiler error, but works as expected when `(i)` is removed.


Geal:
Thanks! It was fixed in #1234

PR #1243, opened 2020-11-12
Fix feature documentation for count

fkloiber:
Documentation for `multi::count` did not have the marker indicating its dependence on the `alloc` feature. This PR adds it.


Geal:
Thanks!

PR #1244, opened 2020-11-13
Remove whitespace parsers from choosing a combinator docs

shssoichiro:
These parsers were removed in 6.0.0 and should be removed from all documentation.


Geal:
Thanks!

Issue #1251, opened 2020-11-25
`eof` combinator is more constrained than necessary


meiomorphism:
(As seen in nom 6.0.x. It looks like this also applies to the `eof!` macro, though I haven't used that.)

When the `eof` combinator succeeds, it returns a copy of its input, with the return type `IResult<I, I, E>`. Unfortunately, this isn't actually useful data — it'll always be an empty slice, if I'm reading it right — and it adds the requirement that the input type implement `Copy`. If `eof` instead were to return `IResult<I, (), E>`, then the `I: Copy` restriction could be removed.

Unfortunately this _would_ be a breaking change, and so would necessitate a major version bump; but it might be reasonable to deprecate it and direct users to an equivalent replacement function (`end_of_input`, perhaps?) until nom 7.

PR #1252, opened 2020-11-25
Relax `Copy` bounds on eof and many1

meiomorphism:
Relaxes the `Copy` requirement on the input types of `eof`, `eof!`, and `many1!` to `Clone`, to match other combinators.

The changelog notes that there was some work done to make all the combinators work with non-`Copy` types for 3.0.0; presumably this is still a desired property.

Related to #1251, but much less intrusive than the initial suggestion there.


Geal:
thanks!

PR #1260, opened 2020-12-15
Correctness & style fixes to `fold_many_m_n` and `many_m_n`.

Lucretiel:
This PR makes the following fixes & chanes:

- `fold_many_m_n` now correctly attaches the error from the subparser with `ParseError::append`.
- `fold_many_m_n` now correctly reports the location of a subparser error. Previously it would always report the location as the front of the *original* input, which inconsistent with all the other `multi::*` parsers.
- Updated tests for the above changes.
- Follows the clippy recommendation to use more descriptive names for the parameters and variables for `fold_many_m_n` and `many_m_n`. I followed clippy's advice here because these variable names are the likely cause of the incorrect error behavior in `fold_many_m_n` (which returned `E::from(input, ...)` instead of `E::from(i, ...)`).
- `many_m_n` now uses a for loop, simplifying its implementation and making it consistent for `fold_many_m_n`.
- Removed an unnecessary clone in `fold_many_m_n` and `many_m_n`.

### DIFF
--- a/src/multi/macros.rs
+++ b/src/multi/macros.rs
@@ -519,7 +519,7 @@ macro_rules! fold_many1(
 ///  let b = b"abcdabcdefgh";
 ///  let c = b"abcdabcdabcdabcdabcdefgh";
 ///
-///  assert_eq!(multi(&a[..]), Err(Err::Error(error_position!(&a[..], ErrorKind::ManyMN))));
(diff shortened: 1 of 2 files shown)


Lucretiel:
Closed because I moved to a different branch

PR #1263, opened 2020-12-20
impl Parser for Box<dyn Parser>

Lucretiel:
This PR implements `Parser` for `Box<dyn Parser>`, allowing more dynamic parsers to be constructed at runtime, if necessary.


Geal:
good idea, that was needed

PR #1265, opened 2020-12-26
Allow `re_match*` to take either `Regex` or `&Regex`

RustyYato:
No related issue, I found a users on [URLO](https://users.rust-lang.org/t/using-lazy-static-with-regex-in-noms-re-match/53255) who had trouble using `re_match` with a `Regex` in a `lazy_static`. This forces them to use `&Regex` at most, and it would be easiest to pass in a `&Regex`. Since `Regex: !Clone` there's no easy to convert a `&Regex` to a `Regex`. Given that it's recommended to stuff a `Regex` in a `lazy_static` or `once_cell`, I think it's best to provide an API that allows passing in a reference for compatibility with this recommendation.

I've made 1 change to all of the `re_matches` functions: add a generic parameter `R: Borrow<Regex>` and make the input type `R`.

For example:
```diff
-  pub fn re_match<'a, E>(re: Regex) -> impl Fn(&'a [u8]) -> IResult<&'a [u8], &'a [u8], E>
-  where
+  pub fn re_match<'a, R, E>(re: R) -> impl Fn(&'a [u8]) -> IResult<&'a [u8], &'a [u8], E>
+  where
+    R: Borrow<Regex>,
     E: ParseError<&'a [u8]>,
   {
```

Personally, I would have preferred using `AsRef`, but given that `Regex: !AsRef<Regex>`, this wouldn't be backward compatible and would prevent some use cases where one needs to store the parser across multiple runs. So I used `Borrow` instead, which has the same intended effect, allow both `Regex` and `&Regex`.

### DIFF
--- a/src/regexp/mod.rs
+++ b/src/regexp/mod.rs
@@ -6,6 +6,7 @@ mod macros;
 pub mod str {
   use crate::error::{ErrorKind, ParseError};
   use crate::lib::regex::Regex;
+  use crate::lib::std::borrow::Borrow;
   #[cfg(feature = "alloc")]
   use crate::lib::std::vec::Vec;
   use crate::traits::{InputLength, Slice};
@@ -22,7 +23,7 @@ pub mod str {
   /// # use nom::regexp::str::re_match;
   /// # fn main() {
   /// let re = regex::Regex::new(r"^\d{4}").unwrap();
-  /// let parser = re_match::<(&str, ErrorKind)>(re);
+  /// let parser = re_match::<_, (&str, ErrorKind)>(re);
   /// assert_eq!(parser("2019"), Ok(("", "2019")));
   /// assert_eq!(parser("abc"), Err(Err::Error(("abc", ErrorKind::RegexpMatch))));
   /// assert_eq!(parser("2019-10"), Ok(("", "2019-10")));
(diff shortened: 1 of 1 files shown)


CAD97:
As written, this is still technically breaking (and not "allowed breakage" such as inference breakage), as people could have named `re_match` with a turbofish. Because [function generic arguments cannot be defaulted](https://github.com/rust-lang/rust/issues/36887), there's no way to add a generic argument to a function without it being a breaking change.

If `re_match` is primarily intended to be used without a turbofish, you could deem this to be "not breaking enough" and publish it anyway. Given the example uses a turbofish, though, I'd have a hard time buying it.

I think `Borrow` is actually the (closer to) the right trait here (over `AsRef`), though, because `re_match` is returning an `impl Fn` that owns its argument. So you could e.g. `lazy_static!` the `re_match` parser, or store it in some other reusable block, rather than calling `re_match` each time.

RustyYato:
@CAD97 yes, I tried to minimize breakage where possible. Should have been more clear about that.

SkiFire13:
> As written, this is still technically breaking (and not "allowed breakage" such as inference breakage), as people could have named re_match with a turbofish. Because function generic arguments cannot be defaulted, there's no way to add a generic argument to a function without it being a breaking change.

What about `impl Borrow<Regex>`? It should act like a generic but it's not in the generic arguments list, so it won't break turbofish usages.

RustyYato:
@SkiFire13 you can't set any generic parameters if you use the `impl Trait` syntax.

i.e.

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b47f2ff396d9b8cb4e656401e80f556f)
```rust
use regex::Regex;
use std::borrow::Borrow;

fn foo<E>(_: impl Borrow<Regex>) {}

fn test() {
    foo::<i32>;
}
```

gives the error

```rust
error[E0632]: cannot provide explicit generic arguments when `impl Trait` is used in argument position
 --> src/lib.rs:7:11
  |
7 |     foo::<i32>;
  |           ^^^ explicit generic argument not allowed

error: aborting due to previous error
```

SkiFire13:
TIL. I feel like this is something that the compiler could allow when the type inference can figure out the actual types of the `impl Trait`s

Geal:
the regex combinators will now live in https://github.com/rust-bakery/nom-regex

PR #1268, opened 2020-12-28
Make documentation for delimited less confusing

drwilco:
The documentation for `delimited` is currently confusing, which seems to stem from having been copy/pasted from `separated_pair` but not properly updated. This PR fixes that, following the example of how `preceded` and `terminated` are documented.


Geal:
thank you, it's much better now :)

Issue #1144, opened 2020-04-18
nom::branch::alt handles at most 21 parsers


luca-i:
Hi, I'm not sure this is a bug, maybe it is a current limitation not documented.
I think `nom::branch::alt` can currently handle only 21 parsers.

rustc 1.42.0 (b8cedc004 2020-03-09)
nom = "5.1.1"

```rust
use nom::branch::alt;
use nom::bytes::complete::tag;
use nom::IResult;

fn match_alternatives(input: &str) -> IResult<&str, &str> {
    alt((
        tag("alternative_01"),
        tag("alternative_02"),
        tag("alternative_03"),
        tag("alternative_04"),
        tag("alternative_05"),
        tag("alternative_06"),
        tag("alternative_07"),
        tag("alternative_08"),
        tag("alternative_09"),
        tag("alternative_10"),
        tag("alternative_11"),
        tag("alternative_12"),
        tag("alternative_13"),
        tag("alternative_14"),
        tag("alternative_15"),
        tag("alternative_16"),
        tag("alternative_17"),
        tag("alternative_18"),
        tag("alternative_19"),
        tag("alternative_20"),
        tag("alternative_21"),
        // tag("alternative_22"), // uncomment to get compiler error
    ))(input)
}

fn main() {}

#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn test_alternatives() {
        assert_eq!(
            match_alternatives("alternative_21"),
            IResult::Ok(("", "alternative_21"))
        );

        // to match "alternative_22" you must uncomment the
        // appropriate tag inside match_alternative, but at that moment
        // the code won't compile any more
        assert_eq!(
            match_alternatives("alternative_22"),
            IResult::Ok(("", "alternative_22"))
        );
    }
}
```

In order to get the compiler error you have to uncomment the following line inside `match_alternatives` function:

```rust
// tag("alternative_22"), // uncomment to get compiler error
```

The error message to me is a bit unclear:

[...]

Issue #994, opened 2019-07-13
Alt with more than 21 elements


rgiot:
In nom 5.0, the `alt` function is limited to 21 elements.
In my use case, I have far more elments that that.
What is the best strategy to handle this case without manually generating larger cases ?

On my side I may code a function fed with a slice of combinators, but maybe nom deserves such kind of function.

Havvy:
Can you put another alt as an element of the alt?

rgiot:
@Havvy Indeed I have finally solved my issue with this trick instead of written another function. The issue is that I had to do it manually

Geal:
unfortunately, to implement `alt` I had to choose an arbitrary limit to the number of parsers. I could raise that number, but there will always be someone who runs into the limit. Meanwhile, `alt` is associative, so it's easy to combine multiple `alt` parsers

PR #1269, opened 2020-12-31
Alt trait for array

snylonue:
Currently the `branch::alt()` only deals with tuples with at most 21 elements(#1144, #994) , so I write this pr trying to change it.

Trait `Alt` has been implemented for `&'a mut [P] where P: Parser` as a result, but due to the limition of my ability and rust compiler, I failed to implement the trait for `[P]`  and P must have the same type.
For example, the code as  follows failed to be compiled:
```rust
use nom::branch::alt;
use nom::character::complete::{alpha1, digit1};
fn main() {
fn parser(input: &str) -> IResult<&str, &str> {
  alt([alpha1, digit1])(input) // using an array
};
```

### DIFF
--- a/src/branch/mod.rs
+++ b/src/branch/mod.rs
@@ -15,6 +15,30 @@ pub trait Alt<I, O, E> {
   fn choice(&mut self, input: I) -> IResult<I, O, E>;
 }
 
+impl<'a, I: Clone, O, E: ParseError<I>, P: Parser<I, O, E>> Alt<I, O, E> for &'a mut [P] {
+  fn choice(&mut self, input: I) -> IResult<I, O, E> {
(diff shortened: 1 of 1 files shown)


Geal:
I did not implement that trait for arrays for that exact reason. Even if some parsers have the same input, output and error types, their actual type might be different due to heavy use of closures and impl trait returns

snylonue:
I don't think it is a good idea now

LoganDark:
> I did not implement that trait for arrays for that exact reason. Even if some parsers have the same input, output and error types, their actual type might be different due to heavy use of closures and impl trait returns

What about for parsers have no impl types and are completely identical? You are forbidding a use case just because it is uncommon. :(

PR #1272, opened 2021-01-07
Update json.rs - documentation nit

jufajardini:
Hey, I'm trying to get a better understanding of `nom` reading this example, and think the following makes sense:
- delete lines `70` and `71`, as `value` was already described (lines `64 to 66`), and this other description seemed incomplete
- change `true` to `false` in line `75`, as it makes more sense with the code that follows


Geal:
thanks!

PR #1276, opened 2021-01-18
Impl std::error::Error for VerboseError

neithernut:
Users of rust libraries usually expect error types to implement the
`std::error::Error` trait, and many error utilities will naturally only
work with types implementing that trait. However, unless nom's custom
`Error` type, the `VerboseError` did not implement the std trait.

This change introduces an empty implementation, meaning that users will
only benefit from `Display` and `Debug` in the end.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -222,6 +222,9 @@ impl<I: fmt::Display> fmt::Display for VerboseError<I> {
   }
 }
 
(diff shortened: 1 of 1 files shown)


Geal:
thanks. I'll plan to properly implement `Display` on `VerboseError`

PR #1292, opened 2021-03-12
Allow fold_many_m_n to take a FnMut fold function instead of just Fn.

marcianx:
Minor change to allow `fold_many_m_n` to take an `FnMut` fold function just like `fold_many0` and `fold_many1` do.


Geal:
Thanks!

PR #1295, opened 2021-04-05
[Docs] Update/Fix example for bits::bits and add tests

Nukesor:
First of all, thanks for providing and maintaining this library!

This PR is related to #976
Overwrites PR #992

Previously, the example code for `bits::bits` has been ignored and didn't work.

I updated/fixed the example and added some additional inline docs, since some things weren't clear to me.

Furthermore, I took the liberty to move the definition of the generic parameters `E1` and `E2` to the `where` block.
I felt that this makes understanding the relationships between the parameters a lot easier, since they're now closer together.
(I can revert this at any time, if it doesn't fit to your code-style)

Old docs page:

https://docs.rs/nom/6.1.2/nom/bits/fn.bits.html

New docs page:

image

### DIFF
--- a/src/bits/mod.rs
+++ b/src/bits/mod.rs
@@ -18,33 +18,42 @@ use crate::traits::{ErrorConvert, Slice};
 /// away.
 ///
 /// # Example
-/// ```ignore
-/// # #[macro_use] extern crate nom;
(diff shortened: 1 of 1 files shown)


Geal:
great work, thanks!

PR #1301, opened 2021-04-10
not_line_ending: clarify and test behavior (#1273)

### LINKED ISSUE #1273: not_line_ending fails with lone '\r' but doesn't document this
Rust version: 1.48.0
nom version: 6.0.1

The documentation of `nom::character::complete::not_line_ending` states that it

> Recognizes a string of any char except '\r' or '\n'.

In fact, if the input has a carriage feed ('\r') not followed by a line break ('\n') the parser fails, even if it subsequently followed by a valid line break:

```
extern crate nom;

use nom::IResult;

fn not_line_ending(s: &str) -> IResult<&str, &str> {
    nom::character::complete::not_line_ending(s)
}

fn main() {
    assert_eq!(not_line_ending("xxx\ry\nz"), Ok(("\ry\nz", "xxx")));
    // Actual result: Err(Error(Error {input: "xxx\ry\nz", code: Tag}))
}
```

Of course, a lone '\r' in a string is badly formed text and it makes sense to treat it as exceptional. Since the behavior that implemented is more complicated than the behavior that was documented I expect it is deliberate. If that's the case the documentation should be updated.

homersimpsons:
Fixes #1273

### DIFF
--- a/src/character/complete.rs
+++ b/src/character/complete.rs
@@ -151,21 +151,24 @@ where
 }
 
 //FIXME: there's still an incomplete
-/// Recognizes a string of any char except '\r' or '\n'.
+/// Recognizes a string of any char except '\r\n' or '\n'.
 ///
 /// *Complete version*: Will return an error if there's not enough input data.
 /// # Example
 ///
 /// ```
-/// # use nom::{Err, error::ErrorKind, IResult, Needed};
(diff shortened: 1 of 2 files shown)


Geal:
Great, it's good to clarify behaviour, thanks

PR #1304, opened 2021-04-18
Update to bitvec 0.22

myrrlyn:
No behavior changes, just a few symbol renames.

Sorry about the mess with `funty`; hewing so closely to names in `core` is always a fun time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ include = [
 ]
 
 [features]
-alloc = ["bitvec/alloc", "funty"]
+alloc = ["bitvec/alloc"]
 std = ["alloc", "bitvec/std", "memchr/use_std"]
 default = ["std", "bitvec", "lexical"]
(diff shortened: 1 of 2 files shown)


myrrlyn:
Just saw you have an MSRV of 1.44! Oops!

I am keeping my 1.51 MSRV; *lowering* is non-negotiable (array const-generics) but I do not know of any reason I will be *raising* it before 1.0 at the end of June. You can leave this PR idling until you're ready to raise MSRV to 1.51.

You should be able to patch to `0.19.4` and lose `funty`. However, this PR contains the symbol changes you'll need to advance past the 0.19 series, including to 1.0 when I complete it. I will *not* be changing the APIs used here between 0.22 and 1.0

collinc97:
Any updates on this?

Geal:
merged in 03a1025, thank you for your help

PR #1312, opened 2021-05-19
Hide duplicated content of std in the docs caused by std re-export

kornelski:
I'm having difficulty searching nom's documentation, because it re-exports a large chunk of `std`, so whatever I search for, I get dozens of `std` results unrelated to `nom`.

This hides internals of the `std` re-export, so that nom's documentation search has mainly nom-related results.

The top-level `std` symbol remains visible in the docs.


Geal:
Nice, this will make the docs more usable, thank you

PR #1320, opened 2021-06-15
fix IResult’s declaration in top-level documentation

ronan-d:
There’s a documentation code block that was left behind when `nom::error::Error` was introduced. This PR updates it to match `IResult` current definition.


Geal:
Nice catch, thanks

Issue #1323, opened 2021-06-19
nom 7.0 roadmap


Geal:
no real plan for now, what do we do?

- update bitvec
- update lexical-core (move to minimal-lexical?)
- update memchr
=> upgrading those dependencies will raise the MSRV. Maybe separate bit parsing in another crate to keep MSRV low?

removing macros?

djc:
I think bitvec should definitely not be part of the default features, and I think moving it into a separate crate would make a lot of sense. See also https://github.com/xd009642/tarpaulin/issues/756 (many of my coverage CI jobs are currently failing because of this issue) and https://github.com/djc/askama/issues/446.

Geal:
other idea: more support for text parsing. nom has been usable for that for a long time, but the experience could be improved:
- better error types. VerboseError is a step in that direction, but there are better solutions now, like [nom-supreme::error](https://docs.rs/nom-supreme/0.4.4/nom_supreme/error/index.html) or https://docs.rs/dangerous/0.2.0/dangerous/error/index.html
- good input types like [nom_locate](https://github.com/fflorent/nom_locate)
- error recovery: https://eyalkalderon.com/blog/nom-error-recovery/
- using grammar descriptions: https://github.com/rust-bakery/nom-peg
- handling precedence in expressions (something like https://hackage.haskell.org/package/parsec-3.1.14.0/docs/Text-Parsec-Expr.html would be nice)
- easy ways to work with a token stream or even an AST as input type
- reintroduce whitespace parsers

This could live in nom, or in a separate crate with a faster release cycle (that could be a crate that aggregates the other solution in one nice package).

a related goal I'm interested in would be to help in the interaction with configuration file formats, especially being able to parse a config file, modify it, serialize it (with cookie-factory) and the file's content would stay stable.

meritozh:
+1 for `nom_locate` and better error types.

eaglgenes101:
I'd also add to the list a type implementing `nom`'s traits such that parser output can outlive the original input string. I don't know if it's within scope, but that was the main pain point I experienced when I was trying to use `nom` for my parsing.

djc:
[...]

Issue #1287, opened 2021-02-15
Remove default dependency on bitvec


djc:
The bitvec crate does not appear to be fundamental to nom, yet it's enabled by default. It also brings in several rarely-used dependencies on its own (funty, radium, tap, wyz). Even if I disable default features, I cannot currently reenable `std` features without also enabling the `bitvec` feature (hopefully [weak feature deps](https://github.com/rust-lang/cargo/issues/8832) are coming soon). In the meantime, I think it would be better if `default`, `alloc` and `std` did not enable `bitvec` by default -- instead we could provide a separate `bitvec` feature and potentially `bitvec-alloc`/`bitvec-std` flags (or just leave those to downstream crates to manage).

Unfortunately I suppose this would be a semver-breaking change, but given the breakage today due to `bitvec` dependencies, it would maybe be nice to get this done sooner rather than later.

djc:
(I'm happy to do the legwork on this if we're aligned on the goals/timeline.)

Geal:
there's no defined timeline for the next major version of nom, but I'd be happy to see a better way around cargo features. I'll check if there's other breaking changes I need to pass at the same time

tweksteen:
FYI, this issue is also relevant for Android (AOSP). We try to minimize the number of crates we rely on and therefore carefully enable features to limit new dependencies. We have [imported Dirkjan's patch](https://android.googlesource.com/platform/external/rust/crates/nom/+/f3f7e4ebc152cb84e657db408f4d90fec44c20f5), until the next release of nom occurs.

Geal:
bitvec input types are now in the nom-bitve, since the release of nom 7: https://crates.io/crates/nom/7.0.0

PR #1325, opened 2021-06-21
Remove bitvec from default features

djc:
As discussed in #1287. That said, the approach suggested in #1323 would probably make more sense in the end.


Geal:
LGTM. Maybe bitvec-alloc could require alloc? (and bitvec-std require std)

djc:
Should we first discuss in #1323 whether the bitvec parsers will remain part of the main crate? If so, this PR doesn't make sense.

Geal:
it makes sense in that I can push a 7.0.0-alpha with that fix right now, which would unlock the tarpaulin issue, and then we can see how to move bitvec in a separate crate for the real release (yes this is kinda cheating with semver)

djc:
Done.

Geal:
thanks

Geal:
@djc https://crates.io/crates/nom/7.0.0-alpha1 (it also includes bitvec upgrades for now)

PR #1327, opened 2021-06-21
docsrs: fix build

homersimpsons:
This should fix the docsrs build issues for 6.2.0

The following command works thanks to this fix:
```sh
cargo +nightly doc --features docsrs,default
```

For reference the build logs can be found there https://docs.rs/crate/nom/6.2.0/builds/401812


Geal:
great, thanks!

PR #1328, opened 2021-06-21
coverage: fix issue with `skip` and `tarpaulin`

homersimpsons:
Looking at https://github.com/xd009642/tarpaulin/issues/756 we could fix coverage generation with the following changes

In local the following commands work, so it should on CI too:
```sh
$ cargo install cargo-tarpaulin --version 0.18.0-alpha3
$ cargo tarpaulin -o Html
```


Geal:
thanks, this should fix the build until the bitvec situation improves

PR #1335, opened 2021-07-08
Add variations to `fold_many` functions that don't require `Clone` on resulting type

sky1e:
Then modify the related macros to use them so that the macros can be used without clone

I and a friend were trying to use Nom in our code to write a parser and were frustrated when it required that our parsed type implement `Clone`, when there didn't seem like there needed to be any reason to clone it. I took a look inside `fold_many0!` macro and realized that it didn't need to, so I set out to change it.

`fold_many0` the function, in its current form, requires cloning the initial value so that the parser closure it returns can be called multiple times with different inputs. `fold_many0` the macro, however, only calls that closure once, so it could be an `FnOnce` instead of an `FnMut`. This would allow it to not need to clone its initial input.

The existing function `fold_many0` is already public and changing its return type to an FnOnce would be a breaking change, so instead I add a new variation of the function, and factor out common implementation details into a private helper.

# Other approach
One other possible interface would be to have the functions take in a repeatably callable closure that generates the initial value. This would also work well for non-Clone types if they have a fairly simple to generate initial value, and would allow returning a repeatably callable FnMut, but if the initial value came from some arbitrarily complex sub-expression this wouldn't work well.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -733,12 +733,14 @@ where
   E: ParseError<I>,
   R: Clone,
 {
-  move |input: I| fold_many_impl(&mut f, init.clone(), &mut g, input)
+  move |input: I| fold_many_impl(input, &mut f, init.clone(), &mut g)
 }
 
-/// Like `fold_many0` except the accumulated type does not have to implement `Clone` and the returned
-/// closure can only be called once.
+/// Like `fold_many0` except takes input immediately and the parsed type does not need to implement
+/// `Clone`. Can be slightly more efficient than `fold_many0` if used once because it doesn't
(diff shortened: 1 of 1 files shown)


sky1e:
I am open to suggestions for better function names, as well as suggestions for how to modify the old and new example code to show the differences between the old and new functions

sky1e:
Thinking about this further, I realize there's probably no reason the `*_once` functions need to return an `FnOnce` at all instead of just taking in an input immediately. i'm working on a change to this pull request to do that, that I'll push to branch when ready.

Geal:
the combinators were fixed in #1354

PR #1337, opened 2021-07-14
lazy_many1 handles Incomplete error with a success if there is at least one match

agourlay:
This PR changes `many1` to return a success if there is at least one match in case of `Err::Incomplete`.

## Motivation

When working with a streaming parser, it is convenient to parse as many items as possible for a given buffer.

Unfortunately `many1` does not enable a user to process the items as there are found because it requires to have a buffer large enough to capture the entire sequence of items.

## Change

With this change, it is possible to process incrementally a streaming parser using `many1`.

If at least one match has been captured so far, `many1` returns it with the appropriate rest so the user can parse the entire dataset by repeating the following process:

1. init or extend buffer
2. `many1` returns matches found with rest
3. `many1` fails because zero matches are found
4. back to 1

## Notes

I understand that this is a breaking change so I would be happy to use another combinator to achieve the same result if necessary.

For the time being I am using a copy of the function in my [project](https://github.com/agourlay/hprof-slurp/blob/master/src/record_parser.rs#L108) and it is working just fine so there is no hurry.

Thank you for this awesome library ❤️

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -109,7 +109,7 @@ where
       loop {
         let len = i.input_len();
         match f.parse(i.clone()) {
-          Err(Err::Error(_)) => return Ok((i, acc)),
+          Err(Err::Error(_)| Err::Incomplete(_)) => return Ok((i, acc)),
           Err(e) => return Err(e),
           Ok((i1, o)) => {
             // infinite loop check: the parser must always consume
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -109,6 +109,69 @@ where
       loop {
         let len = i.input_len();
         match f.parse(i.clone()) {
+          Err(Err::Error(_)) => return Ok((i, acc)),
+          Err(e) => return Err(e),
+          Ok((i1, o)) => {
+            // infinite loop check: the parser must always consume
+            if i1.input_len() == len {
+              return Err(Err::Error(E::from_error_kind(i, ErrorKind::Many1)));
+            }
+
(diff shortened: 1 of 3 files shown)


Geal:
this should probably live in another combinator, because it breaks the core assumption in streaming parsers here: there are two possible results, Incomplete or the produced value, and adding more input data should not produce a different result.
Maybe something like `lazy_many1`?

agourlay:
@Geal I have implemented you suggestion in https://github.com/Geal/nom/pull/1337/commits/e1d6ad47b53a28f71f4f0af58d9311972bccd66b

`lazy_many1` is now independent from `many_1` to avoid breaking users code.

src/multi/mod.rs
```diff
@@ -126,6 +126,69 @@ where
...
+    Err(e) => Err(e),
+    Ok((i1, o)) => {
+      let mut acc = crate::lib::std::vec::Vec::with_capacity(4);
+      acc.push(o);
+      i = i1;
+
+      loop {
+        let len = i.input_len();
+        match f.parse(i.clone()) {
+          Err(Err::Error(_)| Err::Incomplete(_)) => return Ok((i, acc)),
```
agourlay, line 175: Matching on `Err::Incomplete(_)` is the only difference with the `many_1` impl.
Not sure if this deserves a comment.

agourlay:
Seems to be a pretty niche usage after all, I am fine with having a dedicated combinator in my code base.
Thanks again for this library :+1:

PR #1339, opened 2021-07-22
Add to_owned for Err<Error<&[u8]>>, Err<Error<&str>>

### LINKED ISSUE #1254: Err::to_owned uses old tuple instead of new error::Error type
- Rust version : `rustc 1.48.0-nightly (623fb90b5 2020-09-26)`
- nom version : `6.0.1`
- nom compilation features used: `[]`

The `Err::to_owned` uses the old IResult definition. Both the implementations for `&str` and `&[u8]` should probably be rewritten for the updated IResult.

Current version:
https://github.com/Geal/nom/blob/bd1adeb8cee593dfedb971df4f5aa26b8a011578/src/internal.rs#L153-L171

andersk:
When commit 981d036ca8dbc373b8b3c74293491a8c21e18f63 changed the basic error type, it left no way for the user to take ownership of the values of this new type.

Fixes #1254.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -150,6 +150,26 @@ impl<T> Err<(T, ErrorKind)> {
   }
 }
 
+impl<T> Err<error::Error<T>> {
+  /// Maps `Err<error::Error<T>>` to `Err<error::Error<U>>` with the given `F: T -> U`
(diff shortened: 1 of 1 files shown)


Geal:
thanks!

Issue #1343, opened 2021-08-03
removing macros


Geal:
nom 7 will finally remove the old parser macros, to focus on functions and traits. Let's synchronize work on macro removal in that issue

Geal:
I'm currently working on the `tests/` directory first

nickelc:
> I'm currently working on the `tests/` directory first

i just finished mp4

Geal:
oh well, thanks :)

Geal:
going through `src/branch/` now

Geal:
since most `macros.rs` files contain the tests, I'll rename the file to `tests`rs` after removing the macros

nickelc:
> since most `macros.rs` files contain the tests, I'll rename the file to `tests.rs` after removing the macros

ok, i'll do that with `src/characters/` too

Geal:
maybe do sequence first, since parsers in char will probably affect a lot of other tests

nickelc:
> maybe do sequence first, since parsers in char will probably affect a lot of other tests

yeah, makes sense. I'll stick to replacing the macros with functions before remove anything

Geal:
done for `branch/` in e95faa4, I'm doing `multi/` now

nickelc:
i have to pause/stop for now and marked my pr as ready for review.

Geal:
alright, thanks

Geal:
done for `multi/` in ff95cff, stopping for now

Geal:
going through `sequence/` now

nickelc:
i take `bytes/`

djc:
Sorry that I don't have much time to actually contribute code, just know that I'm cheering for all of your hard work!

Geal:
it's done for `sequence/` and `bytes/`, I'm taking `character/` now.

no worries @djc , you gave a good push to that effort :)

Geal:
now doing `combinator/`

Geal:
combinator is done too. According to a quick grep on `macro_rules`, we went from 140 macros to 27 :partying_face:

Geal:
none of the remaining ones are parsers though, they are mainly there for internal implementation details

nickelc:
I noticed that with the removal of the `parse_to` macro `ErrorKind::ParseTo` is no longer used.

Geal:
removed in a2b683b

Geal:
released in nom 7: https://crates.io/crates/nom/7.0.0

PR #1345, opened 2021-08-03
Remove macros from ini benches

nickelc:
Cleans up the ini benches (copied tests, obsolete ini_complete, fix deprecation warnings)

related: #1343


Geal:
maybe I should have rewritten the ini/ini_complete benchmarks to show the differences, but right now the streaming/complete versions of parsers do not make much difference

nickelc:
i compared ini/ini_complete files and i have not seen a difference.

PR #1349, opened 2021-08-04
Remove more macros from tests

nickelc:
I went through all issue tests and removed all macro related issue tests.

`tests/` is now done and green.

related: #1343


Geal:
awesome, thanks!

PR #1354, opened 2021-08-08
Remove Clone trait bound from fold parsers

### LINKED ISSUE #1008: `fold_many1` (and some others) now requires `Clone` attribute
- Rust version : 1.36
- nom version : 5.0

`fold_many1` now requires init/return value to have `Clone`, which is annoying when it can be a complex type (and you don't want to clone it).

It was not the case with nom 4.

## Test case

```rust
struct TypeWithoutClone {
    pub a: u32,
}

named!(multi<&[u8], Vec<TypeWithoutClone>>,
    fold_many1!( tag!( "abcd" ), Vec::new(), |mut acc: Vec<_>, item| {
        acc.push( TypeWithoutClone{ a: 0 } );
        acc
    })
);
```

Nom 4 build: no error

Nom 5 build:
```
error[E0277]: the trait bound `TypeWithoutClone: std::clone::Clone` is not satisfied
  --> src/lib.rs:14:1
   |
14 | / named!(multi<&[u8], Vec<TypeWithoutClone>>,
15 | |     fold_many1!( tag!( "abcd" ), Vec::new(), |mut acc: Vec<_>, item| {
16 | |         acc.push( TypeWithoutClone{ a: 0 } );
17 | |         acc
18 | |     })
19 | | );
   | |__^ the trait `std::clone::Clone` is not implemented for `TypeWithoutClone`
   |
   = note: required because of the requirements on the impl of `std::clone::Clone` for `std::vec::Vec<TypeWithoutClone>`
   = note: required by `nom::multi::fold_many1c`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

cenodis:
This PR fixes #1008.

It removes the `R: Clone` bound for the initial acculumators of the `fold_*` family of nom parsers. This is accomplished by using a supplier function `FnMut() -> R` that returns a new accumulator instead of needing the parser to clone a new one.

This allows `fold_*` to be used even if the resulting type is not cloneable. Cloning the accumulator is still possible if desired but must be done by the supplier function.

This is a breaking change but migration is relatively easy as the intial accumulator can simply be wrapped in a closure/be replaced by a function reference.

### DIFF
--- a/examples/string.rs
+++ b/examples/string.rs
@@ -148,7 +148,7 @@ where
     // Our parser function– parses a single string fragment
     parse_fragment,
     // Our init value, an empty string
-    String::new(),
+    String::new,
     // Our folding function. For each fragment, append the fragment to the
     // string.
     |mut string, fragment| {
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -596,7 +596,7 @@ where
 /// the results using a given function and initial value.
 /// # Arguments
 /// * `f` The parser to apply.
-/// * `init` The initial value.
+/// * `init` A function returning the initial value.
 /// * `g` The function that combines a result of `f` with
 ///       the current accumulator.
 /// ```rust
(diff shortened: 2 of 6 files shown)


Geal:
thanks!

PR #1363, opened 2021-08-15
Add fail parser

cenodis:
This PR introduces a `fail` parser that works as the inverse of the `success` parser. It always fails regardless of what input was passed to it. This complements the `success` parser and allows specifying default cases to return an error instead.

I have come across one use case as part of my precedence parser #1362. It requires three parser for prefix, postfix and binary operators respectively. But some languages might not have any prefix or postfix operators so the respective parser should just fail. 

Compared to a workaround like `verify(tag(""), |_: &str| false)`, `fail` is easier to read and makes the intent more clear when a parser is required (e.g. as part of a parameter) but should just fail because its not needed in the context.

### DIFF
--- a/doc/choosing_a_combinator.md
+++ b/doc/choosing_a_combinator.md
@@ -109,6 +109,7 @@ The following parsers could be found on [docs.rs number section](https://docs.rs
 ## Remaining combinators
 
 - [`success`](https://docs.rs/nom/latest/nom/combinator/fn.success.html): Returns a value without consuming any input, always succeeds
+- [`fail`](https://docs.rs/nom/latest/nom/combinator/fn.fail.html): Inversion of `success`. Always fails.
 
 ## Character test functions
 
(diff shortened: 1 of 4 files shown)


Stargateur:
this return unrecoverable error, since nom have 2 level or failure I would name this function, unrecoverable, not fail. Specially since the `Failure` variant of `Err` misleading name.

Stargateur:
Ah, funny I actually again mix up what is what, so this return a recoverable error, thus fail is ok tier, why not full `failure` word like `success` ?

cenodis:
Its not unrecoverable. Its the normal `nom::Err::Error`. Though I guess the name `fail` could be confused with the unrecoverable `Failure` variant. But I dont want yet another thing called "error" in nom considering that there are already 4 or 5 things with the name "error". Any other suggestions?

Stargateur:
Well, `failure` should be ok want we will rename the error type in nom, for not it's probably confusing. Maybe for now `forward` ? Cause I still wonder how to rename `Err::Error` to either `Failure` or `Forward`

Geal:
naming it `fail` will be fine for now

Geal:
thanks!

PR #1364, opened 2021-08-18
Add simple take_one

Stargateur:
Simple equivalent of `anychar()` unfortunately I try to make it generic over input but it's related quite difficult:

- `InputIter` could have a `split_first` but array implementation prevent `fn split_first(&self) -> Option<(Self::Item, Self)>` cause this require to return Self but Self is reduce by one so the returned array is not Self anymore.
- `InputTake` doesn't have Sized to return `Self` and doesn't have `Item` I try to add `Self: InputIter` but I don't know if this would require GAT to compile I didn't succeed.

Anyway, this combinator only make sense for thing that contains u8, so I guess taking a `&[u8]` is ok.

PS: I wonder if the implementation for array of `InputTake` is really useful and not just a bit limiter to what we can do with nom trait.

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -412,6 +412,27 @@ where
   }
 }
 
+/// Returns one byte from input
+///
+/// # Streaming Specific
+/// *Streaming version* will return `Err::Incomplete(Needed::new(1))` if the input is empty
+///
+/// # Example
+/// ```rust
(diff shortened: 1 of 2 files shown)


Stargateur:
I having problem check doc test cause https://github.com/rust-lang/cargo/issues/6424

Stargateur:
right now I feel this is not very useful unless someone know how to make it generic.

PR #1367, opened 2021-08-21
Update README.md

richardanaya:
Minor typo


richardanaya:
oops, realized this was talking about alloc not no_std

Issue #1370, opened 2021-08-23
Nom 7 error in STM32F405


xieren58:
- Rust version :  1.54.0
- nom version : 7.0.0
- nom compilation features used: 
default-features = false
features = ["alloc"]

```
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabihf` target may not support the standard library
  = note: `std` is required by `minimal_lexical` because it does not declare `#![no_std]`
error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: could not compile `minimal-lexical`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

PR #1371, opened 2021-08-23
minimal-lexical now import with no default feature

Stargateur:
Should closes https://github.com/Geal/nom/issues/1370

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,12 +28,13 @@ include = [
 
 [features]
 alloc = []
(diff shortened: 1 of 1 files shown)


Stargateur:
I wanted to add a test for no_std but I don't have a good idea to how do it, github action check it's compile with or without std, but I don't know how to check it in a test.

dne:
Would it be possible to release/tag a version 7.0.1 with this fix?

PR #1376, opened 2021-08-28
Fix documention for double using f64 and float

NilsIrl:
Commit too small for things template to apply.

Note: Maybe ErrorKind::Double should be added


Geal:
thanks

PR #1378, opened 2021-08-31
Remove version check for i128 functions

nickelc:
The rustc version check for the u128/i128 functions is no longer needed.


Xiretza (approved):
LGTM: this feature was enabled for all rust version starting from 1.44; MSRV is 1.48, so this was a no-op for all supported versions!

PR #1379, opened 2021-09-03
Fix docs: misleading argument order (streaming::escaped_transform)

clonejo:
The docs example for streaming::escaped_transform is very misleading, as
it would transform backslash+newline to just n, not backslash+n to
newline, as most people would expect.

This is already correct in the docs for complete::escaped_transform and
the tests, this seems to be an oversight.


Stargateur:
good catch doc is still not very clear for this one

Geal:
thanks

PR #1381, opened 2021-09-04
fix: fixed doc code blocks

saiintbrisson:
Some code blocks inside the `sequence` module had no spacing between code blocks and lines above, making it fail to render as a valid block.


nickelc:
I can't reproduce this issue locally and [docs.rs] shows the examples too.

[docs.rs]: https://docs.rs/nom/7.0.0/nom/sequence/fn.pair.html

Issue #1075, opened 2019-11-26
JSON example issues


Timmmm:
Using Nom 5.0.1 and [the json example](https://github.com/Geal/nom/blob/master/examples/json.rs) I found the following issues:

1. Empty strings can't be parsed.
2. `null` values can't be parsed.
3. Strings can only contain alphanumeric characters!

I know it's only meant to be an example, but those things seem fairly easy to fix. For example:

1. Add null type:

```
#[derive(Debug, PartialEq)]
pub enum JsonValue {
  Null,
  Str(String),
  Boolean(bool),
  Num(f64),
  Array(Vec<JsonValue>),
  Object(HashMap<String, JsonValue>),
}
```

2. Add it to the value:

```
    alt((
      map(tag("null"), |_| JsonValue::Null),
      map(hash, JsonValue::Object),
      map(array, JsonValue::Array),
      map(string, |s| JsonValue::Str(String::from(s))),
      map(double, JsonValue::Num),
      map(boolean, JsonValue::Boolean),
    )),
```

3. Improve `parse_str` (not handling unicode properly - I guess it will parse but won't detect invalid escapes like `\uzzzz`):

```
fn parse_str<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, &'a str, E> {
    escaped(
        take_while1(|c| c != '\\' && c != '"'),
        '\\',
        one_of("\"\\/bfnrtu"), // Note, this will not detect invalid unicode escapes.
    )(i)
}
```

4. Make the "inside" of the string optional (I kind of feel that this shouldn't be necessary - maybe there should be `escaped` and `escaped1`?

```
fn string<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, &'a str, E> {
    context(
        "string",
        preceded(
            char('\"'),
            cut(terminated(
                map(opt(parse_str), |o| o.unwrap_or_default()),
                char('\"'),
            )),
        ),
    )(i)
}
```

That all seems to work for me!

cecton:
I found the same issue while learning nom. Maybe you can make a PR with your proposal already?

Geal:
yup, a PR would be good for that. I did not pretend to write a spec compliant JSON parser, but it could easily be improved

Geal:
FWIW the empty string thing should be fixed in nom 7

cecton:
Wow I completely forgot where I had that issue. But thanks! :grin:

PR #1382, opened 2021-09-07
add null to json example and fix typos

pymongo:
I add Null to JsonValue enum which mention in #1075

I think Unicode string parsing in json is duplicate to the string.rs example, so I didn't add Unicode support for JsonValue


Geal:
could you split the typo fixes in another commit?

PR #1387, opened 2021-09-10
Fix clippy warnings

nickelc:
clippy::unnecessary_cast

### DIFF
--- a/src/bits/complete.rs
+++ b/src/bits/complete.rs
@@ -50,7 +50,7 @@ where
           ErrorKind::Eof,
         )))
       } else {
-        let mut acc: O = (0 as u8).into();
+        let mut acc: O = 0_u8.into();
         let mut offset: usize = bit_offset;
         let mut remaining: usize = count;
         let mut end_offset: usize = 0;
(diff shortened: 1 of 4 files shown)


Stargateur:
why not `0u8` ?

nickelc:
clippy suggests it like this for some time now

```
warning: casting integer literal to `u8` is unnecessary
 --> src/main.rs:3:13
  |
3 |     let a = 0 as u8;
  |             ^^^^^^^ help: try: `0_u8`
  |
  = note: `#[warn(clippy::unnecessary_cast)]` on by default

Stargateur:
I never see such convention, I'm not opposite but I found it strange that clippy advice it, I never see any discussion about that, rustfmt doesn't fmt it like that neither. I think I will open an issue on clippy to ask about that.

The warning is also quite very pedantic, main purpose is to reduce character number... so add an underscore to "be clear" is opposite to the main goal. So I think `0u8` make more sense.

Geal:
thanks. I don't really care much about `0u8` vs `0_u8` to be honest

PR #1390, opened 2021-09-10
Increment minimal-lexical version.

### LINKED ISSUE #1389: Update Minimal-Lexical Version
Due to the use of `no_alloc` as a feature in older  versions of `minimal-lexical`, Nom currently breaks on `no_std` without a system allocator. The patch is quite simple, as the `no_alloc` feature has been removed and replaced with the more appropriate `alloc` feature in v0.2.0.

Alexhuszagh:
This both removes `stackvec` as a transient dependency, since it is no longer necessary, and ensure builds in a `no_std` environment with a system allocator. Fixes #1389. Since the stack-based vector is sufficient, and relatively compact, I've used it as the default in all cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ default = ["std"]
 docsrs = []
 
 [dependencies.minimal-lexical]
-version = "0.1.2"
+version = "0.2.0"
 default-features = false
 
 [dependencies.memchr]


Geal:
awesome, thanks!

vinc:
Hi @Geal, would it be possible to publish a patch release for this issue?

At the moment no_std+alloc crates that use Nom v7.0.0 must point to github instead of crates.io so they can be published. Thanks! 👍

vinc:
Thanks @Geal for the new release! 🙏 🎉

PR #1391, opened 2021-09-10
Fixed #1384, by adding logic to specially handle special floating-point numbers. This should correctly handle both streaming and complete parsers, and the unittests seem to confirm this, but a double-check of my logic would be appreciated.

### LINKED ISSUE #1384: Updating to Nom 7 has broken `double("NaN")`
```
╰─$ rustc -V                                                                                                                                                                           
rustc 1.54.0 (a178d0322 2021-07-26)
```

Given the following code:

```rust
use nom::{error::Error, number::complete::double};
fn main() {
    println!("{:?}", double::<_, Error<_>>("NaN"));
}
```

In Nom version 6,  this runs and returns:

```
Ok(("", NaN))
```

In Nom version 7, this returns:

```
Err(Error(Error { input: "NaN", code: Float }))
```

Maybe this is intentional? I'm not seeing anything in the release notes about it, however.

Alexhuszagh:
Closes #1384, by adding logic to specially handle special floating-point numbers. This should correctly handle both streaming and complete parsers, and the unittests seem to confirm this, but a double-check of my logic would be appreciated.


src/number/streaming.rs
```diff
@@ -1578,6 +1663,79 @@ where
...
+  T: Clone + Offset,
+  T: InputIter + InputLength + InputTake + crate::traits::ParseTo<i32>,
+  <T as InputIter>::Item: AsChar,
+  <T as InputIter>::IterElem: Clone,
+  T: InputTakeAtPosition,
+  <T as InputTakeAtPosition>::Item: AsChar,
+  T: AsBytes,
+  T: for<'a> Compare<&'a [u8]>,
+{
+  let (i, sign) = sign(input.clone())?;
```
Stargateur, line 1680: the sign parsing is done twice in case double_finite fail.

src/number/complete.rs
```diff
@@ -1517,24 +1520,35 @@ where
...
 
+/// Case-insensitive comparison of digits. Only works if `y` is only ASCII letters.
+#[inline]
+fn case_insensitive_cmp(x: &[u8], y: &[u8]) -> bool {
+    let d = (x.iter().zip(y.iter())).fold(0, |d, (xi, yi)| d | xi ^ yi);
+    // This uses the trick that 'a' - 'A' == 0x20, and this is true
+    // for all characters, so as long as `yi` is a valid ASCII letter,
+    // `xi ^ yi` can only be 0 or 0x20.
+    d == 0 || d == 32
+}
```
Stargateur, line 1531: why not write `d == 0 || d == 0x20` ? any source for this trick ?

src/number/streaming.rs
```diff
@@ -1578,6 +1663,79 @@ where
...
+  let (mut double, count) = if case_insensitive_cmp(i.as_bytes(), b"nan") {
+    (f64::NAN, 3)
+  } else if i.input_len() >= 8 && case_insensitive_cmp(i.as_bytes(), b"infinity") {
+    (f64::INFINITY, 8)
+  } else if case_insensitive_cmp(i.as_bytes(), b"inf") {
```
Stargateur, line 1685: Why not use `tag_no_case` ? Maybe store `i.input_len()` into a variable

Stargateur: EBNF grammar from std parsing float is not case insensitive

src/number/streaming.rs
```diff
@@ -1578,6 +1663,79 @@ where
...
+  <T as InputIter>::IterElem: Clone,
+  T: InputTakeAtPosition,
+  <T as InputTakeAtPosition>::Item: AsChar,
+  T: AsBytes,
+  T: for<'a> Compare<&'a [u8]>,
+{
+  let (i, sign) = sign(input.clone())?;
+  let (mut double, count) = if case_insensitive_cmp(i.as_bytes(), b"nan") {
+    (f64::NAN, 3)
+  } else if i.input_len() >= 8 && case_insensitive_cmp(i.as_bytes(), b"infinity") {
```
Stargateur, line 1683: `infinity` is not in the EBNF of rust std

src/number/streaming.rs
Stargateur, line 1685: why this check after the `infinity` one ?

src/number/streaming.rs
```diff
@@ -1578,6 +1663,79 @@ where
...
+    (f64::INFINITY, 3)
+  } else {
+    return Err(Err::Error(E::from_error_kind(input, ErrorKind::Float)));
+  };
+
+  if !sign {
+    double = -double;
+  }
+
+  if i.input_len() <= count {
```
Stargateur, line 1695: that not possible

src/number/complete.rs
```diff
@@ -1603,6 +1690,78 @@ where
...
+  <T as InputIter>::Item: AsChar + Copy,
+  <T as InputIter>::IterElem: Clone,
+  T: InputTakeAtPosition,
+  <T as InputTakeAtPosition>::Item: AsChar,
+  T: AsBytes,
+  T: for<'a> Compare<&'a [u8]>,
+{
+  match double_finite::<T, E>(input.clone()) {
+    Ok((i, double)) => Ok((i, double)),
+    Err(Err::Incomplete(e)) => Err(Err::Incomplete(e)),
```
Stargateur, line 1760: impossible in complete variant

src/number/complete.rs
```diff
@@ -1577,7 +1627,44 @@ where
...
+  T: InputTakeAtPosition,
+  <T as InputTakeAtPosition>::Item: AsChar,
+  T: AsBytes,
+  T: for<'a> Compare<&'a [u8]>,
+{
+  if let Ok((i, float)) = float_finite::<T, E>(input.clone()) {
+    Ok((i, float))
+  } else {
+    float_nonfinite::<T, E>(input)
+  }
```
Stargateur, line 1646: use `or_else` ? or `alt` parser

Stargateur (commented):
Most comment are the same for float double streaming and complete version

src/number/complete.rs
Alexhuszagh: Whoops my bad, thanks for the catch.

src/number/streaming.rs
Stargateur: But in practice it is https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=82d184db3092ef5e95cb62f2a174871d

src/number/streaming.rs
Alexhuszagh: Because if the string is `b"infinity"`, then `b"inf"` would also be true, so we must check the longer overload first.

src/number/streaming.rs
Stargateur: I starting to doubt about the doc of std be up to date

Alexhuszagh: > `infinity` is not in the EBNF of rust std

It is, actually. See below:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d5ac6a9ba62860ebe04f782970f95342

src/number/streaming.rs
Stargateur: oh yes, ofcourse, maybe we could factorize this

Alexhuszagh: @Stargateur Actually it is. There was a defect report and this was fixed (this was addressed as part of my float parsing work). Um, `tag_no_case` is not used here because of the optimization present which works only if one of the two strings is all ASCII letters (described in the comment).

src/number/complete.rs
Stargateur: actually even better write `'a' - 'A'`

Stargateur: or `matches(d, 0 | 'a' - 'A')`

Alexhuszagh: It's used pretty extensively, and sure, I'll modify that. But the logic is quite simple: it's a lot cheaper than `to_ascii_lowercase()`. You can see this [here](https://godbolt.org/z/nYMEKE7ao). Notice just how much more efficient the resulting ASM is:

```asm
example::fast_nocase_nan:
        cmp     rsi, 3
        jb      .LBB0_2
        mov     al, byte ptr [rdi]
        mov     cl, byte ptr [rdi + 1]
        xor     al, 78
        xor     cl, 65
        or      cl, al
        mov     al, byte ptr [rdi + 2]
        xor     al, 78
        or      al, cl
        test    al, -33
        sete    al
        ret
.LBB0_2:
        xor     eax, eax
        ret

example::slow_nocase_nan:
        cmp     rsi, 3
        jb      .LBB1_4
        mov     al, byte ptr [rdi]
        lea     ecx, [rax - 65]
        cmp     cl, 26
        setb    cl
        shl     cl, 5
        or      cl, al
        cmp     cl, 110
        jne     .LBB1_4
        mov     al, byte ptr [rdi + 1]
        lea     ecx, [rax - 65]
        cmp     cl, 26
        setb    cl
        shl     cl, 5
        or      cl, al
        cmp     cl, 97
        jne     .LBB1_4
        mov     al, byte ptr [rdi + 2]
        lea     ecx, [rax - 65]
        cmp     cl, 26
        setb    cl
        shl     cl, 5
        or      cl, al
        cmp     cl, 110
        sete    al
        ret
.LBB1_4:
        xor     eax, eax
        ret
```

The logic is quite simple: as long as one of the two strings is **all** letters, then `xi ^ yi` will only be `0x00` or `0x20` if  both are ASCII letters. Since there are no other bits that could possibly obfuscate the two (`0b100000` and `0b0`), folding with a bitwise or can only produce `0x20` or `0x00` if the result of `xi ^ yi` is `0x20` or `0x00`. In short, there's no cases where this can fail.

src/number/streaming.rs
Alexhuszagh: True, I can move that out. Good suggestion.

src/number/complete.rs
Alexhuszagh: Sure, that works out. If we're moving the sign parsing out though, that might not be possible anymore.

src/number/complete.rs
Alexhuszagh: > 
> 
> or `matches!(d, 0 | 'a' - 'A')`

Sure, but if we're already using magic numbers, we have to rely on **why** this works, which is why the comments document this. `0x20` sounds good though, since it matches exactly with the comment description.

Stargateur: By the way maybe put this function in number module would avoid some code duplication

Alexhuszagh: > By the way maybe put this function in number module would avoid some code duplication

I was actually just about to ask that... Thanks for actually reading my mind ^_^

Alexhuszagh: Speaking of code deduplication @Stargateur, is it worth using macros since it seems like `recognize_float`, etc. all duplicate the exact same code just using `f32` and `f64`? It would only be local to that single module, and would not be exported.

src/number/streaming.rs
Stargateur: even worse I think this make it wrong cause "infinity"  IS complete no ? this mean that if would fail if input is just what we parse.

src/number/streaming.rs
Alexhuszagh: At the very least, maybe I should leave a comment?

src/number/complete.rs
Stargateur: An other solution would be to use num, but meh. Gael is very good with rust macro so I think it's totally ok to use macro for that if you are good at it.

src/number/streaming.rs
Stargateur: add a test would be better in my opinion

src/number/streaming.rs
Alexhuszagh: > even worse I think this make it wrong cause "infinity" IS complete no ? this mean that if would fail if input is just what we parse. `if i.input_len() < count` is what should be write no ? but again, this check is not needed

I believe the check is needed, but I wasn't sure about `i.input_len() < count` (which is why I asked originally), so that sounds right.

Do we have to consider a case where we parse `b"inf"`, and in the streaming API, we don't know if this is complete or incomplete. Because if the following bytes are `b"inity"`, we've produced an incorrect result. The same is true for the float case: if our input is `b"1.2345"` and we've parsed our entire input, we might actually have `b"1.234567"`. I'm not especially well-versed in the internal implementations of nom, so any feedback or guidance is **highly** appreciated.

src/number/streaming.rs
Alexhuszagh: Wait, the one issue with that is our error handling isn't identical in these two cases. The double sign check might be optimized out by the compiler.

If we return `Err(Err::Error(E::from_error_kind(input, ErrorKind::Float)));` relative to the sign, and not relative to the start of the buffer, we get a different result. If the float format is not valid, then I believe the returned input should start before the sign, rather than afterwards. We could also pass both `input` and the relative input to the function, but that's a bit of API clutter.

Generally, special floats tend to be quite rare in real datasets, so performance for the specials is less of a necessity than other cases.

src/number/complete.rs
Alexhuszagh: FYI: `matches!()` is 1.42+, and we currently support 1.41+ (although we're free to increment that). I'm not sure this is worth breaking that, I'll do whatever more senior contributors think is better.

src/number/streaming.rs
Stargateur: Oh now I understand, that a good point, I **think** we should not if I look at doc of [`double()`](https://docs.rs/nom/7.0.0/nom/number/streaming/fn.double.html) for streaming the test show:

```rust
assert_eq!(parser("11e-1"), Ok(("", 1.1)));
assert_eq!(parser("123E-02"), Ok(("", 1.23)));
assert_eq!(parser("123K-01"), Ok(("K-01", 123.0)));
assert_eq!(parser("abc"), Err(Err::Error(("abc", ErrorKind::Float))));
```

That would say we should return Ok if we parsed something even if it "could" be wrong latter. I don't think there is a consensus here, the problem is that return incomplete forever is not good too cause the end of user output could be "inf". In practice there is very few streaming parser and that often "utils" parser and "real" parser often use complete variant. I don't think I see something truly use the streaming power of nom parser cause... that a very difficult like we see it just for something that simple.

So I think we would not return incomplete if we parse something.

Stargateur: But I think we need @Geal input here

src/number/complete.rs
Stargateur: not possible for nom 7 so would be breaking change for nom policy

Alexhuszagh: > not possible for nom 7 so would be breaking change for nom policy. There is a github action to check that anyway.

Ah ok, I thought Nom7 was just guaranteed to support 1.48+, but currently supports 1.41+.

src/number/streaming.rs
Stargateur: AFAIK to nom convention it's totally ok to say the problem is after the sign cause... it is. I think... anyway the only thing I know to debug a nom parser is to use [nom_tracable](https://crates.io/crates/nom-tracable)

src/number/streaming.rs
Alexhuszagh: Ah I found the PR that fixed this for IEEE-754 compliance:
https://github.com/rust-lang/rust/pull/78618

I'm not sure when this made it into stable, however, and it was definitely after 1.51, so it won't compare to the core implementation.

src/number/streaming.rs
Alexhuszagh: This should be resolved now this this change:

https://github.com/Alexhuszagh/nom/blob/805e5be9d7e16a6fc0d86e3e6a16fd042d97fd3f/src/number/streaming.rs#L2185-L2190

src/number/streaming.rs
Alexhuszagh: This has been implemented, the general approach now parses the sign once, and then only handles adding the right sign after all other parsing has been done:

https://github.com/Alexhuszagh/nom/blob/805e5be9d7e16a6fc0d86e3e6a16fd042d97fd3f/src/number/streaming.rs#L1597-L1620

src/number/complete.rs
Alexhuszagh: This has been moved to the module, and added better documentation for it:

https://github.com/Alexhuszagh/nom/blob/805e5be9d7e16a6fc0d86e3e6a16fd042d97fd3f/src/number/mod.rs#L17-L25

src/number/streaming.rs
Alexhuszagh: I changed this to use the same logic as before, because I was having issues with the test suite and backwards compatibility should be favored. Previously, input like `"i"` would return an error, rather than `Incomplete`, and so I've made the special parser act like a complete parser:

https://github.com/Alexhuszagh/nom/blob/805e5be9d7e16a6fc0d86e3e6a16fd042d97fd3f/src/number/streaming.rs#L1507-L1526

src/number/complete.rs
Alexhuszagh: Done, I use the `Result.or` here, since it works exactly with what we want to do. For the streaming parser, I still use a match statement.

https://github.com/Alexhuszagh/nom/blob/805e5be9d7e16a6fc0d86e3e6a16fd042d97fd3f/src/number/complete.rs#L1623-L1625

src/number/complete.rs
Alexhuszagh: Removed as of the latest commit, this should be resolved.

src/number/complete.rs
```diff
@@ -1426,11 +1427,13 @@ where
...
 }
 
+///
+
 /// Recognizes a floating point number in text format and returns the integer, fraction and exponent parts of the input data
 ///
 /// *Complete version*: Can parse until the end of input.
 ///
-pub fn recognize_float_parts<T, E: ParseError<T>>(input: T) -> IResult<T, (bool, T, T, i32), E>
+pub fn recognize_float_parts<T, E: ParseError<T>>(input: T) -> IResult<T, (T, T, i32), E>
```
Stargateur, line 1436: That breaking change unfortunately, if we want to avoid the double check of the sign we will need to factorize this function to not make breaking change.

Alexhuszagh: Is recognize_float_parts part of the public API? If so, I can change that...

src/number/complete.rs
Stargateur: `or` will always call `float_nonfinite`

src/number/complete.rs
Stargateur: apparently, never used it though... I don't even understand why it's pub, https://docs.rs/nom/7.0.0/nom/number/complete/fn.recognize_float_parts.html

src/number/complete.rs
```diff
@@ -1546,30 +1620,62 @@ where
   T: AsBytes,
   T: for<'a> Compare<&'a [u8]>,
 {
-  let (i, (sign, integer, fraction, exponent)) = recognize_float_parts(input)?;
+  let (input, sign) = sign(input.clone())?;
+  let res = float_finite::<T, E>(input.clone());
+  let (i, mut float) = res.or(float_nonfinite::<T, E>(input))?;
```
Stargateur, line 1625: It will always call `float_nonfinite`; "Arguments passed to or are eagerly evaluated; if you are passing the result of a function call, it is recommended to use or_else, which is lazily evaluated." https://doc.rust-lang.org/std/result/enum.Result.html#method.or

Alexhuszagh: Ok fixed as of the latest commit.

src/number/complete.rs
Alexhuszagh: Ok, I guess it is. Then there's no real way to avoid using `sign!()` twice, since we don't return the sign if recognized the float parts errors, so we just have to trust the compiler inlines it.

Alexhuszagh: I wonder if it was just an implementation detail that was exposed publicly, and now can't be reverted?

Alexhuszagh:
@Stargateur Speaking of which, I'm guessing it's a good idea to not mark any internal functions public? I did this somewhat because nom does it a lot internally, but it's probably not a good idea...

Stargateur:
yes 

> 
> 
> @Stargateur Speaking of which, I'm guessing it's a good idea to not mark any internal functions public? I did this somewhat because nom does it a lot internally, but it's probably not a good idea...

Oh, well, I think unless someone have a good reason yeah we should not introduce too many pub item.

Alexhuszagh:
Ok I believe every requested change has been made.

src/number/streaming.rs
```diff
@@ -1490,6 +1490,84 @@ where
...
+
+    Ok((i, float))
+  }};
+}
+
+macro_rules! float_nonfinite {
+  ($input:ident, $t:ident) => {{
+    let (input, sign) = sign($input.clone())?;
+    let b = input.as_bytes();
+    let (mut float, count) = if b.len() >= 3 {
```
Geal, line 1514: this will not parse correctly in streaming: If we get `Na` as partial input it will return an error instead of Incomplete (once we refill the buffer, the next character could be  `N` and we get `NaN`).
The [Compare trait](https://github.com/Geal/nom/blob/master/src/traits.rs#L717) already handles those cases

PR #1394, opened 2021-09-14
Re-enable the `iterator` example

nickelc:
The `iterator` example can be re-enabled because of the current MSRV.


Geal:
great, thanks

PR #1397, opened 2021-09-14
Support `&str` for hex_u32

### LINKED ISSUE #1396: Support `&str` for `number::complete::hex_u32`
## Prerequisites

- Rust version : rustc 1.56.0-nightly (6d64f7f69 2021-08-19)
- nom version : 7.0.0
- nom compilation features used: (default)

## Test case

```rust
use nom::{
  IResult,
  number::complete::hex_u32,
};

fn hex_u32_on_str() {
    let result: IResult<_, _> = hex_u32("a1");
    println!("{:?}", result.unwrap())
}
```

Ideally, I'd like this to compile (or something close).

Currently, we get 
```
error[E0308]: mismatched types
  --> src/protocol/parser.rs:17:41
   |
17 |     let result: IResult<_, _> = hex_u32("a1");
   |                                         ^^^^ expected slice `[u8]`, found `str`
   |
   = note: expected reference `&[u8]`
              found reference `&'static str`
help: consider adding a leading `b`
   |
17 |     let result: IResult<_, _> = hex_u32(b"a1");
   |                                         +
```

Many of the `crate::number::complete` methods do support both &[u8] and &str, so it seems viable here.

nipunn1313:
Fixes #1396 

There are multiple commits here - the first ones can be landed independently from the final one.
If you'd like me to split into multiple PR's, can do. (github doesn't support [dependent PR's](https://github.com/github/feedback/discussions/4477), so I kept them together in one PR for now).

- Refactor hex_u32 to use fold_many_m_n (simpler, but more importantly, will allow it to use `AsBytes` interface, since it only goes through the input once)
- Update hex_u32 to take an `I: AsBytes`, adding a test using a `&str`

### DIFF
--- a/src/number/complete.rs
+++ b/src/number/complete.rs
@@ -1355,31 +1355,32 @@ where
 ///   hex_u32(s)
 /// };
 ///
-/// assert_eq!(parser(&b"01AE"[..]), Ok((&b""[..], 0x01AE)));
-/// assert_eq!(parser(&b"abc"[..]), Ok((&b""[..], 0x0ABC)));
-/// assert_eq!(parser(&b"ggg"[..]), Err(Err::Error((&b"ggg"[..], ErrorKind::IsA))));
+/// assert_eq!(parser("01AE"), Ok(("", 0x01AE)));
+/// assert_eq!(parser("abc"), Ok(("", 0x0ABC)));
+/// assert_eq!(parser("ggg"), Err(Err::Error(("ggg", ErrorKind::HexDigit))));
+/// assert_eq!(parser(""), Err(Err::Error(("", ErrorKind::HexDigit))));
 /// ```
 #[inline]
-pub fn hex_u32<'a, E: ParseError<&'a [u8]>>(input: &'a [u8]) -> IResult<&'a [u8], u32, E> {
-  let (i, o) = crate::bytes::complete::is_a(&b"0123456789abcdefABCDEF"[..])(input)?;
+pub fn hex_u32<'a, I, E: ParseError<I>>(input: I) -> IResult<I, u32, E>
+where
+  I: InputIter,
+  I: Slice<RangeFrom<usize>> + Slice<RangeTo<usize>>,
+  <I as InputIter>::Item: AsChar,
+{
   // Do not parse more than 8 characters for a u32
-  let (parsed, remaining) = if o.len() <= 8 {
-    (o, i)
+  let mut chars = input
+    .iter_elements()
+    .take(8)
+    // Can replace map-take_while-flatten with `.map_while` once stabilized
+    .map(|c| c.as_char().to_digit(16))
+    .take_while(|digit| digit.is_some())
+    .flatten();
+  if let Some(first) = chars.next() {
+    let (num, len) = chars.fold((first, 1), |(acc, len), item| (16 * acc + item, len + 1));
+    Ok((input.slice(len..), num))
   } else {
-    (&input[..8], &input[8..])
-  };
-
-  let res = parsed
-    .iter()
-    .rev()
-    .enumerate()
-    .map(|(k, &v)| {
-      let digit = v as char;
-      digit.to_digit(16).unwrap_or(0) << (k * 4)
-    })
-    .sum();
-
-  Ok((remaining, res))
+    Err(Err::Error(E::from_error_kind(input, ErrorKind::HexDigit)))
+  }
 }
 
 /// Recognizes floating point number in a byte string and returns the corresponding slice.
@@ -1906,7 +1907,11 @@ mod tests {
   fn hex_u32_tests() {
     assert_parse!(
       hex_u32(&b";"[..]),
-      Err(Err::Error(error_position!(&b";"[..], ErrorKind::IsA)))
+      Err(Err::Error(error_position!(&b";"[..], ErrorKind::HexDigit)))
+    );
+    assert_parse!(
+      hex_u32(&b""[..]),
+      Err(Err::Error(error_position!(&b""[..], ErrorKind::HexDigit)))
     );
     assert_parse!(hex_u32(&b"ff;"[..]), Ok((&b";"[..], 255)));
     assert_parse!(hex_u32(&b"1be2;"[..]), Ok((&b";"[..], 7138)));
@@ -1920,6 +1925,8 @@ mod tests {
     assert_parse!(hex_u32(&b"ffffffff;"[..]), Ok((&b";"[..], 4_294_967_295)));
     assert_parse!(hex_u32(&b"0x1be2;"[..]), Ok((&b"x1be2;"[..], 0)));
     assert_parse!(hex_u32(&b"12af"[..]), Ok((&b""[..], 0x12af)));
+
+    assert_parse!(hex_u32("12af"), Ok(("", 0x12af)));
   }
 
   #[test]


src/traits.rs
```diff
@@ -884,33 +884,21 @@ impl<'a> FindToken<u8> for &'a [u8] {
...
     self.find_token(*token)
   }
 }
 
-impl<'a, 'b> FindToken<&'a u8> for &'b str {
-  fn find_token(&self, token: &u8) -> bool {
-    self.as_bytes().find_token(token)
-  }
-}
-
```
Stargateur, line 904: that breaking change

src/number/complete.rs
```diff
@@ -1357,29 +1358,29 @@ where
...
+  fold_many_m_n(
+    1,
+    8,
+    one_of("0123456789abcdefABCDEF"),
+    || 0,
+    |acc, item| {
+      let digit = item as char;
+      (16 * acc) + digit.to_digit(16).unwrap()
+    },
+  )(input)
```
Stargateur, line 1383: that slow, will check twice the same thing

src/number/complete.rs
```diff
@@ -1357,29 +1358,29 @@ where
 ///
 /// assert_eq!(parser(&b"01AE"[..]), Ok((&b""[..], 0x01AE)));
 /// assert_eq!(parser(&b"abc"[..]), Ok((&b""[..], 0x0ABC)));
-/// assert_eq!(parser(&b"ggg"[..]), Err(Err::Error((&b"ggg"[..], ErrorKind::IsA))));
+/// assert_eq!(parser(&b"ggg"[..]), Err(Err::Error((&b"ggg"[..], ErrorKind::OneOf))));
```
Stargateur, line 1361: breaking change ?

src/traits.rs
nipunn1313: Yes - I'm a little concerned by this as a breaking change.

I needed this to solve:
```
error[E0277]: the trait bound `&str: FindToken<<I as InputIter>::Item>` is not satisfied
    --> src/number/complete.rs:1377:5
     |
1377 |     one_of("0123456789abcdefABCDEF"),
     |     ^^^^^^ the trait `FindToken<<I as InputIter>::Item>` is not implemented for `&str`
     |
```

I believe the issue is that the output parser of `one_of` takes an `I` where `I::Item: AsChar`, but it can't figure out how to deal with that with the input as a `&str`. Thus can't generalize the input of `hex_u32` here while still being able to call `one_of`. I solved by having all `FindToken` on `&str` use the `as_chars()` implementation, but it is definitely a breaking change.

Can you help think of a better way which isn't a breaking change (or alternately, of some way to justify the breaking change)?

src/number/complete.rs
nipunn1313: I did notice that this does clone the input (`fold_many_m_n` does this to check errors), though I think that is cheap for `&str` and `&[u8]`.

I'm not sure I understand. Can you help teach me? In what way does it check twice the same thing?

src/number/complete.rs
nipunn1313: Yes, though in this case it does seem like the internal implementation's error type *should* be an implementation detail. Neither `IsA` or `OneOf` seem like the appropriate contract. What's the library's view on breaking changes? Do them and document? We could do some sort of error conversion from to `ErrorKind::HexU32` or something else.

src/number/complete.rs
Stargateur: Something like:

```rust
fold_many_m_n(
    1,
    8,
    |input| digit.to_digit(16).ok_or_else(|| Error::new(input, ErrorKind::HexDigit)),
    || 0,
    |acc, item| {
      (16 * acc) + item
    },
  )(input)
```

src/traits.rs
Stargateur: well your change is not bad could be nice for v8 but now you could just call `as_char` on the item of the iterator

nipunn1313: I was able to eliminate this change w/ a bit of refactoring and using the as_char suggestion. I had to give up on using preexisting `one_of` method, but it turned out ok. Thanks for pushing on this.

src/number/complete.rs
```diff
@@ -1906,7 +1913,7 @@ mod tests {
   fn hex_u32_tests() {
     assert_parse!(
       hex_u32(&b";"[..]),
-      Err(Err::Error(error_position!(&b";"[..], ErrorKind::IsA)))
+      Err(Err::Error(error_position!(&b";"[..], ErrorKind::HexDigit)))
```
nipunn1313, line 1916: This does end up being a breaking change unfortunately - though it does seem like a more appropriate error message here.

We can explicitly return `IsA` for compatibility here, but `HexDigit` seems more appropriate.

src/number/complete.rs
nipunn1313: Thanks! That makes sense. It's checking if the digit is a valid digit more than once. This turned out to be the case in the previous code as well and survived through this iteration. Not sure I buy that the slowness is meaningful, but it is at least inelegant.

Was able to fix via something like your suggestion.

src/number/complete.rs
```diff
@@ -1357,29 +1358,35 @@ where
...
+        Some(c) => c,
+      };
+      match c.as_char().to_digit(16) {
+        Some(digit) => Ok((input.slice(c.len()..), digit)),
+        None => Err(Err::Error(E::from_error_kind(input, ErrorKind::HexDigit))),
+      }
+    },
+    || 0,
+    |acc, digit| (16 * acc) + digit,
+  )(input)
```
nipunn1313, line 1389: I can rewrite this without any call to `fold_many_m_n` by using `input.iter_elements()` and manually enumerating and counting how many hex digits I find on the way, but I thought this was more elegant.

It might make sense to factor out the parsing lambda into its own method (kinda like hex_digit0, but guaranteed to take at most one character).

nipunn1313: I actually gave this rewrite a shot and IMO it looks nicer. Going to update with that.

src/number/complete.rs
```diff
@@ -1355,31 +1355,33 @@ where
...
+{
   // Do not parse more than 8 characters for a u32
-  let (parsed, remaining) = if o.len() <= 8 {
-    (o, i)
+  let chars: Vec<_> = input
+    .iter_elements()
+    // Can replace map-take_while-map with `.map_while` once stabilized
+    .map(|c| c.as_char().to_digit(16))
+    .take_while(|digit| digit.is_some())
+    .map(|c| c.unwrap())
```
Stargateur, line 1376: `flatten()`

src/number/complete.rs
```diff
@@ -1355,31 +1355,33 @@ where
...
+    .take(8)
```
Stargateur, line 1377: take should be after iter_elements

src/number/complete.rs
```diff
@@ -1355,31 +1355,33 @@ where
...
+    .collect();
```
Stargateur, line 1378: collect ? why ?

src/number/complete.rs
```diff
@@ -1355,31 +1355,33 @@ where
...
-    .rev()
-    .enumerate()
-    .map(|(k, &v)| {
-      let digit = v as char;
-      digit.to_digit(16).unwrap_or(0) << (k * 4)
-    })
-    .sum();
-
-  Ok((remaining, res))
+    let num = chars.iter().fold(0, |acc, item| 16 * acc + item);
```
Stargateur, line 1382: Something like that

```rust
if let Some(first) = iter.next() {
    let num = iter.fold(first, |acc, item| 16 * acc + item);
    Ok((remaining, res))
} else {
    Err(Err::Error(E::from_error_kind(input, ErrorKind::HexDigit)))
}
```

Stargateur (changes requested):

src/number/complete.rs
nipunn1313: sure - I *think* it's correct in either order, but it's probably a little more obvious as to why with take() up top. If there's some subtle way in which the behavior is different which I'm missing, lmk and I can add a unit test.

src/number/complete.rs
nipunn1313: Needed it to check the length - however w/ your suggestion below it won't be needed. Will refactor per the `iter.next()` sugestion

src/number/complete.rs
```diff
@@ -1355,31 +1355,32 @@ where
...
-    (o, i)
+  let mut chars = input
+    .iter_elements()
+    .take(8)
+    // Can replace map-take_while-flatten with `.map_while` once stabilized
+    .map(|c| c.as_char().to_digit(16))
+    .take_while(|digit| digit.is_some())
+    .flatten();
+  if let Some(first) = chars.next() {
+    let (num, len) = chars.fold((first, 1), |(acc, len), item| (16 * acc + item, len + 1));
```
nipunn1313, line 1379: I don't love this for readability. It does work.
Might be better to just call collect() on `chars` as before - to collect the up-to-8 chars for the hex number.

Geal:
hi! thank you for looking into this. Unfortunately, I'm closing this in favor of #1602 that addresses both versions of `hex_u32` and has less breaking changes

PR #1399, opened 2021-09-15
Fix doc for fold_* combinators (init arg takes a lambda)

nipunn1313:
The init method for the fold methods requires a lambda rather than the initial value directly.
This updates the doc to match the behavior.

TBH, I'm not certain why it requires a lambda to delay evaluation, since the methods end up calling the `init()` method quickly - but it does!


cenodis:
To answer your question:
It requires a lambda to ensure that the combinator can be called repeatedly. Passing the initial value directly would mean that either the combinator must be `FnOnce` or `init` must implement `Clone`, which might be undesirable in certain situations.
See #1008

nipunn1313:
ooh good catch! Updated. Double checked and it looks good now
image

#1008 makes sense to me! Thanks for digging that up.

Geal:
thanks

Issue #1393, opened 2021-09-13
Consolidate parser variants using ranges (e.g. `many0`, `many_m_n`)


epage:
## Prerequisites

Here are a few things you should provide to help me understand the issue:

- Rust version : 1.44
- nom version : nom7
- nom compilation features used: basic

## Idea

Nom has `many1`, `many0`, and `many_m_n` functions.  Similar with other parts of the API.

What if instead we had a `IntoRange` trait that took in the different range types and single numbers.

Example:
```rust
many(tag("abc"), 0..) -> many0(tag("abc")) 
many(tag("abc"), 1..) -> many1(tag("abc")) 
many(tag("abc"), 1) -> many_m_n(tag("abc"), 1, 1) 
```

cenodis:
I quite like this idea. At the very least this makes the API much cleaner by unifying the very similar parsers.

This could also be done for the `fold_*` family of parsers since they work very similarly to `many_*` and ranges would bring the same benefit to them.

However Rust Ranges are [~~a nightmare~~ slightly prone to footguns](https://ridiculousfish.com/blog/posts/least-favorite-rust-type.html). I see the following points that need to be taken into account:

- [ ] Anything can be a Range.
  In the case of `many_*` it should probably only take a Range of `usize`. It needs to pack the result into an array so `usize` makes the most sense and prevents passing invalid types accidentally.
For `fold_*` it should be limited to any integer type. `fold_*` is not limited by memory since it only has one accumulator value and as such can safely run for longer than `usize` times (this may even be desireable on systems with very small `usize` limits). So perhaps something like an `IntoIntRange` trait? I dont think nom would currently benefit from other range types, such as floats, but more traits could be introduced later on if desired.

- [ ] Open-ended ranges.
Open-ended ranges (`1..`) with `many_*` could exhibit the behavior you described (be equivalent to `1..usize::MAX`). Alternatively they could fail once they would exceed the `usize` limit.
However the `fold_*` parsers could theoretically run for an infinite amount of cycles, even beyond the limit of its integer type. While this might not be a very common case it is still feasible and the current implementation allows such cases. So I would like to preserve that behavior with ranges as well.
Because of these points any implementation of this should also supply the parser with information about the exact bounds of the range.
Perhaps the exact behavior of unbounded ranges is best offloaded onto the parser. Any parser can then handle such special cases itself and then document them individually. That would allow for the most amount of flexibility and prevents conflicting interpretation of range bounds.

[...]

PR #1401, opened 2021-09-16
Add `fold_count` parser equivalent for `count`

nickelc:


cenodis:
This is already possible with `fold_many_m_n`. I dont think adding a completely new parser just for `m == n` is worth it.

src/multi/mod.rs
```diff
@@ -608,6 +608,74 @@ where
...
+    let mut input = i.clone();
+    let mut acc = init();
+
+    for _ in 0..count {
+      let input_ = input.clone();
+      match f.parse(input_) {
+        Ok((i, o)) => {
+          acc = fold(acc, o);
+          input = i;
+        }
```
cenodis, line 665: There needs to be a check inside this loop to see if the child parser actually consumed any input. The child parser not consuming input inside a loop is an error condition. See fold_many_m_n for an example.

nickelc:
> This is already possible with `fold_many_m_n`. I dont think adding a completely new parser just for `m == n` is worth it.

You could say the same about `count`.

cenodis:
> You could say the same about count.

There are some convenience parsers that allow you to avoid using fold_* (such as many_\*). But I think thats a bit different from adding another fold_* parser.

My current problems with this:
- We already have 3 fold_* functions and adding another one for a use case thats already covered by another (and is trivial to use for such cases) is just unnecessary .
- Its name doesnt fit in with the other functions of the fold_* family. If anything it should be called `fold_many_m`.

Stargateur:
At least name it `fold_many_n`

cenodis:
> At least name it `fold_many_n`

`fold_many_m`. The `n` is the second parameter which doesnt exist here.

Stargateur:
> 
> 
> > At least name it `fold_many_n`
> 
> I think you mean `fold_many_m`. The `n` is the second parameter which doesn't exist here.

in math, `n` come first then you use m, plus `n` is the max in `fold_many_m_n`.

cenodis:
> in math, `n` come first then you use m, .

I dont think thats a set rule. n is often used to denote a ***n***umber of something. However in math one also often uses x or even a. It usually comes down to preference and context.
In this case I would prefer to stay in line with the other function names and make it `fold_many_m`. Alternatively we could drop it altogether and just make it `fold_many`, with the possibility to later expand the parameter to also take ranges as proposed in #1393.

> plus `n` is the max in `fold_many_m_n`

Irrelevant. This parser is about the case of `m == n`.

Stargateur:
> Irrelevant. This parser is about the case of `m == n`.

how irrelevant, that stupid, obviously this is for max only so take `n` is totally relevant.

> Alternatively we could drop it altogether and just make it fold_many

yes why not

cenodis:
> how irrelevant, that stupid, obviously this is for max only so take n is totally relevant.

The only obvious thing is that `n` and `m` are the same in this parser. So they can be used interchangeably. Whether you use the minimum or the maximum is irrelevant if they are the same.

Stargateur:
> 
> 
> > how irrelevant, that stupid, obviously this is for max only so take n is totally relevant.
> 
> The only obvious thing is that `n` and `m` are the same in this parser. So they can be used interchangeably. Whether you use the minimum or the maximum is irrelevant if they are the same.

I will repeat so, no that different you don't understand. if you take `m` this would mean only minimum, so `take_many_m` would be read as "take at least m" cause IS the minimum in `fold_many_m_n`, `n` is more logical ( `n` come before `m`), match the max from already exist function and is often used as "the number n".

cenodis:
This case is also handled by #1402 which in my opinion is a much more elegant solution compared to adding yet another specialized parser.

Xiretza:
This should be solved by #1608, which allows `fold(n, parser, init, fold)`.

Issue #1333, opened 2021-06-26
many_m_n can succeed when min > max


carado:
- nom version: 6.2.1
- nom compilation features used: none

The following code:
```rust
fn main() {
  let res: nom::IResult<&str, Vec<char>> =
    nom::multi::many_m_n(4, 2, nom::character::complete::char('a'))("aaa");
  dbg!(res);
}
```
will succeed, consuming two `'a'`:
```
[src/main.rs:4] res = Ok(
    (
        "a",
        [
            'a',
            'a',
        ],
    ),
)
```
While it is unlikely that someone would hardcode a minimum value greater than a maximum value, they can be the result of more complex code, and a call to `many_m_n` in such conditions should probably systematically fail (as the constraint of parsing something at least min but at most max times would be impossible to satisfy).

Have a nice day

Geal:
thanks for the report, it will be fixed in nom 7 with ab42ced

cenodis:
@Geal looking at ab42ced it looks like the `many_m_n` parser just fails instantly with a standard nom error. This makes it look like the parser works normally but didnt match anything. I think a panic would be more appropriate since min > max is nonsensical input and can never create a working parser. The panic should also occur in the function and not inside the closure since the parser creation itself should fail.

Additionally a nom error might also never reach the user since certain parser such as alt() might swallow it. This could lead to unintended behaviour since the parser runs successfully but doesnt produce the desired output. Since no error is reported finding the source of such an issue would be potentially difficult and time consuming.

Geal:
Parsers should never panic, and since parsers can be created dynamically from the output of other parsers, parser creation should not panic either (that could result in easy DoS attacks by messing with the format).
I think that error should be a failure instead, which would ensure it bubbles up through the parser chain and is not blocked by alt

cenodis:
@Geal I wasnt aware parsers should never panic, my apologies.

However, I would like to argue for `Failure` over `Error` in such cases. Especially with regards to @carado's post.

As already stated `Error` can be easily swallowed by a chain of parsers making it difficult to diagnose when it happens unintentionally. Especially when working with dynamic data where it can be easy to miss the fact that `min` might be greater than `max`.

The fact that the parser fails silently is also not really obvious from the outside. After all the function is called `many_m_n` and not `many_m_n_except_if_n_greater_than_m`. This makes the behaviour of this parser more complex than it needs to be and makes following the program flow more difficult.

[...]

PR #1402, opened 2021-09-18
[WIP] Feature multi range

### LINKED ISSUE #1393: Consolidate parser variants using ranges (e.g. `many0`, `many_m_n`)

cenodis:
This PR intends to fix #1393.

### Description
It replaces existing parser variations that only differ by how many times a subparser may run (like the different `many_*` variants) with a single parser that takes a range.

In addition to a range parsers can also take a single `usize` value as parameter which evaluates to a range containing that value only (`value..=value`). This is primarily done for convenience.

Migration remains fairly simple as the new parsers are a full replacement and just require the respective parameters to be rephrased as a single range.

### Reasoning
Using ranges makes the API cleaner by removing unnecessary function duplication. It also allows nom to integrate with more of Rusts language features (specifically the range syntax).

### Example
```rust
many(2..=4,
  tag("Abcd")
)(i)
```
With the old parsers this would be expressed using the `many_m_n` parser:
```rust
many_m_n(2, 4,
  tag("Abcd")
)(i)
```
Special attention should be drawn to the fact that the other variations, such as `many0` are also possible by using open ended ranges (`0..`, `1..`).

### Notes
The current implementation allows individual parsers to take ranges as parameters but requires them to document how they interpret open ended ranges. This was done to maximize flexibility and allow multiple, potentially contradicting, interpretations to coexist within the codebase.

A good example of this in action is the difference between `many` and `fold`:
- `many` (`a..` -> `a..=usize::MAX`)
`many` packages its results into a `Vec`. Therefore going beyond the `usize` limit makes no sense and any open ended range is capped at `usize::MAX`.
- `fold` (`a..`  -> `a..=∞`)
`fold` has static memory requirements since there is only one accumulator. This accumulator can be used indefinitely and as such the amount of iterations can go beyond even the `usize::MAX` limit (the current implementations of `fold_*` also support such cases). Because of this the parser interprets open ended ranges to mean "can run for an infinite number of times".

# TODOs
### Candidates
The following parsers have been proposed for merging.
- [x] `many_*` (many0, many1, many_m_n)
Resolution: Done
- [x] `fold_*` (fold_many0, fold_many1, fold_many_m_n)
Resolution: Done
- [ ] `take_till*` (take_till, take_till1)
- [ ] `take_while*` (take_while, take_while1, take_while_m_n)
- [ ] `take_until*` (take_until, take_until1)
- [ ] `many_till`

### Open questions
- [x] ~~How should the obsolete parsers be handled?
Currently the obsolete parsers (like `many_m_n`) are deprecated using `#[deprecated=""]` with a message pointing the developer to the replacement. (Usually in the form of a simple `Replaced by <new_parser>`). This allows the changes to remain backwards compatible while steering developers towards using the replacement parsers.
Tests for the deprecated parsers are left as is but are annotated with `#[allow(deprecated)]` to suppress the warnings. I would recommend keeping them in for as long as the parsers still exist to make sure that they arent broken accidentally.~~
Resolution: No deprecations for this release. The old parsers and the new range based parsers will exist side-by-side.
- [ ] Resolve open questions of the individual candidates.
See [here](https://github.com/Geal/nom/pull/1402#issuecomment-922387588)

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -38,6 +38,7 @@ use core::ops::{RangeBounds, Bound};
 /// ```
 #[cfg(feature = "alloc")]
 #[cfg_attr(feature = "docsrs", doc(cfg(feature = "alloc")))]
+#[deprecated = "Replaced with `many`"]
 pub fn many0<I, O, E, F>(mut f: F) -> impl FnMut(I) -> IResult<I, Vec<O>, E>
 where
   I: Clone + InputLength,
@@ -93,6 +94,7 @@ where
 /// ```
 #[cfg(feature = "alloc")]
 #[cfg_attr(feature = "docsrs", doc(cfg(feature = "alloc")))]
+#[deprecated = "Replaced with `many`"]
 pub fn many1<I, O, E, F>(mut f: F) -> impl FnMut(I) -> IResult<I, Vec<O>, E>
 where
   I: Clone + InputLength,
@@ -350,6 +352,7 @@ where
 /// ```
 #[cfg(feature = "alloc")]
 #[cfg_attr(feature = "docsrs", doc(cfg(feature = "alloc")))]
+#[deprecated = "Replaced with `many`"]
 pub fn many_m_n<I, O, E, F>(
   min: usize,
   max: usize,
@@ -638,6 +641,7 @@ where
 /// assert_eq!(parser("123123"), Ok(("123123", vec![])));
 /// assert_eq!(parser(""), Ok(("", vec![])));
 /// ```
+#[deprecated = "Replaced with `fold`"]
 pub fn fold_many0<I, O, E, F, G, H, R>(
   mut f: F,
   mut init: H,
@@ -709,6 +713,7 @@ where
 /// assert_eq!(parser("123123"), Err(Err::Error(Error::new("123123", ErrorKind::Many1))));
 /// assert_eq!(parser(""), Err(Err::Error(Error::new("", ErrorKind::Many1))));
 /// ```
+#[deprecated = "Replaced with `fold`"]
 pub fn fold_many1<I, O, E, F, G, H, R>(
   mut f: F,
   mut init: H,
@@ -793,6 +798,7 @@ where
 /// assert_eq!(parser(""), Ok(("", vec![])));
 /// assert_eq!(parser("abcabcabc"), Ok(("abc", vec!["abc", "abc"])));
 /// ```
+#[deprecated = "Replaced with `fold`"]
 pub fn fold_many_m_n<I, O, E, F, G, H, R>(
   min: usize,
   max: usize,
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -8,7 +8,7 @@ use crate::error::ParseError;
 use crate::internal::{Err, IResult, Needed, Parser};
 #[cfg(feature = "alloc")]
 use crate::lib::std::vec::Vec;
-use crate::traits::{InputLength, InputTake, ToUsize, IntoRangeBounds};
+use crate::traits::{InputLength, InputTake, ToUsize, IntoRangeBounds, RangeIterator};
 use core::num::NonZeroUsize;
 use core::ops::{RangeBounds, Bound};
 
@@ -1047,44 +1047,8 @@ where
     
 
     let mut res = crate::lib::std::vec::Vec::with_capacity(start.unwrap_or(0));
-    let mut count: usize = 0;
-    let mut exhausted = false;
     
-    loop {
-      if match (count.saturating_add(1), range.end_bound()) {
-        (count, Bound::Included(end)) => {
-          if count > *end {
-            true
-          } else {
-            if (count == usize::MAX) && !exhausted {
-              if !exhausted {
-                exhausted = true;
-                false
-              } else {
-              true
-              }
-            } else {
-              false
-            }
-          }
-        },
-        (count, Bound::Excluded(end)) => count >= *end,
-        (count, Bound::Unbounded)=> {
-          if count == usize::MAX {
-            if !exhausted {
-              exhausted = true;
-              false
-            } else {
-              true
-            }
-          } else {
-            false
-          }
-        }
-      } {
-        break;
-      }
-      
+    for count in range.saturating_iter() {
       let len = input.input_len();
       match parse.parse(input.clone()) {
         Ok((tail, value)) => {
@@ -1095,7 +1059,6 @@ where
 
           res.push(value);
           input = tail;
-          count = count.saturating_add(1);
         }
         Err(Err::Error(e)) => {
           if !range.contains(&count) {
@@ -1174,32 +1137,8 @@ where
     }
 
     let mut acc = init();
-    let mut count: usize = 0;
-    let mut exhausted = false;
-    loop {
-      if match (count.saturating_add(1), range.end_bound()) {
-        (count, Bound::Included(end)) => {
-          if count > *end {
-            true
-          } else {
-            if (count == usize::MAX) && !exhausted {
-              if !exhausted {
-                exhausted = true;
-                false
-              } else {
-              true
-              }
-            } else {
-              false
-            }
-          }
-        },
-        (count, Bound::Excluded(end)) => count >= *end,
-        _ => false,
-      } {
-        break;
-      }
-      
+    
+    for count in range.unbounded_iter() {
       let len = input.input_len();
       match parse.parse(input.clone()) {
         Ok((tail, value)) => {
@@ -1210,7 +1149,6 @@ where
 
           acc = fold(acc, value);
           input = tail;
-          count = count.saturating_add(1);
         }
         //FInputXMError: handle failure properly
         Err(Err::Error(err)) => {
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -1047,32 +1047,56 @@ where
       _ => None,
     };
     
-
     let mut res = crate::lib::std::vec::Vec::with_capacity(start.unwrap_or(0));
     
-    for count in range.bounded_iter() {
+    let mut parser = |count: usize| -> Option<Result<I, Err<E>>> {
       let len = input.input_len();
-      match parse.parse(input.clone()) {
-        Ok((tail, value)) => {
-          // infinite loop check: the parser must always consume
-          if tail.input_len() == len {
-            return Err(Err::Error(E::from_error_kind(input, ErrorKind::Many)));
-          }
+        match parse.parse(input.clone()) {
+          Ok((tail, value)) => {
+            // infinite loop check: the parser must always consume
+            if tail.input_len() == len {
+              return Some(Err(Err::Error(E::from_error_kind(input.clone(), ErrorKind::Many))));
+            }
 
-          res.push(value);
-          input = tail;
-        }
-        Err(Err::Error(e)) => {
-          if !range.contains(&count) {
-            return Err(Err::Error(E::append(input, ErrorKind::Many, e)));
-          } else {
-            return Ok((input, res));
+            res.push(value);
+            input = tail;
+            None
+          }
+          Err(Err::Error(e)) => {
+            if !range.contains(&count) {
+              return Some(Err(Err::Error(E::append(input.clone(), ErrorKind::Many, e))));
+            } else {
+              return Some(Ok(input.clone()));
+            }
+          }
+          Err(e) => {
+            return Some(Err(e));
           }
         }
-        Err(e) => {
-          return Err(e);
+    };
+    
+    match range.end_bound() {
+      Bound::Unbounded => for count in (1..=core::usize::MAX).map(|x| x - 1) {
+        match parser(count) {
+          Some(Ok(input)) => return Ok((input, res)),
+          Some(Err(e)) => return Err(e),
+          _ => {},
         }
-      }
+      },
+      Bound::Included(x) => for count in (1..=*x).map(|x| x - 1) {
+        match parser(count) {
+          Some(Ok(input)) => return Ok((input, res)),
+          Some(Err(e)) => return Err(e),
+          _ => {},
+        }
+      },
+      Bound::Excluded(x) => for count in (1..*x).map(|x| x - 1) {
+        match parser(count) {
+          Some(Ok(input)) => return Ok((input, res)),
+          Some(Err(e)) => return Err(e),
+          _ => {},
+        }
+      },
     }
 
     Ok((input, res))
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -991,3 +991,151 @@ where
     Ok((input, res))
   }
 }
+
+/// Range many
+/// ```rust
+/// # #[macro_use] extern crate nom;
+/// # use nom::{Err, error::ErrorKind, Needed, IResult};
+/// use nom::multi::many;
+/// use nom::bytes::complete::tag;
+///
+/// fn parser(s: &str) -> IResult<&str, Vec<&str>> {
+///   many(0..=2, tag("abc"))(s)
+/// }
+///
+/// assert_eq!(parser("abcabc"), Ok(("", vec!["abc", "abc"])));
+/// assert_eq!(parser("abc123"), Ok(("123", vec!["abc"])));
+/// assert_eq!(parser("123123"), Ok(("123123", vec![])));
+/// assert_eq!(parser(""), Ok(("", vec![])));
+/// assert_eq!(parser("abcabcabc"), Ok(("abc", vec!["abc", "abc"])));
+/// ```
+#[cfg(feature = "alloc")]
+#[cfg_attr(feature = "docsrs", doc(cfg(feature = "alloc")))]
+pub fn many<I, O, E, F, G>(
+  range: G,
+  mut parse: F,
+) -> impl FnMut(I) -> IResult<I, Vec<O>, E>
+where
+  I: Clone + InputLength,
+  F: Parser<I, O, E>,
+  E: ParseError<I>,
+  G: NomRange,
+{
+  move |mut input: I| {
+    let start = range.lower_bound();
+    let end = match range.upper_bound() {
+      Some(e) => e,
+      None => UpperBound::Inclusive(usize::MAX),
+    };
+    
+    if let Some(start) = start {
+      if !end.contains(start) {
+        return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN)));
+      }
+    }
+
+    let mut res = crate::lib::std::vec::Vec::with_capacity(start.unwrap_or(0));
+    let mut count = 0;
+    
+    loop {
+      count += 1;
+      if !end.contains(count) {
+        break
+      }
+      let len = input.input_len();
+      match parse.parse(input.clone()) {
+        Ok((tail, value)) => {
+          // infinite loop check: the parser must always consume
+          if tail.input_len() == len {
+            return Err(Err::Error(E::from_error_kind(input, ErrorKind::ManyMN)));
+          }
+
+          res.push(value);
+          input = tail;
+        }
+        Err(Err::Error(e)) => {
+          if count < start.unwrap_or(0) {
+            return Err(Err::Error(E::append(input, ErrorKind::ManyMN, e)));
+          } else {
+            return Ok((input, res));
+          }
+        }
+        Err(e) => {
+          return Err(e);
+        }
+      }
+    }
+
+    Ok((input, res))
+  }
+}
+
+///
+pub enum UpperBound {
+  ///
+  Inclusive(usize),
+  ///
+  Exclusive(usize),
+}
+
+impl UpperBound {
+  ///
+  fn contains(&self, number: usize) -> bool {
+    match self {
+      UpperBound::Exclusive(bound) => number < *bound,
+      UpperBound::Inclusive(bound) => number <= *bound,
+    }
+  }
+}
+
+///
+pub trait NomRange {
+  ///
+  fn lower_bound(&self) -> Option<usize>;
+  ///
+  fn upper_bound(&self) -> Option<UpperBound>;
+  ///
+  fn contains(&self, number: usize) -> bool;
+}
+
+impl NomRange for usize {
+  fn lower_bound(&self) -> Option<usize> {Some(*self)}
+  fn upper_bound(&self) -> Option<UpperBound> {Some(UpperBound::Inclusive(*self))}
+  fn contains(&self, number: usize) -> bool {*self == number}
+}
+
+impl NomRange for core::ops::Range<usize> {
+  fn lower_bound(&self) -> Option<usize> {Some(self.start)}
+  fn upper_bound(&self) -> Option<UpperBound> {Some(UpperBound::Exclusive(self.end))}
+  fn contains(&self, number: usize) -> bool {self.start <= number && number < self.end}
+}
+
+impl NomRange for core::ops::RangeFrom<usize> {
+  fn lower_bound(&self) -> Option<usize> {Some(self.start)}
+  fn upper_bound(&self) -> Option<UpperBound> {None}
+  fn contains(&self, number: usize) -> bool {self.start <= number}
+}
+
+impl NomRange for core::ops::RangeTo<usize> {
+  fn lower_bound(&self) -> Option<usize> {None}
+  fn upper_bound(&self) -> Option<UpperBound> {Some(UpperBound::Exclusive(self.end))}
+  fn contains(&self, number: usize) -> bool {number < self.end}
+}
+
+impl NomRange for core::ops::RangeFull {
+  fn lower_bound(&self) -> Option<usize> {None}
+  fn upper_bound(&self) -> Option<UpperBound> {None}
+  fn contains(&self, _number: usize) -> bool {true}
+}
+
+impl NomRange for core::ops::RangeInclusive<usize> {
+  fn lower_bound(&self) -> Option<usize> {Some(*self.start())}
+  fn upper_bound(&self) -> Option<UpperBound> {Some(UpperBound::Inclusive(*self.end()))}
+  fn contains(&self, number: usize) -> bool {*self.start() <= number && number <= *self.end()}
+}
+
+impl NomRange for core::ops::RangeToInclusive<usize> {
+  fn lower_bound(&self) -> Option<usize> {None}
+  fn upper_bound(&self) -> Option<UpperBound> {Some(UpperBound::Inclusive(self.end))}
+  fn contains(&self, number: usize) -> bool {number <= self.end}
+}
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -1162,7 +1162,7 @@ where
 
     let mut acc = init();
     
-    for count in range.saturating_iter() {
+    for count in range.saturating_iter().map(|x| x - 1) {
       let len = input.input_len();
       match parse.parse(input.clone()) {
         Ok((tail, value)) => {
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -1125,7 +1125,7 @@ where
 /// assert_eq!(parser(""), Ok(("", vec![])));
 /// assert_eq!(parser("abcabcabc"), Ok(("abc", vec!["abc", "abc"])));
 /// ```
-pub fn fold_many<I, O, E, F, G, H, J, K, R>(
+pub fn fold<I, O, E, F, G, H, J, K, R>(
   range: J,
   mut parse: F,
   mut init: H,
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -1019,29 +1019,36 @@ where
   I: Clone + InputLength,
   F: Parser<I, O, E>,
   E: ParseError<I>,
-  G: NomRange,
+  G: core::ops::RangeBounds<usize>,
 {
   move |mut input: I| {
-    let start = range.lower_bound();
-    let end = match range.upper_bound() {
-      Some(e) => e,
-      None => UpperBound::Inclusive(usize::MAX), //TODO instead always fail with upper unbounded?
+    use std::ops::Bound;
+    let start = match range.start_bound() {
+      Bound::Included(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN))),
+      Bound::Excluded(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN))),
+      Bound::Included(start) => Some(*start),
+      Bound::Excluded(start) => Some(*start + 1),
+      _ => None,
     };
     
-    if let Some(start) = start {
-      if !end.contains(start) {
-        return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN)));
-      }
+    if let Bound::Unbounded = range.end_bound() {
+      return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN)));
     }
+    
 
     let mut res = crate::lib::std::vec::Vec::with_capacity(start.unwrap_or(0));
     let mut count = 0;
     
     loop {
       count += 1;
-      if !end.contains(count) {
-        break
+      if match range.end_bound() {
+        Bound::Included(end) => count > *end,
+        Bound::Excluded(end) => count >= *end,
+        _ => false,
+      } {
+        break;
       }
+      
       let len = input.input_len();
       match parse.parse(input.clone()) {
         Ok((tail, value)) => {
@@ -1107,15 +1114,13 @@ where
   G: FnMut(R, O) -> R,
   H: FnMut() -> R,
   E: ParseError<I>,
-  J: NomRange,
+  J: core::ops::RangeBounds<usize>,
 {
   move |mut input: I| {
-    match (range.lower_bound(), range.upper_bound()) {
-      (Some(start), Some(end)) => {
-        if !end.contains(start) {
-          return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN)));
-        }
-      },
+    use std::ops::Bound;
+    match range.start_bound() {
+      Bound::Included(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN))),
+      Bound::Excluded(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::ManyMN))),
       _ => {},
     }
 
@@ -1123,11 +1128,14 @@ where
     let mut count: usize = 0;
     loop {
       count = count.saturating_add(1);
-      if let Some(end) = range.upper_bound() {
-        if !end.contains(count) {
-          break;
-        }
+      if match range.end_bound() {
+        Bound::Included(end) => count > *end,
+        Bound::Excluded(end) => count >= *end,
+        _ => false,
+      } {
+        break;
       }
+      
       let len = input.input_len();
       match parse.parse(input.clone()) {
         Ok((tail, value)) => {
@@ -1141,7 +1149,7 @@ where
         }
         //FInputXMError: handle failure properly
         Err(Err::Error(err)) => {
-          if !range.contains(count) {
+          if !range.contains(&count) {
             return Err(Err::Error(E::append(input, ErrorKind::ManyMN, err)));
           } else {
             break;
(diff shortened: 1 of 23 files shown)


src/multi/mod.rs
```diff
@@ -991,3 +998,170 @@ where
...
+  G: IntoRangeBounds<H>,
+  H: RangeBounds<usize>,
+{
+  let range = range.convert();
+  move |mut input: I| {
+    let start = match range.start_bound() {
+      Bound::Included(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::Many))),
+      Bound::Excluded(start) if !range.contains(start) => return Err(Err::Failure(E::from_error_kind(input, ErrorKind::Many))),
+      Bound::Included(start) => Some(*start),
+      Bound::Excluded(start) => Some(*start + 1),
```
Stargateur, line 1044: unclear, why do that, this need an explanation, why use guard here that doesn't help the lecture.

src/multi/mod.rs
```diff
@@ -991,3 +998,170 @@ where
...
+where
+  I: Clone + InputLength,
+  F: Parser<I, O, E>,
+  E: ParseError<I>,
+  G: IntoRangeBounds<H>,
+  H: RangeBounds<usize>,
+{
+  let range = range.convert();
+  move |mut input: I| {
+    let start = match range.start_bound() {
```
Stargateur, line 1040: `start` => `min` no ?

Stargateur (changes requested):
Overall, I don't like much the saturating range. This would considerably slow down most use of fold like `fold(0.., )`, thus I think we should keep specialization variant of many0 many1 fold_many0 and fold_many1. Cause it's more clear what they do, and would allow us to remove this saturation thing. I very concern with that affecting benchmark significantly.

cenodis:
## Current state of candidate implementation.
### fold
Open questions: None
- [x] Implementation
- [x] Tests
- [x] Documentation
### while
Open questions: None
- [x] Implementation
- [x] Tests
- [x] Documentation
### take_till*
Open questions: It is not possible to just remove the number suffix as with `many` or `fold` because the trimmed version `take_till` is already taken. Only `take_till1` exists with suffix. As such, a different naming scheme must be choosen for such cases.
- [ ] Implementation
- [ ] Tests
- [ ] Documentation
### take_while*
Open questions: Same problem as `take_till*`
- [ ] Implementation
- [ ] Tests
- [ ] Documentation
### take_until*
Open questions: Same problem as `take_till*`
- [ ] Implementation
- [ ] Tests
- [ ] Documentation
### many_till
Open questions: In contrast to all other parsers on this list `many_till` does not have any variants. Because of this, should this parser be augmented with ranges at all? If yes, what name should that parser have?
- [ ] Implementation
- [ ] Tests
- [ ] Documentation

src/multi/mod.rs
cenodis: It checks for inverted ranges. Previously this was a `min > max` check. It is still the same just expressed using `contains()` instead to handle edge cases with included and unbounded bounds.

src/multi/mod.rs
cenodis: Unsure what you mean. The `start` variable is a leftover from the previous version. The only reason it remains is because it is used as the initial allocation size of the `Vec`. Maybe it should be renamed to make that more clear?

cenodis:
> This would considerably slow down most use of fold like fold(0.., )

Im not sure if this would have a significant performance impact. The performance of parsers with both lower and upper bounds should remain identical since they also used ranges internally. And I dont think the impact of a relatively simple iterator in the unbounded cases is that severe.
That being said I could remove the deprecation notice on the 0 and 1 variants. Im still not fully sure how to handle the old parsers (which is why its still an open point in the PR).

> we should keep specialization variant of many0 many1 fold_many0 and fold_many1. Cause it's more clear what they do, and would allow us to remove this saturation thing.

I disagree. Ranges are a first class type in Rust (regardless of how unwieldy their structs can be) and they even have their own syntax. By using that syntax for all cases we can cut down on the amount of functions which makes the API easier to understand because now there are fewer functions whose documentation you have to look up.

Im not sure what you mean with "that saturation thing".
If you mean the interpretation of `a..`, thats a necessary consequence of the inherently conflicting nature of unbounded ranges. In a `many` we are allocating within a `Vec` which has a certain size limit we must never go beyond. As such a truly "endless" range is not and can never be valid. In contrast `fold` can saturate in order to be truly "endless" (which doesnt cause panics and is therefore allowed).
Because of these contradicting interpretation the parsers must be able to decide how they need to behave in these cases. And because this is decided for every parser individually the best place to document that behaviour is in the documentation of the parser. With all that, if you think the documentation can be improved Im all ears.

Actually now that I think about it. This doubles as a bug fix. The current implementation of `many0`/`many1` can panic if the amount of elements exceeds `isize::MAX`. And parsers should never panic so that is definetely a bug.
But in order to check for that `many*` needs to keep track of how many elements it has allocated at which point we are back at the same point as our range iterators. So I think (based on gut feeling, not tests) that the performance penalty is negligible (or potentially even non-existent).

src/multi/mod.rs
Stargateur: I know, what I mean that it's not very obvious, also, I still think it's useless check.

Stargateur:
> And I dont think the impact of a relatively simple iterator in the unbounded cases is that severe.

Well, I think not

> I disagree. Ranges are a first class type in Rust (regardless of how unwieldy their structs can be) and they even have their own syntax. By using that syntax for all cases we can cut down on the amount of functions which makes the API easier to understand because now there are fewer functions whose documentation you have to look up.

That could be used to justify use range anywhere cause you know they are first class thing... not a good argument. fold_many0 express better than `fold(0..` period.

Your idea of using Range for fold was my idea in the behind https://github.com/Geal/nom/issues/1333#issuecomment-903159993, but use range imply we should use it naturally, in this PR you add strange check and add a saturate feature to bend range to what you want. That not what we should do, having a hard limit for overflow is the way range are suppose to be use. And that why we must keep a not using range fold to allow to express a not bounded fold. `0..` doesn't mean 0 to infinity, it's mean 0 to what is max according to context. For example vector `my_vec[..]` doesn't mean infinite to infinite, but 0 to len of vec. nom should be 0..max we can.

cenodis:
> That could be used to justify use range anywhere cause you know they are first class thing... not a good argument.

No, but being a first class type is a good argument to justify using it *where it makes sense*.

> use range imply we should use it naturally

And a "natural" range is what exactly? I am unfamiliar with any such concept (except natural numbers) and the docs make no mention of that as far as I know.

> you add strange check

The check is absolutely necessary. Your comment on #1333 seems to imply that you think ranges somehow prevent the max > min condition. That is not the case. `2..0` is a perfectly valid range in Rust (but makes no sense for any parser in nom, so we have to guard against it).

> add a saturate feature to bend range to what you want

I am not "bending" the range to anything. [Saturation arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic) is a perfectly valid concept and the terminology is even used within Rust. All it means is that instead of terminating after reaching the representable maximum it, it continues to return that maximum indefinitely.

To quote the [Rust docs for RangeFrom](https://doc.rust-lang.org/std/ops/struct.RangeFrom.html):
> Overflow in the Iterator implementation (when the contained data type reaches its numerical limit) is allowed to panic, wrap, or saturate.

So saturating is valid even for normal unbounded ranges.
Note: This would actually mean that limiting an unbounded range is invalid. Luckily we are not using the actual `Iterator` implementation and can therefore make up our own behaviour. But thats fine for as long as its properly documented (which it is). This doc snippet is just to illustrate my point.

> having a hard limit for overflow is the way range are suppose to be use.

Untrue, see the doc snippet above. If anything, a "hard limit" for an unbounded range is the only thing it is **not** allowed to do.

> `0..` doesn't mean 0 to infinity, it's mean 0 to what is max according to context.

You are contradicting yourself. Previously you tell me I shouldnt "bend the range to what you want" and I am adding "strange saturation behaviour" but here you tell me there is no globally defined behaviour and it should instead be interpreted based on the context?

To reiterate, interpreting based on the context is **exactly** what we are doing right now:
- `many` the range is used in context with a `Vec`. Therefore it is limited to an upper bound of `usize::MAX` because going past that would be impossible. So an unbounded upper bound is interpreted as "at most `usize::MAX`" in this context.
- `fold` the range is merely used to check for how often the parser has run. It does not have any limiting factors in its context and can therefore go even beyond the normal `usize::MAX` limit. So an unbounded upper bound is interpreted as "truly infinite" in this context.

Stargateur:
You doesn't seem open to discussion at all, you read but don't listen. You don't look like you code in Rust from long. That not how you are suppose to implement or use unbound range. Your solution is not good. I don't have your level of english so everything I say look stupid so I will end talk and just say, I opposite to this PR.

cenodis:
> You doesn't seem open to discussion at all, you read but don't listen.

I am very much open to discussion. This is what discussions are: two people exchanging their (possibly opposing) viewpoints in hopes of coming to an understanding. And I do read and try very hard to listen. But I am having a tough time understanding what exactly your problems are.

> That not how you are suppose to implement or use unbound range.

Then how exactly would you implement this? Be aware that the restriction is that the parser must be able to perform all functions of the parsers that it replaces. If you can give me an alternative to how the ranges are supposed to be interpreted while staying within that restriction I would be happy to hear it. Because I dont really see any other option.

> Your solution is not good.

Ok. But its a solution, it works and it solves a rather annoying issue (the parser duplication). To me thats all a solution really needs.

> I opposite to this PR

Thats nice. Unfortunately I still have absolutely no idea why. So Im afraid I cant really take that vote into account as anything more than "some dude opposes this PR because reasons".

Stargateur:
see what I mean, I said why but you don't even acknowledge this, you are not open to talk, you talk that all.

Stargateur:
> Then how exactly would you implement this? Be aware that the restriction is that the parser must be able to perform all functions of the parsers that it replaces. If you can give me an alternative to how the ranges are supposed to be interpreted while staying within that restriction I would be happy to hear it. Because I dont really see any other option.

use `end_bound` directly and handle `Bound::*` directly do a while loop for Unbounded.

cenodis:
> use end_bound directly to handle Bound::* directly do a while loop for Unbounded.

This was actually done (kind of) in one of the [earlier revisions](https://github.com/cenodis/nom/blob/c8de0b447509bb5d442684a6496d5798934acc1f/src/multi/mod.rs#L1044-L1050). Ultimately I switched it out for the following reasons:
- It bloats the parser with (mostly) duplicate logic.
- The iteration logic distracts from the actual parsing logic.
- Its hilariously easy to mess this up. Even in that earlier revision it didnt work right because of integer overflows and other edge cases. [This](https://github.com/cenodis/nom/blob/feature_multi_range/src/traits.rs#L1482-L1512) is the current logic for one kind of iteration. Putting all of that into every parser would make the above problems even worse.

I see where you are coming from. As that revision shows I had that exact same thought process. But it ultimately didnt pan so it was replaced with something better.
The current implementation neatly seperates the parsing logic from iteration logics and allows users to write clear, concise parsers without having to worry about falling into one of the edge cases of ranges.

Stargateur:
> * It bloats the parser with (mostly) duplicate logic.

factorize better, but yes, this should have duplicate logic, there ARE different behavior. You try to factorize it by hide ugly thing behind a bound wrapper, doesn't make it disappear.

> * The iteration logic distracts from the actual parsing logic.

No sense

> * Its hilariously easy to mess this up. Even in that earlier revision it didnt work right because of integer overflows and other edge cases. [This](https://github.com/cenodis/nom/blob/feature_multi_range/src/traits.rs#L1482-L1512) is the current logic for one kind of iteration. Putting all of that into every parser would make the above problems even worse.

Cause what you do here is very strange what you should do look like:

```rust
match range.end_bound() {
    Unbounded => loop {},
    Included(max) => for 0..=max {}
    Excluded(max) => for 0..max {}
}
```

What is inside these loop can be factorized, and with this we can clearly see this feature of handle range HAVE a cost, I'm fine with a one match, but your solution, is way way much more that one match for your saturating iter.

cenodis:
> ```rust
> match range.end_bound() {
>     Unbounded => loop {},
>     Included(max) => for 0..=max {}
>     Excluded(max) => for 0..max {}
> }
> ```

Now this is something I can actually understand. Sounds interesting.

Much of the parsing logic could be moved into a seperate function defined inside the parser function (so that way it stays grouped together) and then get called from the loop. And specific interpretation of unbounded ranges is done in the match.

Ill try this tomorrow and see how a practical implementation holds up. Maybe that way we dont need to copy so much from stdlib.

cenodis:
I dislike how the parsers work under the current implementation. And I also dislike the performance overhead that matching the bounds in the iterator incurs (as was done in the very first implementation).

So instead I have gone back and redesigned the trait itself. Instead of relying on `RangeBounds` we define our own trait that abstracts the functionality we need and implement that on every concrete type.

Currently the new approach is used in `fold`. `many` still uses the old functionality with a generic `RangeBounds` so it makes for a good comparison.

In my opinion this is kind of a "best of both worlds" approach:
- Simplicity: The parsers only need minimal modification to work with range types and parsers dont need to handle any range-specific functionality (unlike when trying to `match` the upper bounds in the parser itself).
- Performance: Due to monomorphisation and associated types all decisions with regards to the range types is done at compile time. As far as I understand the type system there should be no runtime overhead in that regard.
- Generics: Because the trait is itself implemented on all concrete types we no longer require a second `RangeBounds<usize>` generic on each parser method.

The downside is that the trait implementation is more verbose and doesnt automatically extend to new implementors of `RangeBounds`. But I would rather have the trait be more complicated if it makes the parsers easier to work with. And I doubt there will be more significant implementors of `RangeBounds` (or even less likely, an extension to the current range syntax that we would need to handle).

From my experience with implementing `fold` this approach feels pretty solid and hopefully addresses all major points. This is something I would consider a candidate for a final implementation.

@epage
This new implementation gets rid of the `RangeBounds` dependency altogether and uses concrete types instead. This also makes the idea of having `usize` implement `RangeBounds` no longer useful (I am not saying its not useful in general, just that this implementation will not benefit from it).

cenodis:
It looks like the minimal rust version check enforces that the project must compile on rust version 1.41.1. However the README states that the MSRV is 1.48.0 and compatibility with 1.41.1 is not guaranteed.

@Geal Is it acceptable to raise the version for the minimal rust version check? Or perhaps just remove it since its not a guarantee and therefore doesnt need to be explicitly enforced?

src/multi/mod.rs
```diff
@@ -991,3 +998,194 @@ where
...
+            }
+          }
+          Err(e) => {
+            return Some(Err(e));
+          }
+        }
+    };
+    
+    match range.end_bound() {
+      Bound::Unbounded => for count in (1..=core::usize::MAX).map(|x| x - 1) {
```
Stargateur, line 1079: don't make any sense, just use an infinite loop or a while let loop

src/multi/mod.rs
```diff
@@ -991,3 +998,194 @@ where
...
+        match parser(count) {
+          Some(Ok(input)) => return Ok((input, res)),
+          Some(Err(e)) => return Err(e),
+          _ => {},
+        }
+      },
+      Bound::Included(x) => for count in (1..=*x).map(|x| x - 1) {
```
Stargateur, line 1086: the map don't make anysense

src/multi/mod.rs
```diff
@@ -991,3 +998,194 @@ where
...
+        match parser(count) {
+          Some(Ok(input)) => return Ok((input, res)),
+          Some(Err(e)) => return Err(e),
+          _ => {},
+        }
+      },
+      Bound::Excluded(x) => for count in (1..*x).map(|x| x - 1) {
```
Stargateur, line 1093: same

src/multi/mod.rs
```diff
@@ -991,3 +998,194 @@ where
...
+  J: NomRange<usize>,
+{
+  move |mut input: I| {
+    if range.is_inverted() {
+      return Err(Err::Failure(E::from_error_kind(input, ErrorKind::Fold)));
+    }
+
+    let mut acc = init();
+    
+    for count in range.saturating_iter().map(|x| x - 1) {
```
Stargateur, line 1165: again your `saturating_iter` don't make any sense and again you hide a ugly `SaturatingIter ` with it, and again a map that don't make any sense

src/multi/mod.rs
cenodis: Assume the parser always succeeds. Having an infinite loop would basically break down to this:
```rust
loop {
  vec.push(result)
}
```
This would mean we could try to push more than `usize::MAX` elements onto the Vec. That would cause a panic and is therefore invalid. To prevent this an unbounded range is instead interpreted as having an upper bound of at most `usize::MAX` iterations.

src/multi/mod.rs
cenodis: The alternative would be this:
```rust
!range.contains(&(count-1)) {
```
Either way, the number of successful parses is the number of the current loop (which starts at 1) minus 1. So somewhere we have to subtract 1. I find the map more pleasant (and more robust) than individually subtracting from `count` inside the parser (which might be easy to forget if you use `count` multiple times.

Stargateur: no you add this problem by wanting factoring what should not be factorized

src/multi/mod.rs
Stargateur: first this would panic at `isize::MAX`, secondly the program will ABORT before so what ? and third this is the impossible case cause Rust std choice to not return an result for allocation, use try_reserve api if you want avoid that, in any case, don't use this useless loop.

src/multi/mod.rs
cenodis: What is ugly about the Iterator now? Its definition is literally:
```rust
fn next(&mut self) -> Option<Self::Item> {
    let old_count = self.count;
    self.count = self.count.saturating_add(1);
    Some(old_count)
  }
```
It does not get more concise than that. And I cant just substitute it with a raw loop because we still have to count the number of successful parses because its later used to check if we fulfill the range (we could still be below the minimum and must therefore fail).

The exact type of `saturating_iter()` gets evaluated at compile time so there should be no additional overhead. What other problem could there be with it? The only alternative would be keeping count manually:

```rust
let mut count = 0;
loop {
  //parser stuff here
  count = count.saturating_add(1);
}
if (!range.contains(&count)) {
  //error here
```
Which I honestly think is more ugly and senseless than having a simple type that does it for you.

Stargateur:
> 
> 
> It looks like the minimal rust version check enforces that the project must compile on rust version 1.41.1. However the README states that the MSRV is 1.48.0 and compatibility with 1.41.1 is not guaranteed.
> 
> @Geal Is it acceptable to raise the version for the minimal rust version check? Or perhaps just remove it since its not a guarantee and therefore doesnt need to be explicitly enforced?

geal probably forget to change the MSRV, I think you can ignore that for now, and anyway this PR is breaking change so will be for next major update anyway

src/multi/mod.rs
Stargateur: it's becoming hard to follow since you name it the same that for your `RangeIterator` trait. And anyway, what is the point now ? use https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add if you want to have it wrap nicely, check at every loop if this overflow is very very very not good and don't need this at all for unbound.

cenodis:
> this PR is breaking change

The parsers and traits being introduced are completely new and nothing is removed (only deprecated). So this isnt a breaking change. If you do see something that causes a breakage let me know because that needs to be fixed.

Stargateur:
also on a note, from the compiler, it would probably optimize away the one match of `end_bound` since it's also compile time information.

cenodis:
> also on a note, from the compiler, it would probably optimize away the one match of end_bound since it's also compile time

That may be true. But my real concern is the complexity of having to explicitly `match` the range inside a parser. It makes writing parsers much more unwieldy and more prone to mistakes. And since we only really care about creating iterators to count to our upper limit I much prefer abstracting it.

The abstraction also comes at no cost since the associated types are evaluated at compile time and the iterator functions immediately return a value (I would be seriously surprised if the compiler doesnt inline those).

Having `bounded_iter()` and `saturating_iter()` do the actual work of translating the ranges into the iterator forms we need is much more comfortable from the parser perspective and shouldnt have any performance downsides compared to the explicit match on `end_bound()`. (If anything the trait should be "better" since its guaranteed that associated types are evaluated at compile time, while the `match` is optimized at the compilers discretion).

It really just boils down to the fact that the abstraction is easier to use at no additional cost.

Stargateur:
> That may be true. But my real concern is the complexity of having to explicitly `match` the range inside a parser. It makes writing parsers much more unwieldy and more prone to mistakes. And since we only really care about creating iterators to count to our upper limit I much prefer abstracting it.

So better make mistake in another part of the code. Again you don't listen

Geal:
Hey! I'm sorry I have not had the time to review this yet, I'll try to take a look on Thursday or Friday? In the meantime, could we put a pin on it and cool off the discussion?

cenodis:
> cool off the discussion?

Sure. Apologies for getting so heated. Im really just trying to get some useful criticism out of Star because he seems to have some good ideas but I just cant understand them.

Also this is just a Draft. Having you weigh in on what is acceptable to do and what not is helpful but a merge should wait until this PR is properly open.

Stargateur:
I'm sorry I probably look aggressive but I'm just very concern with this cause it's a critical part of nom that every parser use.

src/traits.rs
```diff
@@ -1395,6 +1397,163 @@ impl HexDisplay for str {
   }
 }
 
+/// Allows conversion of a type into a RangeBound.
+pub trait IntoRangeBounds<T>
+where
+  T: RangeBounds<usize>
+{
+  /// Convert to a RangeBound
+  fn convert(self) -> T;
```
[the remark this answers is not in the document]
cenodis: Outdated. The new trait uses concrete types instead of doing conversions.

src/multi/mod.rs
cenodis: Outdated.

src/multi/mod.rs
cenodis: Outdated. New implementation avoids having the parser map the iterators.

An infinte loop with a `vec.push` is incorrect so a fixed upper bound of `usize::MAX` is necessary.

src/multi/mod.rs
cenodis: Outdated. The need to map the iterators in parsers has been resolved.

src/multi/mod.rs
cenodis: Wrapping arithmetic is wrong and doesnt work.

src/multi/mod.rs
```diff
@@ -37,6 +39,7 @@ use core::num::NonZeroUsize;
 /// ```
 #[cfg(feature = "alloc")]
 #[cfg_attr(feature = "docsrs", doc(cfg(feature = "alloc")))]
+#[deprecated = "Replaced with `many`"]
```
Geal, line 42: I'd like to hold out on the deprecation warnings. `many0` and others are extensively used, I'd rather keep them around in the next major version and ease transition to range based combinators

src/traits.rs
```diff
@@ -1395,6 +1397,199 @@ impl HexDisplay for str {
...
+
+impl NomRange<usize> for RangeInclusive<usize> {
+  type Iter1 = Range<usize>;
+  type Iter2 = Range<usize>;
+  
+  fn bounds(&self) -> (Bound<usize>, Bound<usize>) {(Bound::Included(*self.start()), Bound::Included(*self.end()))}
+  
+  fn contains(&self, item: &usize) -> bool {RangeBounds::contains(self, item)}
+  
+  fn is_inverted(&self) -> bool {<RangeInclusive<usize>>::is_empty(self)}
```
Geal, line 1479: if possible, I'd like to keep nom usable on 1.41.1 for now. As said in the readme, it works on that version without guarantee, but the debian stable upgrade happened quite recently, so people have not migrated quickly to the new version (which has a more recent rustc)

Geal (changes requested):

src/multi/mod.rs
cenodis: I figured deprecating would direct people towards the new parsers while remaining fully backwards compatible. But I can take out the warnings.

Geal:
do we have concrete evidence of the overhead of the saturating iterator? Like, is it actually slower, or is LLVM smart enough to see that it will loop indefinitely?

Honestly I'm not a huge fan of the saturating iterator trick for fold, I prefer the approach in https://github.com/Geal/nom/pull/1402#issuecomment-922402501 even if it results in some code duplication. But it can be a good tradeoff (especially since it gives some info at compile time). BTW the match could be done in the execution of the combinator instead of the parser closure that is returned.

cenodis:
> do we have concrete evidence of the overhead of the saturating iterator?

I have looked at the output of the bench job and compared them to the results on master. Im not sure how reliable those are but the difference seems to drown in the noise. So from that perspective the saturating iterator seems to remain competitive.

> I prefer the approach in #1402 (comment) 

The solution in https://github.com/Geal/nom/pull/1402#issuecomment-922402501 is incomplete. The match would need to be extended with two additional checks to see if the lower bound is 0 or 1. If its not then we still need a saturating counter to compare against the lower bound later. So I think that would need at least 5 checks.

src/multi/mod.rs
cenodis: Done

src/traits.rs
cenodis: Done

cenodis:
So I tried my implementation (or at least a good approximation without needing to import actual nom types) on godbolt. Looking at the assembler output, the compiler actually translates my implementation to the exact same binary as stars example:

```
example::main:
        sub     rsp, 56
        mov     qword ptr [rsp + 8], 42
        lea     rax, [rsp + 8]
        mov     rax, qword ptr [rsp + 8]
        mov     qword ptr [rsp], rax
        cmp     rax, 42
        jne     .LBB9_2
        add     rsp, 56
        ret
.LBB9_2:
        mov     qword ptr [rsp + 8], 0
        mov     rdi, rsp
        lea     rsi, [rsp + 8]
        call    core::panicking::assert_failed
        ud2
```
So to me that looks like the compiler can easily optimize its way around that. At least for constant values, but dynamic ranges need to be handled via the `m_n` variants anyway and those do counting in the old implementation as well, so no loss there.

Here are the two examples for comparison:
[Current implementation](https://godbolt.org/z/57PM749P7)
[Stargateurs example](https://godbolt.org/z/7TW81nc61)

Stargateur:
> The solution in [#1402 (comment)](https://github.com/Geal/nom/pull/1402#issuecomment-922402501) is incomplete. The match would need to be extended with two additional checks to see if the lower bound is 0 or 1. If its not then we still need a saturating counter to compare against the lower bound later. So I think that would need at least 5 arms.

Why you need to check 0 or 1 ?

cenodis:
> Why you need to check 0 or 1 ?

Assume the wanted range is `4..`. That is a range without an upper bound, which would therefore fall into the `Unbounded` match arm. Now tell me how you want to parse that with a simple `loop` and no count variable to keep track of the number of iterations.

Stargateur:
> Edit: In addition the ranges in your comment are wrong. `0..max` and `0..=max` for exclusive and inclusive respectively would mean the loop runs for one loop too many.

What ? no ? That not clear. `fold(0..18)` mean I want run at least 0 time and max 18 time. `fold(0..=18)` I want at least 0 and max 19.

> Assume the wanted range is 4... That is a range without an upper bound, which would therefore fall into the Unbounded match arm. Now tell me how you want to parse that with a simple loop and no count variable to keep track of the number of iterations.

There is nothing wrong to just use a simple mutable counter or you could use:

```rust
let iter = std::iter::successors(Some(0usize), |n| n.checked_add(1));
for i in iter {
}
```

If you really want to handle overflow.

But I think I would do:

```rust
match (range.start_bound(), range.end_bound) {
  (Bound::Included(start), Bound::Included(end)) => for 0..=end {}, // need to check included start
  (Bound::Included(start), Bound::Excluded(end)) => for 0..end {}, // need to check included start same code then ^
  (Bound::Included(start), Bound::Unbounded) => { for 0..=start { } then loop { } } // the for loop is not allow to error
  (Bound::Excluded(start), Bound::Included(end) => for 0..=end {}, // need to check excluded start
  (Bound::Excluded(start), Bound::Excluded(end) => for 0..=end {}, // need to check excluded start same code then ^
  (Bound::Excluded(start), Bound::Unbounded) => { for 0..start { } then loop { } } // the for loop is not allow to error same code then ^^^
  (Bound::Unbounded, Bound::Included(end)) => for 0..=end {}, // no check
  (Bound::Unbounded, Bound::Excluded(end)) => for 0..end {},  // no check
  (Bound::Unbounded, Bound::Unbounded) => loop {},  // no check
```

This will allow compiler to compile check the code, there is no error possible.

cenodis:
> What ? no ? That not clear. fold(0..18) mean I want run at least 0 time and max 18 time. fold(0..=18) I want at least 0 and max 19.

That breaks the way ranges are interpreted in the old parsers. The range was always meant to represent the minimum and maximum amount of iterations the inner loop is allowed to do. So `0..=18` is anything between 0 and 18 iterations. Your implementation is the one that is "unclear".
Even if there wasnt any precedence. `0..18` means "anything between 0 and 18 *exclusively*". So why would you suddenly handle it like it was inclusive? I would suggest reading up on how ranges work.

> If you really want to handle overflow.

That also breaks compatibility with existing parsers. And with checked arithmetic it still needs to check if an overflow has occured. So I dont really see how that is supposed to be much better compared to just saturating, which is also a check for overflow that just returns the max amount instead of `None`. Doesnt seem that different (performance wise).

> This will allow compiler to compile check the code, there is no error possible.

(warning: sarcasm)
Yes, *because there is absolutely no way for anyone to make a single mistake while writing out that monstrosity of a match statement*. The compiler cant check for your logical errors.

My implementation exposes all the necessary type information about each range as well. It just encapsulates it in the respective trait implementation so all the necessary edge case handling doesnt need to happen in the parser.
My experiment with godbolt has shown that the compiler is perfectly able to optimize the traits and iterators away (to the point that it literally produces the exact same output for both versions).

I want to be careful that this doesnt end in another 20+ post argument because of the language barrier.
So unless you give me something with actual practical impact Im going to call it quits here. If you want to you can write your own "ideal" implementation and then we can compare.

Stargateur:
> That breaks the way ranges are interpreted in the old parsers. The range was always meant to represent the minimum and maximum amount of iterations the inner loop is allowed to do. So `0..=18` is anything between 0 and 18 iterations. Your implementation is the one that is "unclear".

That a new parser so I don't care of old one. Your interpretation is misleading with what user expect from range.

> Even if there wasnt any established precedence. `0..18` means "anything between 0 and 18 _exclusively_". So why would you suddenly handle it like it was inclusive? I would suggest reading up on how ranges work.

What ?

> That also breaks compatibility with existing parsers. And with checked arithmetic it still needs to check if an overflow has occured. So I dont really see how that is supposed to be much better compared to just saturating, which is also a check for overflow that just returns the max amount instead of `None`. Doesnt seem that different (performance wise).

choice whatever you want

> My implementation exposes all the necessary type information about each range as well. It just encapsulates it in the respective trait implementation so all the necessary edge case handling doesnt need to happen in the parser.

Your code is unclear, use magic code like where does come from this 4 ? and hide logic far away from the code. Yes my idea is way less error prone in the long run. It's may look bad but it's very rusty.

cenodis:
> What ?

Please read [this](https://doc.rust-lang.org/reference/expressions/range-expr.html) and the documentation for the individual types [here](https://doc.rust-lang.org/std/ops/index.html#structs). Then please understand why what you are proposing is fundamentally wrong not just from a programming perspective but also from a mathematical perspective.

> choice whatever you want
> Your code is unclear
> where does come from this 4
>  hide logic far away from the code
> Yes my idea is way less error prone in the long run.

Im going to assume this is once again the language barrier and/or you just flat out dont understand the problem. As such, this is not a good use of my time and Im going to call it quits.

Your code doesnt work, breaks assumptions about the basic types and breaks compatibility with the old parsers. One of these points alone would be enough to disqualify it.

My code only looks "magic" and unclear if you dont understand how traits and their implementations work.
Its not "hiding" anything. The proper term is "organizing" which is pretty important if you actually want to maintain your codebase for any reasonable amount of time. Anyone that has a basic understanding of rust would be able to tell how the compiler resolves this (and how that resolution has no real performance penalty since it happens at compile time).

And no, your idea is many things, but definitely not "less error prone". My implementation has a single point of failure (the trait implementations). Your implementation has a point of failure for every parser that is implemented that way (or possibly even for every single match arm if you want to be exact).

In addition to all of the above your implementation has yet another downside: dynamic data. The compiler is not guaranteed to infer anything about dynamic data and may (at worst) include the entire match statement in the final output. With my implementation the compiler will at the very least eliminate conditional checks since the type is known at compile time and guaranteed to be resolved to a concrete implementation.

As for the "rustyness" of the code. I would say taking full advantage of rusts type and trait system makes my implementation fairly "rusty". Because apparently thats a solid, objective metric now.

This is the last post I will make about this until you prove to me that you actually understand what Im trying to tell you as I am fairly convinced that this is just going to end in another dispute that solves nothing and just wastes time that I dont have.

Stargateur:
> Please read [this](https://doc.rust-lang.org/reference/expressions/range-expr.html) and the documentation for the individual types [here](https://doc.rust-lang.org/std/ops/index.html#structs). Then please understand why what you are proposing is fundamentally wrong not just from a programming perspective but also from a mathematical perspective.

```rust
use std::ops::{Bound, RangeBounds};

struct RangeFromInclusive {
    pub start: usize,
}

impl RangeBounds<usize> for RangeFromInclusive {
    fn start_bound(&self) -> Bound<&usize> {
        Bound::Included(&self.start)
    }
    fn end_bound(&self) -> Bound<&usize> {
        Bound::Unbounded
    }
}

fn main() {
    let range = RangeFromInclusive { start: 42 };

    assert_eq!(range.start_bound(), Bound::Included(&42));
}
```

That not because core syntax didn't include this that user can't do it.

You should really take into account that I'm a rust expert. I probably still have to learn but I pretty good in Rust. I code in Rust every day in my work, and have know Rust pre 1.0. My advice are not random, what I see is you probably a very good C++ dev or whatever OOP language, but sometime in Rust one should not use trait. My match reveal we handle 5 different behavior, there is no way to factorize this without complicated code that would really an heavy optimization. We should prefer KISS.

Stargateur:
my example is stupid cause I would have done for excluded, anyway, same point

Stargateur:
if you don't want use range like range are mean to be use, maybe just take two parameters `min: Option<usize>, max: Option<usize>`

Stargateur:
> So unless you give me something with actual practical impact Im going to call it quits here. If you want to you can write your own "ideal" implementation and then we can compare.

I will do that I think I code better than I write, but I would have prefer we work together.

cenodis:
> there is no way to factorize the 5 behaviors without complicated code

Im not sure what your standards for "complicated" are, but at least for me the following is pretty simple:
```rust
fn bounded_iter(&self) -> impl Iterator<usize> {
  0..core::usize::MAX
}
```
All implementations are like this, they just return a single value which gets used as an iterator.
The only complexity arises from the fact that each range type needs to be handled slightly differently. I did not abstract this because I thought of any OOP pattern. I abstracted it because why instantiate the ranges manually in the parser when we can have the compiler do it?
Ultimately this is no different than the ranges found in the individual `match` arms in your example. Except we save ourselves the work of having to write it out in every individual parser. Instead we let the compiler do the matching.

> would need an heavy optimization that will be hard to track down.

Again, not sure what "hard to track down" is for you. But its nothing different from a normal trait implementation which are used everywhere in Rust.

Monomorphisation replaces the generic function call with one of the concrete implementations.
For example, with an inclusive range `0..=12` this:
```rust
for count in range.bounded_iter() {
```
gets substituted by the compiler using the trait implementation for `RangeInclusive<usize>` to this:
```rust
for count in 0..12 {
```
which the exact same thing as the harcoded `m_n` variant, except the compiler can substitue for other range types as well. This of course gets further optimized the same way in either case.

There is really nothing hard to track down in this chain. The functions are annotated with the trait generic and from there its trivial to find the respective implementation for whatever concrete type you pass to it.

And if you look at the comparison between our two godbolt examples above you will see the compiler applies the same optimizations to the point that the binary output is literally identical. So we are not really loosing anything because of this abstraction.

The only real argument for performance is the use of a saturating counter in the case of `0..` and `1..` (all other ranges with an unbounded upper bound require some kind of counter). But going of the bench tests on github any difference (if it exists) drowns in the noise. So unless proven otherwise Im going to say it has no practical impact.

> We should prefer KISS

This is KISS. At least not anymore complicated than having to deal with a dozen ranges and loops in a single function. I dont think looking up a trait implementation counts as "complicated".

Xiretza:
Hi, just checking in to ask about the current status here. From what I can see there has been a lot of discussion with few actual results - are there any specific open questions other than the naming of the `take_*` parsers at this time?

cenodis:
> Hi, just checking in to ask about the current status here. From what I can see there has been a lot of discussion with few actual results

The status can be summed up as "a working implementation exists". Not more, not less.
> are there any specific open questions other than the naming of the `take_*` parsers at this time?

Geal has also raised the question of wether the saturating counter used in the trait implementations has any overhead over the explicit match with 3 different loops. Based on the benchmark tests I couldnt see any significant difference but I havent stress tested it in great detail or inspected the dissasembly.

I personally havent really worked on this recently as I have a second PR open here that was supposed to be reviewed ~6 months ago. I assume Geal is busy but I cant really finish either until I get some proper feedback from a maintainer.

Geal:
This is now merged as part of #1608 :partying_face: 
Sorry it took so long. I might revisit some points later, but for now it's better to integrate it with the rest and iterate. Thanks a lot for implementing this complex feature!

Issue #1408, opened 2021-09-27
Reduce syntactic noise


epage:
## Prerequisites

Here are a few things you should provide to help me understand the issue:

- Rust version : 1.55.0
- nom version : 7.0.0
- nom compilation features used: default

## Test case

I wanted to experiment with porting `toml_edit` to nom.  Papercuts I ran into
- Deep module nesting makes it slow to walk around docs
- Deep module nesting makes it annoying to fully qualify functions
- `use nom::...::*;` led to conflicts between nom, my parserr, and `std` (especially `char`0
- After a foray with `combine`, a lot of the syntactic noise of nom stands out (though I still prefer nom)

An example parser
```rust
use nom::{
    branch::*, bytes::complete::*, character::complete::*, combinator::*, error::context, multi::*,
    sequence::*, AsChar, IResult,
};

...

pub(crate) fn dec_int(input: &str) -> IResult<&str, &str> {
    recognize(tuple((
        opt(alt((char('+'), char('-')))),
        alt((
            char('0'),
            map(
                tuple((
                    satisfy(|c| ('1'..='9').contains(&c)),
                    take_while(is_dec_digit_with_sep),
                )),
                |t| t.0,
            ),
        )),
    )))(input)
}
```
Ideal:
```rust
pub(crate) fn dec_int(input: &str) -> nom::IResult<&str, &str> {
    ((
        nom::opt(nom::alt(('+', '-'))),
        nom::alt((
            '0',
            ((
                nom::satisfy(|c| ('1'..='9').contains(&c)),
                nom::take_while(is_dec_digit_with_sep),
            )).map(|t| t.0)
        ))
    )).recognize()(input)
```

Unsure how much is even possible or what downsides there might be, but ideas include
- Flatten modules, where possible (work like https://github.com/Geal/nom/issues/1393 can help keep it navigatable)
- ~~`impl Parser` for `u8`, `&[u8]`, `char`, and `&str`, reducing the need for `char` and `tag`~~
  - ~~Maybe even implement parsers for ranges of `u8` and `char`~~
  - Divide of `streaming` vs `complete` prevents this
- `impl Parser` for tuples, reducing the need for `tuple(())`, https://github.com/Geal/nom/issues/1417
- Emphasize `Parser::map` in "choose your combinator" and docs (I didn't know it existed until I started writing this), https://github.com/Geal/nom/issues/1415
- Add additional inherent functions to `Parser`, like
  - `Parser::between(first, third)`
[...]

PR #1410, opened 2021-09-27
Provide shortcut for `streaming::char`

epage:
The hope is that intent is clear enough while reducing syntactic noise.

The main ambiguity is `streaming` vs `complete`.  I assume `complete` is
more common but `streaming` is the more universal one, so I went with
that.

### DIFF
--- a/src/internal.rs
+++ b/src/internal.rs
@@ -326,6 +326,34 @@ where
   }
 }
 
+impl<'a, I, E> Parser<I, char, E> for char
+where
+  I: crate::traits::Slice<crate::lib::std::ops::RangeFrom<usize>>
+    + crate::traits::InputIter
+    + crate::traits::InputLength,
+  <I as crate::traits::InputIter>::Item: crate::traits::AsChar,
+  E: crate::error::ParseError<I>,
+{
+  /// Recognizes one character.
+  ///
+  /// *Streaming version*: Will return `Err(nom::Err::Incomplete(_))` if there's not enough input data.
+  /// # Example
+  ///
+  /// ```
+  /// # use nom::{Err, error::{ErrorKind, Error}, Needed, IResult};
+  /// # use nom::Parser;
+  /// fn parser(i: &str) -> IResult<&str, char> {
+  ///     'a'.parse(i)
+  /// }
+  /// assert_eq!(parser("abc"), Ok(("bc", 'a')));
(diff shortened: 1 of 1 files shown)


src/internal.rs
```diff
@@ -326,6 +326,34 @@ where
...
+  /// # use nom::Parser;
+  /// fn parser(i: &str) -> IResult<&str, char> {
+  ///     'a'.parse(i)
+  /// }
+  /// assert_eq!(parser("abc"), Ok(("bc", 'a')));
+  /// assert_eq!(parser("bc"), Err(Err::Error(Error::new("bc", ErrorKind::Char))));
+  /// assert_eq!(parser(""), Err(Err::Incomplete(Needed::new(1))));
+  /// ```
+  fn parse(&mut self, i: I) -> IResult<I, char, E> {
+    crate::character::streaming::char(*self)(i)
```
epage, line 353: @Stargateur I figured `streaming` was a safe default  Know of any reason to use `complete` or concerns with having to pick one over the other?

src/internal.rs
```diff
@@ -326,6 +326,34 @@ where
   }
 }
 
+impl<'a, I, E> Parser<I, char, E> for char
+where
+  I: crate::traits::Slice<crate::lib::std::ops::RangeFrom<usize>>
+    + crate::traits::InputIter
+    + crate::traits::InputLength,
+  <I as crate::traits::InputIter>::Item: crate::traits::AsChar,
+  E: crate::error::ParseError<I>,
```
epage, line 335: @Stargateur 

It looks like `internal` is a self-contained core of the crate, not depending on anything else.

- Is it ok to use stuff from outside `internal`?
- If not, should this `impl` live somewhere else, like next to `char`?

Once we've decided on how this should look, I'll clean this up and get it implemented for the other types discussed in #1408

Stargateur: I can't speak for geal but I think it's ok but I would reverse, move code from `character::char` to here and call this implementation in `character::char`. Trait are ok but maybe parser is too much. Anyway, this file should probably be call `parser` anyway.

> If not, should this impl live somewhere else, like next to char?

Since this impl Parser for a foreign type I think the correct place is here close to trait definition.

src/internal.rs
Stargateur: Oh this could be a big problem...

PR #1412, opened 2021-09-28
Clarify the `internal` module name

epage:
As part of #1410, the name `internal.rs` was discussed and it was
brought up that `parser.rs` would provide clearer intent.


epage:
@Stargateur what would be the path for moving this and https://github.com/Geal/nom/pull/1413 forward so I can work on the issues I've created?

Stargateur:
> 
> 
> @Stargateur what would be the path for moving this and #1413 forward so I can work on the issues I've created?

I have no control on this repo, gael have not much time so... wait until gael have time ? Thus I considering ask gael to be a maintainer of nom.

epage:
Ok, didn't realize he was the only one with merge permission.

epage:
> parser is a better name than internal, yes. This will have to wait for the 8.0 nom though, I'm opening the milestone for it

How come this has to wait for nom 8?  The `internal` mod isn't exposed, so this shouldn't be a breaking change.

Stargateur:
> 
> 
> > parser is a better name than internal, yes. This will have to wait for the 8.0 nom though, I'm opening the milestone for it
> 
> How come this has to wait for nom 8? The `internal` mod isn't exposed, so this shouldn't be a breaking change.

As I said, more clarity, having duplicate path is not helping. Better do it in one time. For example, std have this problem https://doc.rust-lang.org/std/?search=usize.

src/lib.rs
```diff
@@ -437,24 +437,24 @@ pub mod lib {
...
 pub use self::str::*;
 
 #[macro_use]
 pub mod error;
 
-mod internal;
-mod traits;
 pub mod combinator;
+mod parser;
+mod traits;
```
Xiretza, line 450: Can you split these rustfmt changes (reordering of `mod {traits,sequence,bytes}`) to a separate commit? Otherwise git blame becomes unnecessarily noisy.

Xiretza (changes requested):
Apart from the rustfmt noise, this LGTM, and I agree that this is **not** an externally visible change and can thus be merged immediately.

src/lib.rs
epage: Done.

Xiretza (approved):

PR #1423, opened 2021-10-07
Fix broken "take_while" link

jihchi:
The "take_while" link is broken (you'd see `The requested resource does not exist`).


Geal:
thanks!

Issue #1417, opened 2021-09-28
impl Parser for tuples


epage:
## Prerequisites

Here are a few things you should provide to help me understand the issue:

- Rust version : 1.55.0
- nom version : 7.0.0
- nom compilation features used: default

## Test case

We call the sequence function `tuple`, its a small step from there to go ahead an impl `Parser` for tuples.

I took this idea from [combine](https://docs.rs/combine/4.6.1/combine/trait.Parser.html) and found porting combine code to nom, this added a lot of extra noise, obscuring the actual grammar among all the nom syntax.

*This is forked from #1408*

An example parser
```rust
use nom::{
    branch, bytes::complete as bytes, character::complete as character, combinator, multi,
    sequence, AsChar, IResult as Outcome,
};

pub(crate) fn dec_int(input: &str) -> IResult<&str, &str> {
    combinator::recognize(sequence::tuple((
        combinator::opt(multi::alt((character::char('+'), character::char('-')))),
        multi::alt((
            character::char('0'),
            combinator::map(
                sequence::tuple((
                    character::satisfy(|c| ('1'..='9').contains(&c)),
                    multi::take_while(is_dec_digit_with_sep),
                )),
                |t| t.0,
            ),
        )),
    )))(input)
}
```
Ideal
```rust
use nom::{
    branch, bytes::complete as bytes, character::complete as character, combinator, multi,
    sequence, AsChar, IResult as Outcome,
};

pub(crate) fn dec_int(input: &str) -> IResult<&str, &str> {
    combinator::recognize((
        combinator::opt(multi::alt((character::char('+'), character::char('-')))),
        multi::alt((
            character::char('0'),
            combinator::map(
                (
                    character::satisfy(|c| ('1'..='9').contains(&c)),
                    multi::take_while(is_dec_digit_with_sep),
                ),
                |t| t.0,
            ),
        )),
    ))(input)
}
```

Geal:
good idea!

Geal:
fixed in #1425

PR #1425, opened 2021-10-18
Implement Parser for tuples

kompass:
Hello ! My first PR to this project.

This PR does :
- Implement the issue #1417.
- Add some test.
- Modify the `fn tuple` to use this new feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,7 @@ extern crate nom;
 use nom::{
   IResult,
   bytes::complete::{tag, take_while_m_n},
-  combinator::map_res,
-  sequence::tuple
+  combinator::{map_res, parse},
 };
 
 #[derive(Debug,PartialEq)]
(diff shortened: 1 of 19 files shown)


kompass:
I think the `tuple` function is still useful in case the tuple in question is not composed in another parser. Indeed, tuples implement `Parser`, not` FnMut`, so it is impossible to parse in the same way as other parsers. This code doesn't compile : 
```rust
let (i, result) = (parser1, parser2)(i);
````

To maintain consistency in calls, I think it could be a good idea to keep it. The trait `Tuple` is now only used to be able to implement this function, I don't see any other way.

**Comment by Stargateur** [comment]
I don't follow you, you want use it as documentation ? Unless I miss something remove `Tuple` trait the code still compile.

**Comment by kompass** [comment]
The `Tuple` trait is still used to implement the `tuple` function.

We can still do :
```rust
let (i, result) = tuple((parser1, parser2)(i)?;
```

Geal:
sorry for the delay, it is now merged! I removed the `parse` combinator as it is not really necessary, but the rest was great, thanks for your help!

PR #1440, opened 2021-10-24
Added two allows so clippy would pass a very complicated math formula for benchmarking.

chotchki:
Clippy doesn't like the contrived arithmetic example for benchmarking. Suppressed the errors.


Geal:
thanks!

PR #1441, opened 2021-10-24
Updated the pull request from #469 so that it will cleanly apply against master.

chotchki:
Cleaned up the conflicts from #469 so that the parser can be merged cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@
 
 ### Changed
 
+## 7.1.0 - 2021-11-04
+
+### Thanks
+
+- @nickelc
+- @Stargateur
+- @NilsIrl
+- @clonejo
+- @Strytyp
+- @schubart
+- @jihchi
+- @nipunn1313
+- @Gungy2
+- @Drumato
+- @Alexhuszagh
+- @Aehmlo
+- @homersimpsons
+- @dne
+- @epage
+- @saiintbrisson
+- @pymongo
+
+### Changed
+
+- documentation fixes
+- Ci fixes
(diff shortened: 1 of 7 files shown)


chotchki:
@Stargateur I like the approach working on it.

chotchki:
@Stargateur This is pushing my understanding of nom but I'm trying to understand the signature fold_until0 completely.

I have the following signature with the where clause:
```
pub fn fold_until0<F, G, H, A, Input, Output, Error>(
  f: F,
  g: G,
  init: H,
  acc: A,
) -> impl FnMut(Input) -> IResult<Input, (A, Output), Error>
where
  Input: InputTake + InputIter + InputLength + Clone,
  F: Parser<Input, Output, Error>,
  G: FnMut(A, Output) -> (A, Output),
  H: FnMut() -> (A, Output),
  Error: ParseError<Input>,
```

- "f" is the child parser that will recognize the end
- "g": I don't understand in the context of the other arguments
- "init": Function for the initial state for the accumulator
- "acc": Function for building the accumulator up. I don't know if this really should be "g"
- I'm returning the remaining input, the accumulation, the content until "f" matched (dropping "f"'s output).

Can you please help me understand if BOTH "g" and "acc" make sense? I'm reading https://docs.rs/nom/7.0.0/nom/multi/fn.fold_many0.html and it seems to just use g to accumulate.

Stargateur:
```rust
pub fn fold_until0<P, Until, Init, Acc, Fold, Input, Output, UntilOutput, Error>(
    parser: P,
    until: Until,
    init: Init,
    fold: Fold,
) -> impl FnMut(Input) -> IResult<Input, (Acc, UntilOutput), Error>
where
    Input: InputTake + InputIter + InputLength + Clone,
    F: Parser<Input, Output, Error>,
    Until: Parser<Input, UntilOutput, Error>,
    Fold: FnMut(Acc, Output) -> Acc,
    Init: FnMut() -> Acc,
    Error: ParseError<Input>;
```

- `parser` is used to parse what is before could be `anychar` for quick and dirty parsing
- `until` is the parser used to stop the output of this parser is returned by fold_until
- `init` is the init fct used to get the accumulator of the user
- `fold` is the user fct that use the accumulator and return it

the body of `fold_until0` is very simulator to https://github.com/Geal/nom/pull/1341

Issue #1288, opened 2021-02-20
`separated_list1` fails if separator is empty


reivilibre:
The [`separated_list1`](https://docs.rs/nom/6.1.2/nom/multi/fn.separated_list1.html) combinator fails if the separator is empty.

I suppose this is because of this excerpt here:

https://github.com/Geal/nom/blob/0b92df971a7d2090023b353e0283f9b2db6f2a84/src/multi/mod.rs#L353-L355

This behaviour is not obvious; perhaps it is unusual to desire it, though.

My use case was to have space-separated values, but allow the space to be omitted if a `(` or `[` bracket followed immediately, as in

```rust
    let space_or_peek_bracket = alt((
        char(' '),
        peek(one_of("(["))
    ));
    let (i, terms) = separated_list1(space_or_peek_bracket, parse_term)(i)?;
```

I notice that `separated_list1` was previously extended to allow empty elements.
I also notice that `many1` forbids empty elements (because of the potential for an infinite loop).

I'm sure I will be able to work around this without too much trouble, but perhaps `separated_list1` could be made to allow this, or have documentation added that makes it obvious?

(I suspect the documentation route may be the best, because otherwise you could have both empty elements and empty separators and this would mean that you could have an infinite loop, similar to `many1`…)

In any case, thought it was worth dropping a note here in case anyone else scratches their head.

nom version : `6.1.0`

reivilibre:
For anybody that comes in the future, this was how I resolved this case for me:

```rust
    let mut terms = Vec::new();

    let (i, term) = parse_term(i)?;
    terms.push(term);

    let space_or_peek_bracket = alt((
        char(' '),
        peek(one_of("(["))
    ));

    let (i, extra_terms) = many0(pair(space_or_peek_bracket, parse_term))(i)?;
    for (_, term) in extra_terms.into_iter() {
        terms.push(term);
    }
```

Lucretiel:
The [looping combinators](https://docs.rs/nom-supreme/0.4.5/nom_supreme/multi/index.html) in [nom-supreme](https://docs.rs/nom-supreme/) also address this, by using a smarter 0-length match checker (0-length matches allowed, but terminates if an infinite loop is detected)

PR #1444, opened 2021-10-25
Allow empty sep for separated_list but not sep and f empty

Stargateur:
Fix https://github.com/Geal/nom/issues/1288

This make more sense to consider both should be empty to be in infinite loop, and follow expected behavior of:

```rust
let (input, output) = separated_list0(opt(char(',')), parser)(input);
```

equivalent to manually do:

```rust
let (input, output) = parser(input)?;
let acc = vec![output];
let (input, output) = fold_many0(
  preceded(opt(char(',')), parser),
  || acc,
  |mut acc, output| { acc.push(output); acc },
)(input)
```

where `fold_many0` fail only if the both seq and item parser are empty.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -229,20 +229,18 @@ where
     }
 
     loop {
-      let len = i.input_len();
       match sep.parse(i.clone()) {
         Err(Err::Error(_)) => return Ok((i, res)),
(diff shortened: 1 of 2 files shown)


dalcde:
Why was this PR closed?

Issue #1450, opened 2021-10-29
take_split return is not intuitive


Stargateur:
[`take_split()`](https://docs.rs/nom/7.0.0/nom/trait.InputTake.html#tymethod.take_split) doesn't precise what is returned and implementation do:

```rust
fn take_split(&self, count: usize) -> (Self, Self) {
  let (prefix, suffix) = self.split_at(count);
  (suffix, prefix)
}
```

That really not intuitive when you are expected std nomenclature and it's more logical to return it as std do it. Plus, I think this function should return either a Result or a Option (like std should) to avoid unexpected panic.

Open https://github.com/rust-lang/rust/issues/90410.

Geal:
I don't think there's an issue with `take_split`, it's consistent with the rest of nom. Could you avoid opening an issue then immediately opening a PR for an architecture change or new feature without waiting for my opinion? This was not necessary

PR #1453, opened 2021-10-30
Introduce a InputSplit trait

Stargateur:
Closes https://github.com/Geal/nom/issues/1450

I choice to introduce a new trait to allow a transition and avoid mistake. As we can see this produce a much better code for `anychar()` removing one step (maybe save CPU cycles)

This also allow to introduce a nice `take_one()` combinator by updating https://github.com/Geal/nom/pull/1364/ (we could also think about remove `anychar` and have a generic `take_one()` combinator)

Maybe change to just introduce a `SplitFirst` trait

### DIFF
--- a/src/character/complete.rs
+++ b/src/character/complete.rs
@@ -11,7 +11,7 @@ use crate::lib::std::ops::{Range, RangeFrom, RangeTo};
 use crate::traits::{
   AsChar, FindToken, InputIter, InputLength, InputTake, InputTakeAtPosition, Slice,
 };
-use crate::traits::{Compare, CompareResult};
+use crate::traits::{Compare, CompareResult, InputSplit};
 
 /// Recognizes one character.
 ///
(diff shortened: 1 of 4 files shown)


Stargateur:
Maybe do:

```rust
/// Abstracts split_first
pub trait InputSplit: Sized {
  /// The current input type is a sequence of that `Item` type.
  ///
  /// Example: `u8` for `&[u8]` or `char` for `&str`
  type Item;

  /// Divides one input into two at an index.
  fn split_at(self, mid: usize) -> Result<(Self, Self), (Needed, Self)>;

  /// Returns the first and all the rest of the elements of the slice, or None if it is empty.
  fn split_first(self) -> Result<(Self::Item, Self), Self>;

  /// Returns the last and all the rest of the elements of the slice, or None if it is empty.
  fn split_last(self) -> Result<(Self::Item, Self), Self>;
}
```

Allowing consuming `self`.

Issue #1421, opened 2021-10-05
Parsing of long floating point numbers causes panic


MatejKastak:
Hello, first of all thank you, nom is a great tool!

I have tried to migrate to a new nom version (7.0.0) and encountered some unexpected behavior.
- parsing of floating point numbers (`float`, `double`) panics for long inputs
- sadly I found no way to handle this error, program is aborted
- are there any tips to avoid such panics?
- recent `6.*` versions does not panic
- migrating to the most recent version is not critical to me, but I wanted to ask/note this behavior for others

After a little digging, I think this is caused by switching to `minimal-lexical`.
> the lexical-core crate used for float parsing has now been replaced with minimal-lexical: the new crate is faster to compile, faster to parse, and has no dependencies

Somewhat related to #1384 

## Prerequisites

- Rust version : `rustc 1.55.0 (c8dfcfe04 2021-09-06)` also tried it with nightly
- nom version : `v7.0.0`
- nom compilation features used: default

## Test case

```rust
mod tests {
    use nom::number::complete::float;
    use nom::IResult;

    fn parser(s: &str) -> IResult<&str, f32> {
        float(s)
    }

    #[test]
    fn float_panic() {
        // 7.0.0 -> panic
        // 6.2.1 -> pass
        assert!(std::matches!(parser("0.00000000000000000087"), Ok(_)));
        assert!(std::matches!(
            parser("0.00000000000000000000000000000000000000000000000000000000000000087"),
            Ok(_)
        ));

    }
}               
```

nom `v7.0.0`
```
---- tests::float_panic stdout ----
thread 'tests::float_panic' panicked at 'attempt to shift left with overflow', /home/anon/.cargo/registry/src/github.com-1ecc6299db9ec823/minimal-lexical-0.1.4/src/lemire.rs:155:5
```

nom `v6.2.1`
```
running 1 test
test tests::float_panic ... ok
```

parasyte:
This should have been closed by #1455

Geal:
Yes. Leaving it open for now, I'd like to get a proper fix in minimal-lexical, because we take a perf hit with that workaround

PR #1455, opened 2021-11-02
introduce a workround for float parsing

### LINKED ISSUE #1384: Updating to Nom 7 has broken `double("NaN")`

Geal:
this will afford some time until we get a proper fix with minimal lexical for floats that are to large

not cleaned up yet

Fixes #1384 and #1421

### DIFF
--- a/src/number/streaming.rs
+++ b/src/number/streaming.rs
@@ -1395,6 +1395,40 @@ where
   )(input)
 }
 
+// workaround until issues with minimal-lexical are fixed
+#[doc(hidden)]
+pub fn recognize_float_or_exceptions<T, E: ParseError<T>>(input: T) -> IResult<T, T, E>
+where
+  T: Slice<RangeFrom<usize>> + Slice<RangeTo<usize>>,
+  T: Clone + Offset,
+  T: InputIter + InputTake + InputLength + Compare<&'static str>,
+  <T as InputIter>::Item: AsChar,
+  T: InputTakeAtPosition,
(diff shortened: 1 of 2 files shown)


src/number/streaming.rs
```diff
@@ -2045,6 +2100,14 @@ mod tests {
...
       Err(Err::Incomplete(Needed::new(1)))
     );
+
+    let (_i, nan) = float::<_, ()>("NaN").unwrap();
+    assert!(nan.is_nan());
+
+    let (_i, inf) = float::<_, ()>("inf").unwrap();
+    assert!(inf.is_infinite());
+    let (_i, inf) = float::<_, ()>("infinite").unwrap();
+    assert!(inf.is_infinite());
```
Stargateur, line 2110: I think this test should fail cause it's only eat 'inf' and not 'infinite' https://github.com/Geal/nom/pull/1391/files# parse infinity before inf cause of this

PR #1462, opened 2021-11-13
length_count: use Vec::with_capacity

tanriol:
This change avoids reallocation while filling in the array. In my case this causes a ~20% perf increase.

This might cause regressions if some users rely on nom *not* allocating the whole array before an error happens, causing their parsers to change behavior from "starts parsing, encounters an error, backtracks to a previous `alt`/`opt`/whatever" to "starts parsing, causes an OOM abort". I don't think nom provides any guarantees in either direction.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -958,7 +958,7 @@ where
   move |i: I| {
     let (i, count) = f.parse(i)?;
     let mut input = i.clone();
-    let mut res = Vec::new();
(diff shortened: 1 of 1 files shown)


Geal:
I will close this now, length_count will use `with_capacity` in nom 8, with an additional check for an initial max capacity to make sure the allocation is reasonable:

https://github.com/rust-bakery/nom/blame/main/src/multi/mod.rs#L1546-L1551

Issue #1282, opened 2021-02-12
Implement FindToken<char> for &[char]?


MggMuggins:
Would you accept a PR to implement `FindToken<char> for &[char]`? It's not a super common use case but it would be convenient.

```rust
use nom::character::complete::one_of;

const SUBSCRIPTS: [char; 10] =
    ['₀', '₁', '₂', '₃', '₄', '₅', '₆', '₇', '₈', '₉'];

fn main() {
  assert_eq!(one_of(&SUBSCRIPTS[..])("₈rest"), Ok(("rest", "₈")));
}
```
```
error[E0277]: the trait bound `&[char]: FindToken<_>` is not satisfied
   --> src/main.rs:7:21
    |
7   |   assert_eq!(one_of(&SUBSCRIPTS[..])("₈rest"), Ok(("rest", '₈')));
    |                     ^^^^^^^^^^^^^^^ the trait `FindToken<_>` is not implemented for `&[char]`
    |
   ::: ~/.cargo/registry/src/github.com-1ecc6299db9ec823/nom-6.1.0/src/character/complete.rs:89:6
    |
89  |   T: FindToken<<I as InputIter>::Item>,
    |      --------------------------------- required by this bound in `nom::character::complete::one_of`
    |
    = help: the following implementations were found:
              <&'a [u8] as FindToken<char>>
              <&'a [u8] as FindToken<u8>>
              <&'b [u8] as FindToken<&'a u8>>
              <[u8; 0] as FindToken<&'a u8>>
            and 65 others
```

ctrlcctrlv:
Solved in #1463.

PR #1463, opened 2021-11-14
`impl<'a> FindToken<char> for &'a [char]`

ctrlcctrlv:
Closes Geal/nom#1282. Test case added to `tests/`. Useful for me.


Geal:
thanks!

PR #1469, opened 2021-12-04
Doc example for take

ThomasdenH:
This PR adds an example to `take` to show how the results will be different when using `char` and `u8`. This is a property of the `InputIter` trait rather than this method, but I think it is still useful. I have not changed the other `take_...` methods to keep the PR small, but also because they all take a closure where the type is specified in the input argument, which makes it harder to use incorrectly.


Geal:
that's a great example, it was not obvious!

PR #1470, opened 2021-12-05
Add example for integer parsing to digit1

ThomasdenH:
This PR adds an example on how to parse an integer to `digit1`.


Geal:
thanks! that will help a lot

PR #1475, opened 2021-12-16
Add Meilisearch to the list of project using nom

irevoire:
Since this PR in milli meilisearch/milli#400 and its integration in [Meilisearch v0.25.0](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.25.0rc0) we use nom to parse our mini [filters](https://docs.meilisearch.com/reference/features/filtering_and_faceted_search.html#filtering-and-faceted-search) language.

Not sure we can call that a “programming language” but it's the best category it think :thinking:


Geal:
that's an awesome example, its great that you took the time to fuzz it!

PR #1480, opened 2021-12-20
Add docs of number limitation of tuple

doehyunbaek:
Hi, this is a small documentation PR documenting number limit of tuple combinator. I just copied description from alt's docs and applied a little modification.

https://github.com/Geal/nom/blob/dfa55914dc9673e3092974e00f8967220fbfaf78/src/branch/mod.rs#L44-L46

### DIFF
--- a/src/sequence/mod.rs
+++ b/src/sequence/mod.rs
@@ -253,8 +253,7 @@ tuple_trait!(FnA A, FnB B, FnC C, FnD D, FnE E, FnF F, FnG G, FnH H, FnI I, FnJ 
   FnM M, FnN N, FnO O, FnP P, FnQ Q, FnR R, FnS S, FnT T, FnU U);
 
 ///Applies a tuple of parsers one by one and returns their results as a tuple.
-///There is a maximum of 21 parsers. If you need more, it is possible to nest them in other `tuple` calls,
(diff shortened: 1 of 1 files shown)


Xiretza:
This doesn't work as well as for `alt()`, since it actually creates nested tuples (whereas nested `alt()`s are equivalent to one big `alt()`). Also, is it ever really necessary to parse more than a 21-tuple? It would indicate some severe code smell to me...

src/sequence/mod.rs
```diff
@@ -253,7 +253,8 @@ tuple_trait!(FnA A, FnB B, FnC C, FnD D, FnE E, FnF F, FnG G, FnH H, FnI I, FnJ
   FnM M, FnN N, FnO O, FnP P, FnQ Q, FnR R, FnS S, FnT T, FnU U);
 
 ///Applies a tuple of parsers one by one and returns their results as a tuple.
-///
+///There is a maximum of 21 parsers. If you need more, it is possible to nest them in other `tuple` calls,
+/// like this: `tuple(parser_a, tuple(parser_b, parser_c))`
```
Geal, line 257:
```suggestion
///There is a maximum of 21 parsers
```

PR #1483, opened 2021-12-29
fix typo in docs

punkeel:
nom no longer relies on macros, but an exclamation mark was left over


Xiretza (approved):
Typo fix, LGTM. There aren't any other mentions of macro parsers at least in the top-level crate docs.

Geal:
Thanks!

PR #1484, opened 2022-01-03
Small typo

max-sixty:
Very small!


Geal:
Small indeed, but every little bit helps!

PR #1485, opened 2022-01-04
Remove reexport of bits::* in crate root

Xiretza:
This reexport seems to be a remnant from the days of macro parsers, which no longer exist.

It has caused confusion on several occasions. `nom::complete::tag`, for example, refers to `nom::bits::complete::tag`, while users actually want `nom::bytes::complete::tag` in the vast majority of cases.

This is (obviously) a breaking change and thus needs to wait for v8.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Thanks
 
+### Removed
+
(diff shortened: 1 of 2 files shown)


Geal:
I agree, this will go in 8.0. Thank you for pointing it out though

Xiretza:
I tried adding a `#[deprecated]` marker to the reexport to warn users about this early, but that doesn't actually work - see https://github.com/rust-lang/rust/issues/30827.

Geal:
thanks! That sounds reasonable to remove them from the global exports yes

PR #1488, opened 2022-01-14
MSRV clarifications

Xiretza:
I struggled a bit to find out what nom's MSRV is, since the keyword isn't mentioned in the README (I didn't notice the badge, banner blindness...), so I added it to the relevant heading.

I also added the `rust-version` field to Cargo.toml to allow this check to be automated.


Geal:
thanks!

PR #1490, opened 2022-01-21
Wrap the input parsers to the `tuple` sequence combinator in a tuple

5c077m4n:
A minor fix to the [choosing a combinator](https://github.com/Geal/nom/blob/main/doc/choosing_a_combinator.md#sequence-combinators) docs page - the `tuple` combinator accepts only one input (a tuple of parsers) according to [docs.rs](https://docs.rs/nom/latest/nom/sequence/fn.tuple.html)


Xiretza (approved):
LGTM, although this may be unnecessary if #1425 gets accepted.

Geal:
thanks!

PR #1491, opened 2022-01-22
fix: support empty sep for separated_list0 & separated_list1

suikammd:
When separator eats nothing(e.g. `opt(tag(","))`), `separated_list0` & `separated_list1` should parse without error. But current implementation returns error when separator eats nothing to prevent infinite loop.
I think the infinite loop check should between loops. If no bytes eaten during the whole loop body, then this will result in infinite loop and should return error.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -234,17 +234,16 @@ where
         Err(Err::Error(_)) => return Ok((i, res)),
         Err(e) => return Err(e),
         Ok((i1, _)) => {
-          // infinite loop check: the parser must always consume
-          if i1.input_len() == len {
-            return Err(Err::Error(E::from_error_kind(i1, ErrorKind::SeparatedList)));
-          }
-
           match f.parse(i1.clone()) {
             Err(Err::Error(_)) => return Ok((i, res)),
             Err(e) => return Err(e),
             Ok((i2, o)) => {
               res.push(o);
               i = i2;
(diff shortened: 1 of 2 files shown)


SaltyKitkat:
Seems this will fix #1573 and #1691

epage:
If I understand this fix, it changes it from require progress from the `sep` to requiring progress from either `sep` or `f`.  Personally, I don't think this is a good idea.  While this would allow the caller to choose whether `sep` or `f` can be empty, it can also allow both to be empty *sometimes* and requires either careful reading of the documentation (rarely done) and auditing of the code or exhaustive testing to catch that an infinite loop exists.

A lot of these higher level parsers are relatively trivial (well, barring the 8.0 GAT changes) and people can easily write a parser tailored to their needs.  For myself, I only use parsers like this for prototyping but when I'm lining up with a grammar, I use lower level parsers so my code lines up with my BNF grammar.

gdennie:
Just spent a while "discovering" this issue. The documentation does not stipulate the constraint that the separator **must** consume something and that an optionally consuming separator such as `space0` is therefore invalid. The documentation more general approach is ideal and intuitive and so I vote for this fix.

Geal:
sorry, I did not see this PR when writing https://github.com/rust-bakery/nom/commit/c6559557a8066fd4fe0a65447791a254c4be2b91, which was roughly equivalent. I will close this PR now

PR #1496, opened 2022-01-29
add usize and isize to character specific parsers

weichweich:
I would like to add parsers for isize and usize to the character parsers.


Geal:
merged in another PR https://github.com/rust-bakery/nom/commit/9745930d5da957cdd5622b4d5d6e1027792e83ff
thanks for the proposal anyway!

PR #1497, opened 2022-02-05
Add a binary parser example in README page

erihsu:
Hi, Geal
I have released a new binary database parser([GDSII](https://en.wikipedia.org/wiki/GDSII)) based on nom on [crates.io](https://crates.io/crates/gds2_io). GDSII is an industrial binary format that is widely adopted in layout tool(semi-manual or full automatic). The code is in [gds-io](https://github.com/erihsu/gds2-io) repo

The gds-io crate support parse/save gds2 file and is already tested pass by self loopback

I have written several parsers based on nom but it's my first time want to share it because it's my first binary file parser.


Geal:
this is really cool, thanks!

PR #1504, opened 2022-03-01
Clarify the return value of recognize_float_parts

TheNeikos:
Heya! I saw that `recognize_float_parts` returns several values as its output but without reading the source code I was unable to make out which is which. This PR hopefully clarifies it for the next person to come along!


Geal:
it's great to point it out, thanks!

PR #1507, opened 2022-03-09
Increase permutation trait tuples

Luni-4:
In https://github.com/rust-av/matroska/pull/75 the number of tuples required for a `track_entry` is equal to 37, so it is necessary to increase the number of tuples implemented for the `Permutation` trait. The numbers of tuples have been increased up till 41 to give further room.

Thanks in advance for your review!


Geal:
oh my. That is a LOT of tuple elements :sweat_smile: 
That will go in nom 8

sdroege:
@Geal Do you have a suggestion how this could be done in a nicer way?

Luni-4:
We reimplemented the `Permutation` trait in the matroska crate to cover further scenarios, so this PR is no longer necessary. We can close it, feel free to reopen if someone else is interested in increasing the permutation trait tuples

PR #1518, opened 2022-03-30
feat(readme): add Wordle result parser to list

Fyko:
This PR adds another example to the "Parsers created with Nom" section.


Geal:
great work :)

Issue #1506, opened 2022-03-08
Challenges with using nom


Firstyear:
Hey mate,

Nom is a really powerful library and when it works well, it can do amazing things. But sometimes it can be extremely hard to find which of the many pieces of nom you need to use, and how to combine them. Even I have sat staring at the docs for hours, feeling ultimately very stupid, wondering at what point did my life choices go so wrong. 

Anyway, generally when I get to this point, I always ask "what was the interaction failure that led to this feeling". Here in nom, I think there is a combination of things, but the docs are a really large part. 

For example, I was wondering "how do I assert there are no bytes left in an input". So, since this is "bytes" and I have all the input, I went to:

https://docs.rs/nom/latest/nom/bytes/complete/index.html

But there is nothing here that does what I needed. Searching for "complete", "end", "size", "limit", did not find any results. 

At this point I gave up and did something else, and when working on a different part of the code with a seperate nom parser I found: https://docs.rs/nom/latest/nom/combinator/fn.eof.html 

So what was the "failure" here? 

The primary issue is that it was hard to find and discover what it is that I needed. Often it's extremely difficult to locate what piece of nom you need, because the library being fragmented over so many modules, means that you effectively have ~9 different modules you need to look through. This makes it very difficult to locate combinators, and then effectively identify which one you may wish to use. 

A possible solution is to "flatten" all the combinators to a single module, so that they can all be listed on a single page with the first line of their docs. This also has the benefit that every combinator needs to then be uniquely named, which helps to potentially resolve conflicts and confusion such as complete and streaming bytes. Alternatelly, this "flattened" module, could be made from pub-use of the other modules with aliases. Another benefit to this could be aliases to some combinators to help them be discovered easier. 

Another barrier is that the docs of any parser can be extremely terse. Here's flat map for example:

https://docs.rs/nom/latest/nom/combinator/fn.flat_map.html

It took me multiple attempts to read this to understand *what* this combinator did, and how or why I would use it. The type signature is *huge* and has 8 generics. Even reading that it's pretty hard to see what was going on. The doc is a single line, that doesn't really explain the use case. And the example has no comments to indicate "what" that combinator is doing.

In otherwords, the current docs has basically a single line of (terse) human text to explain the type signature, and a single example that you are expected to reverse engineer and interpret. Especially for myself who failed mathematics and formal types and all that jazz, this makes it extremely hard for me to understand "how" or "why" I might care about flat_map!

I think a constructive improvement to this is that any combinators docs should have:

[...]

PR #1525, opened 2022-05-02
A Book for Nom

tfpk:
Hi,

I'm a newbie to Nom (and I'm probably only an intermediate Rust programmer), and I have personally found that this project lacks documentation for newcomers. I thought that it would be valuable to have a "book" for Nom, in the vein of the Rust Book or the many other excellent books that projects around the Rust ecosystem have.

Based on my limited experience, I've written about a quarter of a short book. This is a book I think would have helped me, had I had access to it a few weeks ago. 

I'm posting it here for a few reasons:

1. Whether it's useful enough to continue working on, and whether the Nom project would be interested in accepting it.
2. If I've put in any significant (or even minor) mistakes that could be corrected.
3. If the order and plan for content is reasonable.

I'm happy to move this to an issue if that's the right place to discuss, though I suppose a useful issue to reference is #1506 .

I'd welcome any and all feedback!

Thanks :)


doc/nom-guide/src/chapter_1.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+3. Lastly, Nom parsers are normally anchored to the beginning of their input.
+   In other words, if you converted a Nom parser to regex, it would generally
+   begin with `/^/`. This is sensible, because it means that nom parsers must
+   (conceptually) be sequential -- your parser isn't going to jump
+   ahead and start parsing the middle of the line.
+
+
+To represent this model of the world, nom uses the `IResult<(I, O)>` type.
```
Xiretza, line 47: This should be `IResult<I, O>`.

Xiretza, line 47: This should be `IResult<I, O>`.

doc/nom-guide/src/chapter_2.md
```diff
@@ -0,0 +1,105 @@
...
+has no special characters, it just matches a string.
+
+Imagine, for example, the regex `/abc/`. It simply matches when the string
+`"abc"` occurs.
+
+In `nom`, we call a simple collection of bytes a tag. Because
+these are so common, there already exists a function called `tag()`.
+This function returns a parser for a given string.
+
+<div class="example-wrap" style="display:inline-block"><pre class="compile_fail" style="white-space:normal;font:inherit;">
```
Xiretza, line 13: Not sure what's up with this HTML, it breaks the markdown in the contents.

doc/nom-guide/src/chapter_2.md
```diff
@@ -0,0 +1,105 @@
...
+
+    let err_input = "defWorld";
+    match parse_input(err_input) {
+        Ok((leftover_input, output)) => unreachable!(),
+        Err(_) => assert!(true),
+    }
+```
+
+If you'd like to, you can also check case insensitive `/tag/i`
+with the `tag_case_insensitive`.
```
Xiretza, line 62: [`tag_no_case`](https://docs.rs/nom/latest/nom/bytes/complete/fn.tag_no_case.html#)

doc/nom-guide/src/chapter_3.md
```diff
@@ -0,0 +1,124 @@
...
+    }
+    
+
+    let input = "def";
+    let parsed_input = parse_numbers_or_abc(input);
+    match parsed_input {
+        Ok(_) => unreachable!(),
+        Err(_) => assert!(true)
+    }
+```
```
Xiretza, line 113: Example seems to be mismatched (does not contain `tuple()` at all)?

Xiretza (commented):
This is great, thanks a lot! I found a few nits, but overall this seems like a great start for a tutorial-like introduction to nom!

tfpk:
Thank you for the feedback! I'll fix those nits up.

Did you have any broader feedback on the structure I'm planning in SUMMARY.md? And if I kept working on this is it something you'd be interested in integrating?

I'm new enough to the project I don't know if there are things I should be explaining!

Xiretza:
The overall structure looks great for a comprehensive tutorial, which is something that nom kind of lacks right now (there is some tutorial-style documentation in `docs/`, but it's not as coherent as an mdbook).

> if I kept working on this is it something you'd be interested in integrating?

Well, you'll have to ask @Geal about that, I'm not a maintainer, but personally I'd gladly accept it ;)
 
> I'm new enough to the project I don't know if there are things I should be explaining!

I hang out in (#geal_nom:gitter.im)[https://matrix.to/#/#Geal_nom:gitter.im], and while it's hard to find a *most common* issue, error handling is definitely something a lot of users struggle with. There are just a lot of levels to the current `IResult`, and it takes a while to dig through them to understand it.

doc/nom-guide/book.toml
```diff
@@ -0,0 +1,6 @@
+[book]
+authors = ["Tom Kunc"]
+language = "en"
+multilingual = false
+src = "src"
+title = "The Nom Guide (Nominomicon)"
```
Geal, line 6: that name is awesome :smile:

doc/nom-guide/src/chapter_1.md
```diff
@@ -0,0 +1,82 @@
...
+A regex, in a sense, controls its whole input. Given a single input,
+it decides that either some text **did** match the regex, or it **didn't**.
+
+```text
+             ┌────────┐           ┌─► Some text that matched the regex
+ my input───►│my regex├──►either──┤
+             └────────┘           └─► None
+```
+
+As we mentioned above, Nom parsers are designed to be combined.
```
Geal, line 15: where is it previously mentioned?

doc/nom-guide/src/chapter_1.md
```diff
@@ -0,0 +1,82 @@
+# Chapter 1: The Nom Way
+
+First of all, we need to understand the way that regexes and nom think about
```
Geal, line 3: I am not sure starting by comparing with regexes helps in understanding. We should be able to explain parser combinators independently (ie, parsers, so takes text or bytes as input and returns structured data and combinators, taking smaller parsers and making larger ones), then extend to a comparison with other approaches likes regular expressions (powerful, but limited to regular languages).
Otherwise we are already assuming that the reader knows enough about regular expressions, which might not be the case.

doc/nom-guide/src/chapter_2.md
```diff
@@ -0,0 +1,105 @@
...
+
+Imagine, for example, the regex `/abc/`. It simply matches when the string
+`"abc"` occurs.
+
+In `nom`, we call a simple collection of bytes a tag. Because
+these are so common, there already exists a function called `tag()`.
+This function returns a parser for a given string.
+
+<div class="example-wrap" style="display:inline-block"><pre class="compile_fail" style="white-space:normal;font:inherit;">
+ **Warning**: `nom` has multiple different definitions of `tag`, make sure you use this one for the
```
Geal, line 14: that will need a link to the future chapter about parsers of complete VS streaming inputs

doc/nom-guide/src/chapter_2.md
```diff
@@ -0,0 +1,105 @@
...
+        Ok((leftover_input, output)) => {
+            assert_eq!(leftover_input, "World");
+            assert_eq!(output, "abc");
+        },
+        Err(_) => unreachable!()
+    }
+
+    let err_input = "defWorld";
+    match parse_input(err_input) {
+        Ok((leftover_input, output)) => unreachable!(),
```
Geal, line 56: if all those examples are only looking at one branch, they should use `unwrap` and `unwrap_err` instead of a `match` with one branch using `unreachable!()`, that will make them shorter

doc/nom-guide/src/chapter_3.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+## Extra Nom Tools
+
+After using `alt()` and `tuple()`, you might also be interested in the `permutation()` parser, which
```
Geal, line 118: `permutation` may be a bit advanced for an introduction. Since we're looking at sequences of parsers, it would be more interesting to talk about `preceded` or `terminated` (applying 2 parsers, returning the result of one of them) or the `many*` variants.

doc/nom-guide/src/introduction.md
```diff
@@ -0,0 +1,31 @@
...
+
+Nom is a parser-combinator library. In other words, it gives you tools to define:
+ - "parsers" (a function that takes an input, and gives back an output), and
+ - "combinators" (functions that take parsers, and _combine_ them together!).
+
+By combining parsers with combinators, you can build complex parsers up from
+simpler ones. These complex parsers are enough to understand HTML, mkv or Python!
+
+Before we set off, it's important to list some caveats:
+ - This guide is for Nom7. Nom has undergone significant changes, so if
```
Geal, line 20: adding that warning is a great idea :)

Geal (changes requested):
I think this is a great idea, a comprehensive suite of small examples from the ground up will really help in understanding nom!

I am not sure about requiring knowledge of regular expressions, parser combinators should be able to stand by themselves. But keeping the regex equivalent with the examples can help some readers.

I'd be happy to merge this after a few fixes. I'm not sure I'll have the time to write a lot in it, but I'll make sure to help and review

tfpk:
@Xiretza  + @Geal  -- I've updated this PR with the next section of the book.

Particularly curious if there's anything major that needs changing in the order or in topics covered.

I'm also blocked on the last three chapters:
 - Re Streaming -- I couldn't find an easy to grok example of how they are used. Is there an up-to-date one? My thinking is that the best "simple" example would be reading in from stdin; and doing something like bracket-matching; but I don't know if that plays to nom's strengths.
 - Re bytes + bits -- again, I don't know the area well enough to have a good example; and I don't know what's worth covering.
 -  Re Exercises, is this worth having? I come from a teaching background, so having exercises is the norm for me, but I'm not sure if it's the done thing in Rust projects. I could also go through and add one or two per chapter?

One other thought -- some of the other documentation in `docs/` would fit well in the book (I haven't written an errors chapter, because the errors document that exists is great!) Is it worth moving some/all of the documentation into the book? I've borrowed particularly heavily from the list of parsers/combinators -- I'm not sure if duplicating them was the best idea, so maybe references to them either by link; or as an appendix might be helpful?

Keen to hear your thoughts/feedback!

doc/nom-guide/src/chapter_2.md
tfpk: I've chosen to go with returning a `Box<dyn Error>`, so that I can use Try. I've done this because (from my limited understanding) the Rust guide to writing documentation suggests avoiding `unwrap` in documentation examples. 

Happy to change this if you think it's the wrong direction.

doc/nom-guide/src/chapter_2.md
tfpk: Agreed; I'll add it in when that chapter is written.

doc/nom-guide/src/chapter_1.md
tfpk: Agreed, and modified :)

doc/nom-guide/book.toml
tfpk: Thanks :D

PR #1526, opened 2022-05-12
Implement `nom::sequence::Tuple` for `()` (Unit)

LoganDark:
Right now, `()` (Unit) does not implement the [`nom::sequence::Tuple` trait](https://docs.rs/nom/7.1.1/nom/sequence/trait.Tuple.html) required to use the [`tuple` combinator](https://docs.rs/nom/7.1.1/nom/sequence/fn.tuple.html). This makes it difficult to use in macros.

## Example

I'm currently mass-producing parsers using a macro for a fictional instruction set assembler. Each instruction may take a variable number of arguments so I am using this macro component to parse all arguments:

```rust
		tuple(($(
			preceded(space1, $parser)),*
		))
```

This works fine for any number of arguments `>=1` and `<=21`, but notably does not work for a number `==0`. Here's why: 

```
the trait bound `(): nom::sequence::Tuple<_, _, _>` is not satisfied
```

Looks like `()` does not implement `Tuple`, even though it could concievably implement a tuple of zero elements.

This pull request implements `Tuple` for `()` such that it will always succeed and it is always a no-op. It also adds a test case that verifies that `tuple(())` never fails and always returns the input unchanged.

I am unsure if the `ParseError<I>` bound is needed on the impl, but I figured it's better to be safe than sorry and stay consistent with the existing tuple impls. I would also appreciate assistance with removing the ugly turbofishes from the test.

### DIFF
--- a/src/sequence/mod.rs
+++ b/src/sequence/mod.rs
@@ -252,6 +252,15 @@ macro_rules! tuple_trait_inner(
 tuple_trait!(FnA A, FnB B, FnC C, FnD D, FnE E, FnF F, FnG G, FnH H, FnI I, FnJ J, FnK K, FnL L,
   FnM M, FnN N, FnO O, FnP P, FnQ Q, FnR R, FnS S, FnT T, FnU U);
 
+// Special case: implement `Tuple` for `()`, the unit type.
+// This can come up in macros which accept a variable number of arguments.
+// Literally, `()` is an empty tuple, so it should simply parse nothing.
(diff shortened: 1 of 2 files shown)


Geal:
good idea, thanks

LoganDark:
It looks like the way you merged the PR clobbered my GPG signature, so it's showing as unverified in the commit log. Unfortunate :/ although my original commit from this PR is signed, so it's not a huge deal.

Geal:
Right, I generally use the "squash and merge" to make one commit per PR and link it to the PR in the title

PR #1529, opened 2022-05-26
Remove duplicated section from error_management.md

wolthom:
In the introduction of the error_management.md file, the section explaining the three different error types is duplicated (with minimal changes between the two sections). This (small) PR removes the redundancy.

### DIFF
--- a/doc/error_management.md
+++ b/doc/error_management.md
@@ -22,10 +22,10 @@ pub enum Err<E> {
 The result is either an `Ok((I, O))` containing the remaining input and the
 parsed value, or an `Err(nom::Err<E>)` with `E` the error type.
 `nom::Err<E>` is an enum because combinators can have different behaviours
-depending on the value:
+depending on the value. The `Err<E>` enum expresses 3 conditions for a parser error:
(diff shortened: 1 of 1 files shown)


doc/error_management.md
```diff
@@ -22,10 +22,10 @@ pub enum Err<E> {
 The result is either an `Ok((I, O))` containing the remaining input and the
 parsed value, or an `Err(nom::Err<E>)` with `E` the error type.
 `nom::Err<E>` is an enum because combinators can have different behaviours
-depending on the value:
+depending on the value. The `Err<E>` enum expresses 3 conditions for a parser error:
+- `Incomplete` indicates that a parser did not have enough data to decide. This can be returned by parsers found in `streaming` submodules to indicate that we should buffer more data from a file or socket. Parsers in the `complete` submodules assume that they have the entire input data, so if it was not sufficient, they will instead return a `Err::Error`
```
Geal, line 26:
```suggestion
- `Incomplete` indicates that a parser did not have enough data to decide. This can be returned by parsers found in `streaming` submodules to indicate that we should buffer more data from a file or socket. Parsers in the `complete` submodules assume that they have the entire input data, so if it was not sufficient, they will instead return a `Err::Error`. When a parser returns `Incomplete`, we should accumulate more data in the buffer (example: reading from a socket) and call the parser again
```

Geal (approved):

Geal:
Thanks!

PR #1541, opened 2022-08-15
docs: add ignore empty lines example

guissalustiano:
Hey! I'm working with name on a personal project and I love it. The [documentation](https://github.com/Geal/nom/tree/main/doc) really helped me and I would like to collaborate.

I write a small function to ignore comments in my parser, and I think it might be useful for other projects and show some combinators (I already have a question [on stack overflow](https://stackoverflow.com/questions/55009665/how-do-i-make-a-nom-whitespace-parser-that-also-skips-line-oriented-comments), but in the old version).

I still need to explain the example but I wanted an ok about the code and the example first.


guissalustiano:
I write [some test example](https://github.com/guissalustiano/nom-docs-example/blob/main/src/lib.rs) too

Issue #1021, opened 2019-08-17
parsing number with an "E" fails if it's not followed by another number


mockersf:
- Rust version : rustc 1.37.0 (eae3437df 2019-08-13)
- nom version : 5.0.0
- nom compilation features used: either default, or with default feature disabled

for context, I'm trying to parse a number with a unit that can be "EB" for exabyte.

## Test case

with default features:
```
fn main() {
    use nom::number::complete::float;

    let parser = |s| float::<_, ()>(s);
    assert_eq!(parser("1B"), Ok(("B", 1.0)));
    assert_eq!(parser("1E"), Ok(("E", 1.0)));

    // expected
    assert_eq!(parser("1EB"), Ok(("EB", 1.0)));
    // actually
    assert_eq!(parser("1EB"), Ok(("B", 1.0)));
}
```

with default features disabled:
```
fn main() {
    use nom::number::complete::float;
    use nom::Err;

    let parser = |s| float::<_, ()>(s);
    assert_eq!(parser("1B"), Ok(("B", 1.0)));

    // expected
    assert_eq!(parser("1E"), Ok(("E", 1.0)));
    // actually
    assert_eq!(parser("1E"), Err(Err::Failure(())));

    // expected
    assert_eq!(parser("1EB"), Ok(("EB", 1.0)));
    // actually
    assert_eq!(parser("1EB"), Err(Err::Failure(())));
}
```

Geal:
a `E` following a number can be part of the syntax for a number, it's the exponential notation, example here: '1E2'.

If you do not need floating point numbers you could combine the `digit` parser and the `parse_to` combinator.

mockersf:
Sadly I do need floating point numbers...

I tried following the code, seems hard to fix, either going to crate `lexical-core` (which seems not easy to understand) or to [`recognize_float`](https://github.com/Geal/nom/blob/master/src/number/complete.rs#L749) (which would be made more complex)

I had something working with nom 4:
```rust
named!(
    parse_float<types::CompleteStr, f64>,
    complete!(flat_map!(recognize_float, parse_to!(f64)))
);
```

AndresParraSilva:
I'm kind of a beginner so I don't dare to create a PR, but the current version also recognize as numbers "." and "E", so I managed to cook this merging `recognize_float` (removing a `cut`) and `double`:

```
fn parse_float<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str, f64, E> {
[...]

PR #1547, opened 2022-09-05
float: Reject invalid exponent instead of failing

TobTobXX:
Previously, recognize_float() (and by extension various other
float-related functions) returned Err::Failure when encountering an
invalid exponent.

This commit changes the behaviour to rejecting and not consuming the
invalid exponent, but parsing the float before just fine.

Fixes: #1021

### DIFF
--- a/src/number/complete.rs
+++ b/src/number/complete.rs
@@ -1414,7 +1414,7 @@ where
       opt(tuple((
         alt((char('e'), char('E'))),
         opt(alt((char('+'), char('-')))),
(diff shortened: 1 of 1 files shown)


Geal:
if we are expecting to parse a float, that we see numbers followed by an `E`, we're expecting to see a valid exponent, the parser should not succeed if the exponent is invalid, that would be a surprising behaviour. The case in #1021 should be solved by a custom parser instead

PR #1549, opened 2022-09-12
Fix clamped capacity to be divided by element size

jkugelman:
Fixes https://github.com/Geal/nom/issues/1459#issuecomment-1244461213: the capacity of a `Vec<O>` is measured in number of elements not number of bytes, so divide 64KiB by the element size. This will round towards 0, which is desirable.


src/multi/mod.rs
```diff
@@ -20,7 +20,7 @@ use core::num::NonZeroUsize;
 ///
 /// This does not affect correctness. Nom will always read the full number
 /// of elements regardless of the capacity cap.
-const MAX_INITIAL_CAPACITY: usize = 65536;
+const MAX_INITIAL_CAPACITY_BYTES: usize = 65536;
```
Geal, line 23:
```suggestion
#[cfg(feature = "alloc")]
const MAX_INITIAL_CAPACITY_BYTES: usize = 65536;
```

Geal (approved):

Geal:
thanks!

PR #1556, opened 2022-09-28
Allow using arrays as input to `alt`

glittershark:
Now that min-const-generics is stabilized, it's possible to (backwards-compatibly) write a version of `alt` that uses a const array as input instead of being limited to only tuples - this allows us to lift the 21-element limitation on tuples passed to `alt`.


MTRNord:
Hi,

any update on this? I am running into a case where I have a grammar with 35 cases. Obviously this doesnt't work with current nom

Geal:
there were a few things missing to make the DX on this reasonable, mainly having most parsers return a concrete type instead of a closure, and having a `Either` parser. All elements of the array have to be of the same type. Now we can revisit this :)

Geal (approved):

Geal:
great work @glittershark, thanks!

src/branch/mod.rs
```diff
@@ -178,22 +178,34 @@ impl<
...
+    &mut self,
+    input: Input,
+  ) -> crate::PResult<OM, Input, Self::Output, Self::Error> {
     let mut error = None;
 
-    for branch in self {
-      match branch.parse(input.clone()) {
+    for branch in &mut self.parser {
+      match branch.process::<OM>(input.clone()) {
+        //branch.parse(input.clone()) {
```
glittershark, line 196: should this have been committed?

Geal:
good catch, fixed in 8c68e22

PR #1557, opened 2022-10-02
Update making_a_new_parser_from_scratch.md

barower:
When trying `dbg_dmp` snippet, I noticed it does not compile. This change updates this snippet so that it matches https://docs.rs/nom/latest/nom/error/fn.dbg_dmp.html


Geal:
thanks!

PR #1560, opened 2022-10-16
Fix closure syntax in error_management.md

jtracey:
I suppose it might be nice to have these examples to run as tests to prevent this sort of thing in the future; no idea how feasible that would be though.


Geal (approved):

Geal:
@jtracey thanks! I'll make sure to make those examples executable soon

Issue #1223, opened 2020-11-01
is there a "skip" or "drop", like "take"?


ckampfe:
Hi there,
Firstly, thank you so much for nom, it is a true pleasure to use. We use it regularly in production and it has been rock solid, and a real joy to develop with. Hopefully I will be able to open-source a parser we have written soon.

I have no issue to report, just a question:

I am wondering if there is a combinator that is similar to `take`, but drops/discards/skips the given number of bytes rather than returning them. Saying this another way, is there something like:

```rust
let s = skip(2usize)(s)?;
```
that simply returns the input with a certain number of prefix bytes removed, as opposed to using `take` like so:

```rust
let (s, _skipped_bytes) = take(2usize)(s)?;
```

I certainly can make a function that does this, I'm just curious if such a thing exists as a primitive in nom.
I have searched the documentation and the source and so far not been able to find anything, but I may have missed it. Thank you!

ikrivosheev:
Hello! any updates?)

epage:
Since `take` can provide the same functionality at the same performance (I presume), I'm curious what your motivation is.  Is it to make the code more declarative / read as a grammar?

karlseguin:
I'm trying to learn rust, and for the project I decided on, I need a parser and decided on nom.  I literally just started looking at this library 10 minutes ago, and I was looking for "skip". Since @ikrivosheev didn't provide a reasoning, I thought I'd provide mine.

My experience comes from [nimble_parsec](https://github.com/dashbitco/nimble_parsec) which, being Elixir, uses chaining *heavily*. It's a small thing, but I was just after the same type of streamlined code. You can see they have an `ignore` which will ignore the output of any combinator, so it wouldn't quite be `skip(3)` like `take(3)` but more generally, `skip(take(3))(s)`.

Again, in my case, it was just about removing the unnecessary extra `let (s, _) = ....`, which I think decreases the readability of the code.

(I'm not suggesting that what's idiomatic in Elixir would make sense or belongs in Rust. Obviously, I need to adapt to the Rust-way.)

epage:
@karlseguin 
> Again, in my case, it was just about removing the unnecessary extra `let (s, _) = ....`, which I think decreases the readability of the code.

Except `skip` would just change the output type to `()` at best and you would still have `let (s, ()) = ...` which isn't much different.

[...]

PR #1566, opened 2022-10-27
[issues-1223] Add InputSkip trait

ikrivosheev:
Close: https://github.com/Geal/nom/issues/1223


Xiretza:
I think this might be better implemented as a combinator `void(p)` with the semantics of `value((), p)` (which would be used like `void(take(n))`), rather than a specialized combinator for skipping a fixed number of elements.

ikrivosheev:
> I think this might be better implemented as a combinator `void(p)` with the semantics of `value((), p)` (which would be used like `void(take(n))`), rather than a specialized combinator for skipping a fixed number of elements.

`take(n)` - also create new `slice`. The idea was not to create unnecessary `slice` and immediately skip a piece

epage:
> Not everyone works only with slices, some one can work with `Read + Write` or something else. And it looks very strange that in order to skip a piece of data, you need to call `take`. It's logical to have some way to skip (real skip, not take then skip) piece of data.

To clarify, are you working with `Read + Write` and needing ti for that?

I'm also concerned that this would need a proliferation of different combinators to fully flesh out the idea of "needing" skip support (e.g. `skip_while`, `skip_until`) which would bloat `nom`.  I feel like more concrete justification would be needed to justify that.

ikrivosheev:
@Geal hello! I have rebased PR, can you review?

Issue #1567, opened 2022-10-28
Add support varint


ikrivosheev:
Hello! Thank you for the great library! I I want to add support for `varint`.

Link: https://en.wikipedia.org/wiki/Variable-length_quantity

hackaugusto:
It would be nice to have both variants for little and big endian encodings, and maybe also zig-zag encoding.

hackaugusto:
If there is any interest, this is the big-ending version w/ overflow handling that I'm using:

```rust
#[inline]
pub fn be_varint_u64<I, E: ParseError<I>>(mut i: I) -> IResult<I, u64, E>
where
    I: Slice<RangeFrom<usize>> + InputIter<Item = u8> + InputLength,
{
    let mut result = 0;

    'end: {
        let mut byte;
        for _ in 0..(u64::BITS / 8) {
            (i, byte) = be_u8(i)?;
            result = (result << 7) + u64::from(byte & 0x7F);
            if byte < 0x80 {
                break 'end;
            }
        }
        (i, byte) = be_u8(i)?;
        result = (result << 8) + u64::from(byte);
    }

    Ok((i, result))
}
```

Because of the `BITS` associated constants this can't be a generic function, but should be easy enough to write a macro for it.

PR #1568, opened 2022-10-28
[issues-1567] Add varint parsing

ikrivosheev:
Add support:

- `varint_u64`
- `varint_u128`

This PR close: https://github.com/Geal/nom/issues/1567.


Xiretza:
Wouldn't this be better implemented as a separate crate, much like https://crates.io/crates/nom-leb128? nom itself really only provides the most fundamental parsers and combinators.

ikrivosheev:
> Wouldn't this be better implemented as a separate crate, much like https://crates.io/crates/nom-leb128? nom itself really only provides the most fundamental parsers and combinators.

This make sense but i think varint is one of the fundamental parser.

Xiretza:
I think for this to go into nom proper, there would have to be parsers for every integer type at the very least; though I wouldn't put too much effort into it until @Geal confirms it'd be accepted.

src/number/streaming.rs
```diff
@@ -1538,6 +1538,79 @@ where
...
+  I: Slice<RangeFrom<usize>> + InputIter<Item = u8> + InputLength + Copy,
+  R: std::ops::AddAssign<R> + std::ops::Shl<usize, Output = R> + Default + From<u8>,
+{
+  let mut result = R::default();
+  let mut ofs: usize = 0;
+  let mut remainder = i;
+  loop {
+    let (i, byte) = le_u8(remainder)?;
+    result += (<R as From<u8>>::from(byte & 0x7F)) << ofs;
+    if byte < 0x80 {
```
hackaugusto, line 1554: I believe this should exit if enough bytes for the target data type have been collected. Otherwise the right shift or addition will overflow.

PR #1569, opened 2022-10-28
Add clippy into CI and fix clippy warnings

ikrivosheev:
## Improvements

- add `cargo clippy`
-  fix all **clippy** warnings (some of them are `allow`)
-  increase minimal rust version up to 1.50

CI example: https://github.com/ikrivosheev/nom/actions/runs/3344876947

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 autoexamples = false
 
 # also update in README.md (badge and "Rust version requirements" section)
-rust-version = "1.48"
+rust-version = "1.50"
 
 include = [
   "CHANGELOG.md",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on: [push, pull_request]
 
 env:
-  RUST_MINVERSION: 1.41.1
+  RUST_MINVERSION: "1.50"
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
 
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.48.0
+          - 1.50.0
 
         features:
           - ''
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,12 +7,17 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
 
+concurrency:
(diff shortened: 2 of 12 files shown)


ikrivosheev:
Tests are failed on **nightly** and **beta** after rustc enable **niche-filling** optimization.

Issue: https://github.com/rust-lang/rust/issues/46213
MR: https://github.com/rust-lang/rust/pull/94075

Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 autoexamples = false
 
 # also update in README.md (badge and "Rust version requirements" section)
-rust-version = "1.48"
+rust-version = "1.50"
```
Xiretza, line 17: The MSRV bump shouldn't be neccesary once #1549 is merged.

ikrivosheev: Well, you can see there is an error now: https://github.com/Geal/nom/actions/runs/3040066364

ikrivosheev: This PR fix error and also add some improvements.

Xiretza: Yes, and #1549 fixes that by using `min()` instead of `clamp()`.

ikrivosheev: Ahh, sorry, I see that!

ikrivosheev: Well, I don't see a problem to drop support for rust version which had a release 2 years ago.

cky: @ikrivosheev It is a huge problem, because it would require a major version bump. Quoting the README:

> The 7.0 series of nom supports **Rustc version 1.48 or greater**. It is known to work properly on Rust 1.41.1 but there is no guarantee it will stay the case through this major release.
>
> The current policy is that this will only be updated in the next major nom release.

MSRV guarantees are serious business for widely used libraries. It's fine if your library doesn't commit to a MSRV, but if your library does make such a commitment, you cannot break it.

epage: Hopefully 8.0 changes the MSRV guarantee to minor compatibility breaks (8.x.0) rather than major (x.0).  This is the approach of most libraries I work with and it seems to work out well.  Having more MSRV smarts in cargo would he,lp though.

Geal: MSRV will probably change for nom 8 yes, but will stay the same for nom 7. Contrary to some libraries that bump MSRV in minor versions, we can't do that for nom because those versions have to be shipped in software for distributions like debian

Geal:
closing this in favor of #1585, thanks for the help though!

PR #1576, opened 2022-11-28
impl `InputLength` for all `AsRef<str>`

arniu:


Xiretza:
I don't think this is a good idea, if forbids downstream impls for any type that also happens to be `AsRef<str>`.

Geal:
I agree with @Xiretza, this is bound to create issues in the long run. @arniu is this linked to a specific problem? I'm closing this for now, feel free to reopen if it is relevant

Issue #993, opened 2019-07-12
Parser from scratch docs is hard to follow


ottosson:
Hi, thanks for all your hard work with Nom in general, and the new version specifically. Nom 5 made me want to try it out for the first time, but I find it a bit hard to get started since I need clear and precise documentation to understand.

I'm novice in Rust and completely new to Nom, so I look at the documentation and the chapter "Making a new parser from scratch".

The start is ok, but you talk about a parser for "12345", but never really show how it's implemented and start to talk about the IResult type instead, never to come back to the 12345-parser.

In the HTTP example you declare a bunch of parsers in what seems to be the global scope of the program and not in a function which will result in a `expected item, found keyword let` error when copying the code to my own test project, which makes it hard to understand where they should be declared.

If possible I would also like to see the "Testing the parsers" and the rest of the chapters continue with the HTTP parser as an example to make the tutorial feel more coherent.

As a beginner I would love to be able to copy/paste some working examples with associated tests and just be able to play around to see what affect changes have on the working code. And the more examples you add, simple and complex, the better :)

Not sure if this is of any value, I just wanted to share an absolute beginners point of view. Hope you'll have time to write some more documentation at some point. And thanks again. Nom looks really interesting and I hope I manage to wrap my head around it with time.

shonfeder:
I agree on this point. afaik, the code in https://github.com/Geal/nom/blob/master/doc/making_a_new_parser_from_scratch.md#writing-a-first-parser is syntactically invalid, so it's completely unclear to me what it's trying to illustrate.

anahata0108:
Is there an update available on when this will be addressed? 6.0 has been released, 6.1 even, and the docs are still impenetrable to a newcomer.

eminence:
> The start is ok, but you talk about a parser for "12345", but never really show how it's implemented and start to talk about the IResult type instead, never to come back to the 12345-parser.


I would also like to see the guide follow through on this :)

PR #1577, opened 2022-12-02
Replaces broken tutorial code with working example

ltrujello:
Hi, 

This is a simple PR to fix some code in the documentation. The documentation itself is well written and gives newcomers a great idea for how this library is meant to be used, but the example code doesn't work. Others have pointed this out [here](https://github.com/Geal/nom/issues/993).  

I followed what the documentation code was attempting to do (it was trying to use the library as an example to parse HTTP request lines). So I put together the bare minimum to get the intended example working, keeping the original comments/typing as well. 

I think this will get newcomers up and running with the library much faster and they'll be more motivated to use the library if the example code is working. Personally I still found the docs useful in understanding this library.

### DIFF
--- a/doc/making_a_new_parser_from_scratch.md
+++ b/doc/making_a_new_parser_from_scratch.md
@@ -99,29 +99,83 @@ As an example, here is how we could build a (non spec compliant) HTTP request
 line parser:
 
 ```rust
+use nom::{
+    bytes::complete::{tag, take_while1},
+    character::{is_alphabetic, is_space},
+    sequence::{preceded, tuple},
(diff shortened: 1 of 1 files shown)


epage:
As an alternative, I've been experimenting with different ideas for nom in a fork which includes pulling the various markdown files into docs.rs which 
- Puts all documentation in one place
- Ensures the documentation at least compiles

See https://github.com/epage/nom-experimental/pull/1

ltrujello:
Looks very cool! Feel free to take what I wrote up if you like.

PR #1578, opened 2022-12-03
New combinator: `multi::reduce`

bertptrs:
This PR adds a new combinator: `multi::reduce`. It is essentially a special case of `multi::fold_many1` where the accumulator type is the same as the parsed type.

Analogous to `Iterator::reduce`, this new combinator takes the first parsed element as the initial value rather than use a separate initializer function. This is useful in many cases, but most commonly for when you do not have an easy base element for your reduction, for example when computing a minimum over a list of elements without a well-defined absolute maximum.

Aside from this new combinator, the PR also includes the following changes (in separate commits for easy review):

- Fix a test for a specific struct size. Apparently rustc got smarter?
- Refactor shared code between `fold_many0` and `fold_many1`
- **Breaking:** change the infinite loop error for `fold_many1` from `Error` to `Failure`. I believe this was always the correct version since it indicates a programming error and not a data issue.

Tests for new code included.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -417,6 +417,7 @@ pub enum ErrorKind {
   Float,
   Satisfy,
   Fail,
+  Reduce,
 }
 
 #[rustfmt::skip]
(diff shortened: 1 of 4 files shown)


bertptrs:
This PR is now almost a year old and has various conflicts with the main branch so it cannot be merged but I would still like the combinator added here.

Is this something that could be added? If so, I'll clean up the PR so that it works again.

PR #1591, opened 2022-12-16
doc: wording in choosing_a_combinator

Fumon:
I took a stab at introducing some new wording.
- Surfaces `and_then` and its functional variant `map_parser` which I didn't find while first learning Nom.
- Attempts to provide some clarity that I found was lacking around `flat_map`
- Reorders the Modifiers list to try and draw parallels between Trait and Combinator versions.

While I was getting up to speed on Nom, discovering the correct combinator was much harder than discovering a trait function. This was difficult as the example code and explanation usually lives on the combinator. This lead to me using the couple of combinators and base parsers introduced through tutorials and examples, and getting stuck a lot reading this page and things not fitting in the slots like I expected.

Happy for feedback and to change things to fit. If this feels like the right kind of wording, I may go in and update trait function documentation to make things a little friendlier for discovery of the related combinators through intellisense. Also feel free to tell me to go make a blog post instead.

### DIFF
--- a/doc/choosing_a_combinator.md
+++ b/doc/choosing_a_combinator.md
@@ -75,11 +75,15 @@ The following parsers could be found on [docs.rs number section](https://docs.rs
 
 ## Modifiers
 
-- [`cond`](https://docs.rs/nom/latest/nom/combinator/fn.cond.html): Conditional combinator. Wraps another parser and calls it if the condition is met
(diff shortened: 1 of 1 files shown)


Geal (approved):

Geal:
this is great, thanks

PR #1596, opened 2022-12-28
fix build issues on different rust versions

Geal:
* `clamp` is not available in 1.48
* since https://github.com/rust-lang/rust/pull/94075 that shipped in 1.65, a test asserting the size of `IResult` is failing because the structure is now small by 8 bytes


Xiretza:
#1549 should take care of the `clamp()` and another issue in the same area.

Geal:
right, that works better

PR #1608, opened 2023-01-02
feature multi range

### LINKED ISSUE #1393: Consolidate parser variants using ranges (e.g. `many0`, `many_m_n`)

Geal:
Fix #1393
follow up for #1402


cenodis:
Thank you for merging this. The current implementation only contains range variants for the `many` and `fold` parsers. More parsers were mentioned in #1402 so I have opened a followup issue #1613 to make sure those don't get forgotten.

Issue #1535, opened 2022-07-06
Allow types to be both stream and complete parsers


epage:
Right now, `nom` has distinct types for complete and streaming parsers.

Problems with this
- Requires duplicate module hierarchies (#1414 would move them around)
- Can't provide convenience implementations for types like `&str` because one of the two has to be offered (#1408)
- Defers programming errors to run time that make catching them hard (require incomplete cases)
  - `Finish::finish` will [panic on Incomplete](https://docs.rs/nom/latest/src/nom/internal.rs.html#36-44)
  - Accidentally using a parser intended for one in the context of the other

What if a type could implement have both types of parsing on it?

epage:
`combine` has a single trait with two distinct parse methods.  This doesn't work well with automatically implementing the parse trait for functions because you don't know which will be involved.

epage:
What if we had
- `Parser` that used an `IResult` with a `CompletedErr`
- `StreamParser` that used an `IResult` with a `StreamErr`

This would
- make the [return types distinct](https://docs.rs/nom/latest/nom/type.IResult.html), catching "wrong parse method" errors. 
- hopefully allow blanket impls of the traits on functions based on the `Err` used

We'd need to have `to_completed` and `to_streamed` conversion functions on what is currently the `Finish` trait to allow mixing parsers.

Stargateur:
The following is a idea I have while doing my own parser. This is maybe not possible to do in nom or would require a lot of change but I don't see why not share ideas.

The idea was to not have streaming and complete parser like nom. I wanted to remove the problem of duplicate code in nom but I also wanted to have a way to make "streaming" parser like nom, currently nom have a major problem you can't "resume" where you stop. I saw two solutions:

* use async but: this require async trait for now that allocate a lot. Also this have a cost for user that doesn't need async. This could be a good solution in the future when we could have async trait maybe this would also require some degree of duplicate code I didn't push this a lot yet.
* shift this responsibility to `Input`. That what I go for in my current experimentation. In nom the trait [`InputIter`](https://docs.rs/nom/7.1.1/nom/trait.InputIter.html) this would allow to do mostly what I did.

For example:

```rust
#[derive(Clone)]
struct Buf<Reader, const N: usize> {
  buf: Vec<u8>,
  reader: Reader,
}

pub(crate) enum Position {
  RangeFrom(RangeFrom<usize>),
  Range(Range<usize>),
}

pub struct ReaderStream<Reader, const N: usize> {
[...]

PR #1614, opened 2023-01-07
test the Streaming wrapper on HTTP benchmarks

Geal:
This is an experiment in merging streaming and complete implementations of parsers, using a condition on the input type, loosely inspired from https://github.com/epage/nom-experimental/pull/28

### DIFF
--- a/benchmarks/benches/http_streaming.rs
+++ b/benchmarks/benches/http_streaming.rs
@@ -4,7 +4,7 @@
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
 use criterion::*;
-use nom::{IResult, bytes::streaming::{tag, take_while1}, character::streaming::{line_ending, char}, multi::many};
+use nom::{IResult, bytes::{tag,  take_while1}, character::{line_ending, char}, multi::many, Streaming};
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 #[derive(Debug)]
@@ -67,7 +67,7 @@ fn is_version(c: u8) -> bool {
   c >= b'0' && c <= b'9' || c == b'.'
 }
 
-fn request_line(input: &[u8]) -> IResult<&[u8], Request<'_>> {
+fn request_line(input: Streaming<&[u8]>) -> IResult<Streaming<&[u8]>, Request<'_>> {
   let (input, method) = take_while1(is_token)(input)?;
   let (input, _) = take_while1(is_space)(input)?;
   let (input, uri) = take_while1(is_not_space)(input)?;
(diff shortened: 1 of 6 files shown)


Geal:
@epage it looks like the const bool is not necessary, a static method on the `Input` trait is enough information for the compiler to only generate the version requested by the input type

epage:
> @epage it looks like the const bool is not necessary, a static method on the `Input` trait is enough information for the compiler to only generate the version requested by the input type

I had originally intended to just have [a function on the trait](https://github.com/rust-bakery/nom/issues/1535#issuecomment-1312354167) but switched to `const bool` through development.
- A paranoid part of me wanted the contract very clear and without ways to circumvent it
- The `const bool` allows something to only be implemented for one or the other
  - Example: Distinct `InputTakeAtPosition` traits.  I had backed off from this, mostly for an easier migration path for users.  The trait merger also complicates ever doing this again.
  - Example: limiting `Finish` to only complete parsers.  [`Finish` panics on `Incomplete`](https://github.com/rust-bakery/nom/blob/main/src/internal.rs#L40) and a user has to be careful with tests to realize it would do that and likely a user using a Streaming parser should be deconstructing the error anyways, so limiting `Finish` helps guide people to the right solution earlier in the process.

To be clear, these examples are meant to be indicative of what can be done.

Of course, going the `const bool` route comes at the cost of an extra generic parameter when users have to explicitly list then.

In general, I would recommend looking over my notes in #1535 as they will provide more context and might give other ideas.

Geal:
> - A paranoid part of me wanted the contract very clear and without ways to circumvent it
Circumvent what? 
> - The `const bool` allows something to only be implemented for one or the other
>   - Example: Distinct `InputTakeAtPosition` traits.  I had backed off from this, mostly for an easier migration path for users.  The trait merger also complicates ever doing this again.
This is not the path I'm going towards, I want to simplify here. I've tried various ways to write the low level parts of streaming and complete parsers over the years, it does not need new ways to specialize implementations. It mainly needs a way to merge them.
>   - Example: limiting `Finish` to only complete parsers.  [`Finish` panics on `Incomplete`](https://github.com/rust-bakery/nom/blob/main/src/internal.rs#L40) and a user has to be careful with tests to realize it would do that and likely a user using a Streaming parser should be deconstructing the error anyways, so limiting `Finish` helps guide people to the right solution earlier in the process.

Making it clear in the documentation that `Finish` can panic is enough, I don't want the implementation to overindex on a convenience trait that amounts to a match.
> 
> To be clear, these examples are meant to be indicative of what can be done.
> 
> Of course, going the `const bool` route comes at the cost of an extra generic parameter when users have to explicitly list then.
> 
> In general, I would recommend looking over my notes in #1535 as they will provide more context and might give other ideas.

I read them again, thank you for the context and the exploration. Here I'll choose the simpler solution though, I want that change to be minimally invasive, because I'll probably have to introduce more complex changes in that release (like the GAT idea)

PR #1616, opened 2023-01-10
Fix example in doc comment for `pair()`

### LINKED ISSUE #1615: Doc-comment example for `pair()` does not work as displayed
- nom version : 7.1.2

The example code shown for `nom::sequence::pair()` does not work as-is. An additional `use` is needed: `use nom::{error::ErrorKind, Err};`

## Problem

The example code for `nom::sequence::pair` displays like this, and does not compile (https://docs.rs/nom/latest/nom/sequence/fn.pair.html):

image

## Fix

Add `use nom::{error::ErrorKind, Err};` to the example. These are in the doc-comment, but are hidden in the generated documentation for some reason. My expectation was that I should be able to copy-paste examples and have them work as-is.

The following code compiles and runs as expected:

```rust
use nom::sequence::pair;
use nom::{error::ErrorKind, Err};

let mut parser = pair(tag("abc"), tag("efg"));

assert_eq!(parser("abcefg"), Ok(("", ("abc", "efg"))));
assert_eq!(parser("abcefghij"), Ok(("hij", ("abc", "efg"))));
assert_eq!(parser(""), Err(Err::Error(("", ErrorKind::Tag))));
assert_eq!(parser("123"), Err(Err::Error(("123", ErrorKind::Tag))));
```

zanedp:
Fixes #1615

### DIFF
--- a/src/sequence/mod.rs
+++ b/src/sequence/mod.rs
@@ -13,11 +13,11 @@ use crate::internal::{IResult, Parser};
 /// * `first` The first parser to apply.
 /// * `second` The second parser to apply.
 ///
+/// # Example
 /// ```rust
-/// # use nom::{Err, error::ErrorKind, Needed};
-/// # use nom::Needed::Size;
 /// use nom::sequence::pair;
 /// use nom::bytes::complete::tag;
+/// use nom::{error::ErrorKind, Err};
 ///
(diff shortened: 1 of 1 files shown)


Geal (approved):

Geal:
thanks!

PR #1618, opened 2023-01-10
avoid panic when counting zero-sized outputs in count()

### LINKED ISSUE #1617: `count` panics when parser returns zero-sized `()`
## Prerequisites

- Rust version : `rustc 1.66.0 (69f9c33d7 2022-12-12)`
- nom version : 7.1.2
- nom compilation features used: default (std)

## Test case

Example test case [(Playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d67b1d4f394814329a842fde81f9fe67):

```
use nom::{combinator::map, bytes::complete::tag, error::Error, multi::count};

fn main() {
    let parser = map(tag::<_, _, Error<&str>>("abc"), |_| ());
    let result = count(parser, 3)("abcabcabcdef").expect("parsing should succeed");
    println!("{:?}", result);
}
```

Expected: Runs fine, prints:

```
("def", [(), (), ()])
```

Result:

```
Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.99s
     Running `target/debug/playground`
thread 'main' panicked at 'attempt to divide by zero', /playground/.cargo/registry/src/github.com-1ecc6299db9ec823/nom-7.1.2/src/multi/mod.rs:576:32
note: [run with `RUST_BACKTRACE=1` environment variable to display a backtrace](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0487bcd80ada6e9943cd7b41b3dcdcbb#)
```

Shadow53:
## Prerequisites

Resolves #1617.

- Added a test case to `tests/issues.rs`
-  I also annotated an unrelated test-looking function with `#[test]` since it looked like it should be marked as such

## Code style

> This project follows a [code style](https://github.com/Geal/nom/blob/main/rustfmt.toml)
> checked by [rustfmt][https://github.com/rust-lang-nursery/rustfmt].

> Please avoid cosmetic fixes unrelated to the pull request. Keeping the changes
> as small as possible increase your chances of getting this merged quickly.

There were a couple of files that were modified by `cargo fmt` that I did not edit, so I undid those changes.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -579,7 +579,8 @@ where
 {
   move |i: I| {
     let mut input = i.clone();
-    let max_initial_capacity = MAX_INITIAL_CAPACITY_BYTES / crate::lib::std::mem::size_of::<O>().max(1);
+    let max_initial_capacity =
+      MAX_INITIAL_CAPACITY_BYTES / crate::lib::std::mem::size_of::<O>().max(1);
     let mut res = crate::lib::std::vec::Vec::with_capacity(count.min(max_initial_capacity));
 
     for _ in 0..count {
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -579,7 +579,7 @@ where
 {
   move |i: I| {
     let mut input = i.clone();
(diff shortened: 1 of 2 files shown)


Geal (approved):

Geal:
Thanks!

PR #1626, opened 2023-01-21
Parser associated types

Geal:


src/internal.rs
```diff
@@ -410,82 +412,100 @@ impl<I, O1, O2, E, F: Parser<I, O1, E>, G: FnMut(O1) -> O2> Parser<I, O2, E> for
...
   }
 }
 
 /// Implementation of `Parser::flat_map`
 #[cfg_attr(nightly, warn(rustdoc::missing_doc_code_examples))]
-pub struct FlatMap<F, G, O1> {
+pub struct FlatMap<F, G> {
   f: F,
   g: G,
-  phantom: core::marker::PhantomData<O1>,
```
epage, line 465: Nice that this lets you drop the phantom data

epage:
Any hint if this improved type inference at all?  Or any other benefit noticed besides dropping the number of generic parameters / phantom data?

src/internal.rs
Geal: yes, it simplifies them nicely. I have to check again the methods like `flat_map`, to make sure that they require all the trait bounds, otherwise it would be possible for them to return a value that does not implement `Parser`, which would be confusing

Geal:
so far the main issue with type inference was that I had to add the error type explicitely (`F: Parser<I, Error=E>` instead of `F: Parser<I>`) in most combinators. It is possible to remove the error generic type, like I've done for [many0](https://github.com/rust-bakery/nom/pull/1626/files#diff-bb724cea4e6803b90537697b87afb638d340a58d2d49600e8b684022cb5fcceeR57-R62), but this makes the type signatures harder to read

epage:
Just gave this a try in winnow-rs/winnow#220 and it looks like the compiler isn't happy about implementing `Parser` for `u8` and other built-ins because there is nothing to constrain the generic `E` parameter on that is used for the `Error` associated type.
```rust
impl<I, E> Parser<I> for u8
where
    I: StreamIsPartial,
    I: Stream<Token = u8>,
    E: ParseError<I>,
{
    type Output = u8;
    type Error = E;

    fn parse_next(&mut self, i: I) -> IResult<I, u8, E> {
        crate::bytes::one_of(*self).parse_next(i)
    }
}
```
```
error[E0207]: the type parameter `E` is not constrained by the impl trait, self type, or predicates
   --> src/parser.rs:606:9
    |
606 | impl<I, E> Parser<I> for u8
    |         ^ unconstrained type parameter
```

This would require hard coding the error type which would be unpleasant to then make it seemless with other parts of the API.

I guess only `O` could be turned into an associated type.  That would solve most type inference issues.  The inconsistency on output types isn't great though.

epage:
I found doubling down on `PhantomData` was another way of resolving the type inference issues: winnow-rs/winnow#222

PR #1631, opened 2023-01-29
Add an OutputMode type parameter to drive parser results

Geal:
this work takes inspiration from [chumsky's exploration of generic associated types](https://github.com/zesterer/chumsky/pull/82/). In chumsky's PR, output value creation depends on a `Mode` type parameter to either produce the value, or return `()`, and the call site decides if it needs the value or not. That mode can be transmitted from parent to child parser, and prevent the compiler from generating the code for an entire chain of results.
As an example, in the `delimited` combinator, we test tree parsers in sequence, but we only care about the result of the second one. For the other two parsers, we only need to know that they succeeded, we do not need their result, so they might as well not be generated.

This PR extends this solution further, by having a mode parameter for errors too: in a lot of cases, nom combinators just need to know that a parser failed, but do not want to process the error. The `OutputMode` type also carries a parameter that will drive streaming VS complete behaviour.

TODO:
- [ ] validate conversion of `multi` combinators
- [ ] validate conversion of `branch` combinators
- [ ] validate conversion of `number` combinators
- [ ] rewrite some tests and examples to avoid the `.parse` syntactic noise
- [ ] test the effect on benchmarks
- [ ] convert the rest of the parsers
- [ ] docs


src/internal.rs
```diff
@@ -242,7 +361,16 @@ pub trait Parser<Input> {
 
   /// A parser takes in input type, and returns a `Result` containing
   /// either the remaining input and the output value, or an error
-  fn parse(&mut self, input: Input) -> IResult<Input, Self::Output, Self::Error>;
+  fn parse(&mut self, input: Input) -> IResult<Input, Self::Output, Self::Error> {
+    self.process::<OutputM<Emit, Emit, Streaming>>(input)
+  }
```
epage, line 366: `Parser::parse` is being hard coded to `Streaming`?  That seems odd.  I assume I am I missing something about this streaming design

src/internal.rs
```diff
@@ -242,7 +361,16 @@ pub trait Parser<Input> {
...
+
+  /// A parser takes in input type, and returns a `Result` containing
+  /// either the remaining input and the output value, or an error
+  fn process<OM: OutputMode>(
+    &mut self,
+    input: Input,
+  ) -> PResult<OM, Input, Self::Output, Self::Error>;
```
epage, line 373: > test the effect on benchmarks

I'm curious to see if you are able to see the dramatic savings that chumsky did or if some of that is more inherent to their design.  My gut feeling is that in a lot of cases, the overhead from a lot of these outputs and error should be pretty low with nom's design as-is and that this wouldn't offer much improvement unless people were going heavy on complex `preceded`  / `termianted` parsers which json doesn't do and I've not done in my various parsers.

Besides runtime performance, it might also be good to check the impact to binary size or release binaries as this might cause more instances of the generic functions to be generated.  I'd also be curious about compile times but I'm assuming that won't be big enough to notice.

src/internal.rs
```diff
@@ -233,6 +234,124 @@ where
...
+  /// TODO
+  type Incomplete: IsStreaming;
+}
+
+/// TODO
+pub trait IsStreaming {
+  /// TODO
+  fn incomplete<E, F: FnOnce() -> E>(needed: Needed, err_f: F) -> Err<E>;
+  /// TODO
+  fn is_streaming() -> bool;
```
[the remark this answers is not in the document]
Geal: I'm not decided on that yet. Right now my thinking is that having that at runtime would generate a lot of code in all cases, while having it static would depend only on the call site: if you want both, you get both implementations, but probably with some code duplication

src/internal.rs
Geal: yes, that part is not written yet. That might end up with two methods, like `parse` and `parse_streaming`

src/internal.rs
Geal: currently most of nom's overhead is in creating values and errors and discarding them, and that will be even more present with heavy error types. So far I have not measured it though, because I have not converted enough code to affect the benchmarks. I'm worried about compilation times, though. That might be a mix of increasing compilation times due to heavy use of impl traits (returning closures was erasing type information), but decreasing it due to less code  being generated.

src/internal.rs
```diff
@@ -242,7 +361,16 @@ pub trait Parser<Input> {
 
   /// A parser takes in input type, and returns a `Result` containing
   /// either the remaining input and the output value, or an error
-  fn parse(&mut self, input: Input) -> IResult<Input, Self::Output, Self::Error>;
+  fn parse(&mut self, input: Input) -> IResult<Input, Self::Output, Self::Error> {
```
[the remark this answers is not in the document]
Geal: yes, to work properly, the entire chain, at least in what nom provides, should support it. I'd like to see if converting the existing way of writing parsers (functions returning closures) to this would not be too noisy

src/internal.rs
```diff
@@ -343,12 +477,15 @@ macro_rules! impl_parser_for_tuple {
...
+      fn process<OM: OutputMode>(&mut self, i: I) -> PResult<OM, I, Self::Output, Self::Error> {
         let ($(ref mut $parser),+,) = *self;
 
-        $(let(i, $output) = $parser.parse(i)?;)+
+        // FIXME: is there a way to avoid producing the output values?
+        $(let(i, $output) = $parser.process::<OutputM<Emit, OM::Error, OM::Incomplete>>(i)?;)+
 
-        Ok((i, ($($output),+,)))
+        // ???
+        Ok((i, OM::Output::bind(|| ($($output),+,))))
```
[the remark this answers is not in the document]
Geal: I haven't fixed that one yet, passing the mode along should be enough, but ideally I'd like a way to selectively discard some tuple elements as you said in https://github.com/winnow-rs/winnow/issues/82

Geal:
@epage this is still pretty early, so don't overindex on the current state of the code and API, I'm still exploring how it can work in the end

src/internal.rs
epage: > currently most of nom's overhead is in creating values and errors and discarding them

Can you qualify this with some more details?

For example, if you are doing `delimited(char('['), <something>, char(']'))`, then I would expect the value wouldn't have much overhead.  From the nom examples I've looked at and in my own parsers, this is a fairly common case.  Is the overhead more than I expect or is this for more complex usage patterns?  chumsky saw the gains on `json` but I see these more simple cases in nom's json benchmark.

One possible gain is people not discover the `_count` variant of `many` parsers.  I know I make that mistake.  However, that also ties into needing to also pay attention to using `terminated` / `preceded` rather than tuples     Something I was considering was an  alternative trait to `Extend` that (1) allowed pre-allocation and (2) was implemented for `usize` and `()`, replacing `_count` variants.  This is a bottom-up driven approach to not creating expensive values, rather than the Mode's top-down approach.  One benefit to top-down is when reusing parsers between throw-away and non-throwaway contexts.  So far, I've not needed to do it too much and I would be interested in looking at cases for those that do and how much it impacts performance.

However, I can see it in errors with `alt`.  With `combine`, I did see performance overhead from their version of `alt` with heavy errors.  On the other hand, I suspect it'd be better to replace it with combine's  [dispatch](https://docs.rs/combine/latest/combine/macro.dispatch.html).  It is a macro that lets you write a `match` with parsers which cuts down on the number of parse attempts made, reducing overhead further.  I can only speak for switching `choice` / `alt` to `dispatch` but it  dramatically sped up my code.  While this can be done with an imperative parser in `nom`, I found [duplicating `dispatch`](https://github.com/toml-rs/toml/blob/main/crates/toml_edit/src/parser/macros.rs) made it more ergonomic which makes it more likely to be used.   I suspect some of my speed gains over `combine` were from finding another place I could insert a `dispatch` call.

Even with that, I've noticed my performance is sensitive to changes in my error type and I've intentionally kept  it simple, putting all of the expensive operations when converting to the public error type for my crate.  An `opt` that doesn't apply can still easily do a single allocation that is thrown away.  Dropping that allocation and dropping any of the panic handling around allocations can be beneficial.  I was considering a non-GAT approach here as well by making the backtracking error type accepted by `opt` to be `()`).  GATs do simplify the process.

epage:
> @epage this is still pretty early, so don't overindex on the current state of the code and API, I'm still exploring how it can work in the end

I understand and am commenting now in case any of my thoughts are helpful for that exploratory process as I've been thinking about parallel problems.  If its not helpful at this stage, I can back off.

src/multi/mod.rs
```diff
@@ -55,36 +59,61 @@ const MAX_INITIAL_CAPACITY_BYTES: usize = 65536;
...
 {
-  move |mut i: I| {
-    let mut acc = crate::lib::std::vec::Vec::with_capacity(4);
+  Many0 { parser: f }
+}
+
+/// Prser implementation for the [many0] combinator
+pub struct Many0<F> {
+  parser: F,
+}
```
epage, line 74: In winnow-rs/winnow#6, I gave up on moving the parsers to be structs in part because of type inference issues.  Looking back at the errors, it looks like it all came down to my `PhantomData`:
```
error[E0282]: type annotations needed
Error:     --> src/number/complete.rs:1411:9
     |
1411 |         map(tuple((digit1, opt(pair(char('.'), opt(digit1))))), |_| ()),
     |         ^^^ cannot infer type of the type parameter `PO` declared on the function `map`
     |
help: consider specifying the generic arguments
     |
1411 |         map::<_, PO, _, I, O, E>(tuple((digit1, opt(pair(char('.'), opt(digit1))))), |_| ()),
     |            +++++++++++++++++++++
```
*(`PO`, `I`, `O`, and `E` were `PhantomData` fields)*

I suspect rust-bakery/nom#1626 would help with this type inference problem and, now looking at it, I think it does, making life easier on this PR than it was for me on winnow-rs/winnow#6.

Geal:
it's able to run the JSON parser now (not fully, still missing the float parsing), and so far it looks slower (between 3% and 16%). I do not know yet where that comes from, I suspect that either there's an issue with recursin, or that some parts are not properly inlined. The UX looks ok, it requires adding a `.parse()` here and there but it would not be too traumatic to convert a nom 7 parser, and it's possible to convert selected parts of a parser as needed, while keeping the functions everywhere else.

Writing an implementation of `Parser::process` is a bit annoying though, I'll spend more time on UX

epage:
I've been finding the json benchmark is too trivial to get reasonable numbers from it, so I've added a larger one: https://github.com/winnow-rs/winnow/pull/119/commits/1a4889380045e1da953eba7b5e71a2b365000ab7

FYI the `dispatch!` macro dropped json parse time from 12ms to 9ms for the large file: https://github.com/winnow-rs/winnow/pull/119

Geal:
Yes, I've been testing with the canada file too, that's where I get a 3% difference.
`dispatch` is a good idea, I wrote similar switches in other parsers but never took the time to add it back.
FWIW the goal of that benchmark was not to make the fastest json parser, but to check the evolution between nom versions with a naive parser, so that's why I have not optimized it much

Geal:
flamegraph of execution from main:

flamegraph-main-no-gat

flamegraph of execution from this PR:

flamegraph-gat

epage:
> `dispatch` is a good idea, I wrote similar switches in other parsers but never took the time to add it back.

Ah, I should have clarified.  As `dispatch!` is probably the closest to the idealized result of GAT, the benchmark I ran provides an theoretical lower limit on parse times for the `alt` calls I replaced.

> FWIW the goal of that benchmark was not to make the fastest json parser, but to check the evolution between nom versions with a naive parser, so that's why I have not optimized it much

Yes, I've kept a naive version which is what I'd highlight in winnow's cookbook.  I also am considering writing an optimization guide that show cases the various stages of optimization and report the benchmark results so people can see the relative impact of various changes.

epage:
Are your benchmarks with `Error` or `VerboseError`?  While I'm surprised by a slow down, I would expect more of a speed up with an error type that does more processing.

tisonkun:
It seems customized parser become harder after this patch.

Previously works:

```rust
fn comma_separated_list1<'a, T>(
    f: impl FnMut(Input<'a>) -> ParseResult<'a, T>,
) -> impl FnMut(Input<'a>) -> ParseResult<'a, Vec<T>> {
    separated_list1(Comma, f)
}
```

Now breaks:

```
error[E0277]: expected a `FnMut<(&'a [Token<'a>],)>` closure, found `impl Parser<&[Token<'_>], Output = Vec<<impl FnMut(Input<'a>) -> ParseResult<'a, T> as Parser<&[Token<'_>]>>::Output>, Error = parser::error::ParseError>`
   |
67 | ) -> impl FnMut(Input<'a>) -> ParseResult<'a, Vec<T>> {
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an `FnMut<(&'a [Token<'a>],)>` closure, found `impl Parser<&[Token<'_>], Output = Vec<<impl FnMut(Input<'a>) -> ParseResult<'a, T> as Parser<&[Token<'_>]>>::Output>, Error = parser::error::ParseError>`
68 |     separated_list1(Comma, f)
   |     ------------------------- return type was inferred to be `impl Parser<&[Token<'_>], Output = Vec<<impl FnMut(Input<'a>) -> ParseResult<'a, T> as Parser<&[Token<'_>]>>::Output>, Error = ParseError>` here
   |
   = help: the trait `FnMut<(&'a [Token<'a>],)>` is not implemented for `impl Parser<&[Token<'_>], Output = Vec<<impl FnMut(Input<'a>) -> ParseResult<'a, T> as Parser<&[Token<'_>]>>::Output>, Error = parser::error::ParseError>`

```

tisonkun:
@epage you're right. I actually change the return type when debugging, but not for the argument.

After changing the argument also, it seems work now:

```rust
fn comma_separated_list1<'a, T>(
    f: impl Parser<Input<'a>, Output = T, Error = ParseError>,
) -> impl Parser<Input<'a>, Output = Vec<T>, Error = ParseError> {
    separated_list1(Comma, f)
}
```

... while I may introduce a local trait bound to reduce the longy type name.

epage:
For `trait ParserBound`, did you add a blanket impl for `Parser<Input<'a>, Output = O, Error = ParseError>`?

tisonkun:
> blanket impl

I think I should do something like that. But actually I don't know what does it mean and how to do 🤣

epage:
```rust
pub(self) trait ParserBound<'a, O>:
    Parser<Input<'a>, Output = O, Error = ParseError>
{
}

impl<'a, P, O> ParserBound<'a, O> for P
where
    P: Parser<Input<'a>, Output = O, Error = ParseError>
{
}
```

tisonkun:
@epage Thank you! It works.

PR #1632, opened 2023-01-31
For `many` and `fold`, consider `..0` an invalid range, just as `0..0` is

cky:
`nom::multi::many` and `nom::many::fold` consider empty ranges, as checked by `NomRange::is_inverted`, to be invalid. Unfortunately, that method is incorrectly implemented for `RangeTo`, meaning that `..0` is not considered invalid when it should be.

While I'm here, I also fixed up off-by-one errors in `RangeFrom::bounded_iter` and `RangeFull::bounded_iter`.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1251,23 +1251,15 @@ impl NomRange<usize> for Range<usize> {
   }
 
   fn is_inverted(&self) -> bool {
-    !(self.start < self.end)
+    self.is_empty()
   }
 
   fn saturating_iter(&self) -> Self::Saturating {
-    if self.end == 0 {
-      1..0
-    } else {
-      0..self.end - 1
-    }
+    0..self.end - 1
   }
 
   fn bounded_iter(&self) -> Self::Bounded {
-    if self.end == 0 {
-      1..0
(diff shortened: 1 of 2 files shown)


src/traits.rs
```diff
@@ -1251,23 +1251,15 @@ impl NomRange<usize> for Range<usize> {
...
+    self.is_empty()
   }
 
   fn saturating_iter(&self) -> Self::Saturating {
-    if self.end == 0 {
-      1..0
-    } else {
-      0..self.end - 1
-    }
+    0..self.end - 1
```
cenodis, line 1258: Wouldn't this panic when overflow checking is enabled (such as when building with debug)?
And even if the underflow does not panic it would return `0..usize::MAX` as the iterator for a range of `x..0`. This is incorrect because the range should be empty. Similar reasoning applies to the other range implementations.

cenodis (commented):
Thanks for catching these. It has been a long while since I last touched this piece of the code but the off-by-one fixes seem correct to me. I'm just not sure about removing the explicit `self.end == 0` checks. Those still seem necessary to me.

PR #1643, opened 2023-02-22
fix take_while_m_n for multibyte UTF-8 chars

### LINKED ISSUE #1630: `take_while_m_n` is invalid for multi-byte UTF-8 characters
`take_m_n` does:
```rust
    match input.position(|c| !cond(c)) {
      Some(idx) => {
        if idx >= m {
          if idx <= n {
            let res: IResult<_, _, Error> = if let Ok(index) = input.slice_index(idx) {
              Ok(input.take_split(index))
            } else {
              Err(Err::Error(Error::from_error_kind(
                input,
                ErrorKind::TakeWhileMN,
              )))
            };
```

while `Input::position` is defined as
```rust
  /// Returns the byte position of the first element satisfying the predicate
  fn position<P>(&self, predicate: P) -> Option<usize>
  where
    P: Fn(Self::Item) -> bool;
```
So it is looking up the byte position of the first non-matching character but then treats it as the `Item` count, comparing it with `m` and `n` and then using `slice_index` to convert it to a byte position.  This should cause it to hit `m` and `n` sooner than it should and split at the wrong position.

I see that #913 was previously reported but the 7483885 and #1097 just added the "`Item` count to byte position" conversion but didn't address that it was using a byte position.

The reason the test from #913 is working is the else-clause for when there are more valid elements than `n` (4 is greater than 1), it caps it by `n` (1), making the  `Item` count valid that it passes to `slice_index`.

To show this failing, we need to change `m`, `n`, and the input slightly
```rust
  #[test]
  fn take_while_m_n_utf8() {
    named!(parser<&str, &str>, take_while_m_n!(1, 2, |c| c == 'A' || c == '😃'));
    assert_eq!(parser("A!"), Ok(("!", "A")));
    assert_eq!(parser("😃!"), Ok(("!", "😃")));
  }
```
This fails, with the left-hand side reporting `Ok(("", "😃!")`.  The only reason it doesn't crash is that `slice_index`, when exhausted, returns `self.len()`.

I believe this problem predates #1612

ackxolotl:
fixes #1630

### DIFF
--- a/src/bytes/complete.rs
+++ b/src/bytes/complete.rs
@@ -3,7 +3,7 @@
 use crate::error::ErrorKind;
 use crate::error::ParseError;
 use crate::internal::{Err, IResult, Parser};
-use crate::lib::std::result::Result::*;
+use crate::lib::std::{result::Result::*, cmp::min};
 use crate::traits::{Compare, CompareResult, FindSubstring, FindToken, InputLength, ToUsize};
 use crate::Input;
 
@@ -208,7 +208,7 @@ where
   }
 }
 
-/// Returns the longest (m <= len <= n) input slice  that matches the predicate.
+/// Returns the longest (m <= len <= n) input slice that matches the predicate.
 ///
 /// The parser will return the longest slice that matches the given predicate *(a function that
 /// takes the input and returns a bool)*.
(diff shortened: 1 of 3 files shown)


src/bytes/complete.rs
```diff
@@ -242,55 +243,27 @@ where
 {
   move |i: I| {
     let input = i;
+    let input_len = input.input_len();
+
+    // get m and n in bytes
+    let m = match input.slice_index(m) {
```
Geal, line 249: calling slice_index twice before even processing the predicate can be very inefficient for strings, since it has to iterate over chars. Do you see another way to do it?

ackxolotl: Hm. We have to get the byte offsets for `m` and `n` before we can return any result and it doesn't really make a difference if we do that before or after matching the pattern, right? The only alternative I can think of is changing `Input::position()` to return a byte offset for `&[u8]` and a char offset for `&[str]`, then we are consistently using either byte or char offsets in `take_while_m_n` and can skip those calls to `Input::slice_index()`. But I'm not sure how much that would break.

Geal: none of those sound like good alternatives. every "split" function should only work with byte offsets, but we cannot compare those with m and n. I feel like the best approach would be to reimplement manually with the `iter_elements` method, which for `&str` calls into `char_indices`, instead of using `position`. Like, iter while the condition is satisfied, stop after n iterations or when the condition fails, then check that the number of iterations was higher than m

epage: If it helps, winnow-rs/winnow#130 is how I fixed this though there will be slight differences due to the design of `Input` vs `Stream`

src/bytes/complete.rs
```diff
@@ -242,55 +242,26 @@ where
...
-              Err(Err::Error(Error::from_error_kind(
-                input,
-                ErrorKind::TakeWhileMN,
-              )))
-            };
-            res
+    for (i, item) in input.iter_elements().enumerate() {
+      if !cond(item) {
+        if i >= m {
+          if let Ok(index) = input.slice_index(min(i, n)) {
```
Geal, line 250: it looks like the loop runs until the first element that does not match, but it should check that `i <=n` on every iteration. Otherwise it could get very far

Geal (changes requested):

src/bytes/complete.rs
```diff
@@ -242,55 +242,29 @@ where
...
-        if len >= n {
-          match input.slice_index(n) {
-            Ok(index) => Ok(input.take_split(index)),
-            Err(_needed) => Err(Err::Error(Error::from_error_kind(
-              input,
-              ErrorKind::TakeWhileMN,
-            ))),
+      if !cond(item) {
+        if i >= m {
+          if let Ok(index) = input.slice_index(min(i, n)) {
```
Geal, line 253:
```suggestion
          if let Ok(index) = input.slice_index(i) {
```
here we already know that i <=n

Geal: maybe I removed `iter_indices` too early. it would have avoided going through the data twice here. I think we'll merge this and I'll see if I can simplify it

src/traits.rs
```diff
@@ -234,7 +234,7 @@ impl<'a> Input for &'a str {
   type Iter = Chars<'a>;
 
   fn input_len(&self) -> usize {
-    self.len()
+    self.chars().count()
```
Geal, line 237: why this change? This was always supposed to be the bytes length

Geal:
I feel like I'm asking for too many changes, and I want to be mindful of your time here. If you'd prefer, I can take over and finish it

ackxolotl:
Feel free to take over!

PR #1647, opened 2023-03-12
Change fail output type to unit

MariaSolOs:
Hello! First of all, thank you for the amazing project. I've been using `nom` for writing a new parser and it's been a joy <3

While working in my project, I noticed that every time I use the `fail` combinator I have to specify the output type despite the parser always failing, hence such type feels irrelevant (?). In this PR I change the output type to `()` and hence simplify the use of the combinator.

### DIFF
--- a/src/combinator/mod.rs
+++ b/src/combinator/mod.rs
@@ -811,8 +811,8 @@ pub fn success<I, O: Clone, E: ParseError<I>>(val: O) -> impl Fn(I) -> IResult<I
 /// use nom::combinator::fail;
 ///
 /// let s = "string";
-/// assert_eq!(fail::<_, &str, _>(s), Err(Err::Error((s, ErrorKind::Fail))));
+/// assert_eq!(fail(s), Err(Err::Error((s, ErrorKind::Fail))));
 /// ```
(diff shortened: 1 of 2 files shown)


epage:
> While working in my project, I noticed that every time I use the fail combinator I have to specify the output type despite the parser always failing, hence such type feels irrelevant (?). In this PR I change the output type to () and hence simplify the use of the combinator.

I'm curious where you are using `fail` that the output type isn't relevant.  For me, it almost always is.  For example, in the last case of an `alt`, I might use a `fail` to specialize the error message.  The output type needs to be the same type as all of the other cases in my `alt`.

MariaSolOs:
@epage oh you're right, this change would break the usage of `fail` in `alt`. You can dismiss this PR then.

To answer your question, my use cases for `fail` involve parsing something, performing some verification which may involve modifying the input span (I'm using `nom_locate` with a `LocatedSpan` that has some state inside its `extra`), and if such verification fails, I directly use the `fail` parser with `let (span, _) = fail::<_, (), _>(span)?`. I know that this looks odd, but since my verifications might require access to the `span`, I cannot use something like `verify` (since there I only have access to the output type).

PR #1650, opened 2023-03-14
Make Error clonable

xmakro:
Err is clonable, but error::Error was not.

I ran into this while updating the warc crate from nom 5.1.1 to nom 7.1.3 on this callsite:
https://github.com/jedireza/warc/blob/b2b9dfca0d81ed5a91e0016ea89022caf5283ec4/src/record.rs#L182


Geal (approved):

Geal:
thanks!

PR #1655, opened 2023-04-08
Improve wording on nom::Err description

WhyNotHugo:
The previous description assumed that the reader is writing the code that returns `nom::Err`. It was very confusing when reading it as a someone handling this return type.

Hopefully this wording is more neutral and clearer is both contexts.


Geal (approved):

Geal:
thanks!

PR #1663, opened 2023-05-09
Make is_a example more obvious

smheidrich:
In the ["Choosing a combinator" guide](https://github.com/rust-bakery/nom/blob/1309141ae58265ccbbc330b46233076feb6ff98b/doc/choosing_a_combinator.md), the example for `is_a` is to run `is_a("ab")` on `ababc`, returning `Ok("c", "abab")`. The example by itself (without the accompanying explanation) leaves ambiguous whether the reason `is_a` produced that output was because it parsed `ab` twice (wrong - that's what `tag` is for) or because it parsed `a` or `b` 4 times (correct). IMHO it would be better if, at least for simple functions like that, it was immediately obvious from the examples alone what they do, or at least for the examples to dispel any confusion remaining after reading the explanations. In this case that can be accomplished by just using `abbac` as the example string instead, which this MR does.


Geal (approved):

Geal:
giid idea yes, the examples should be more obvious

Issue #1664, opened 2023-05-10
"Choosing a combinator" guide is missing some combinators


smheidrich:
## Prerequisites

Here are a few things you should provide to help me understand the issue:

- Rust version : irrelevant
- nom version : 7.1.2-41-g1309141
- nom compilation features used: irrelevant

## Issue

As others have mentioned (e.g. https://github.com/rust-bakery/nom/issues/1506#issuecomment-1195541465), the ["Choosing a combinator" guide](https://github.com/rust-bakery/nom/blob/1309141ae58265ccbbc330b46233076feb6ff98b/doc/choosing_a_combinator.md) is missing a few combinators such as [`value`](https://docs.rs/nom/latest/nom/combinator/fn.value.html) or [`many0_count`](https://docs.rs/nom/latest/nom/multi/fn.many0_count.html). It would be better if it included all of them, of course.

A list can of all missing ones I could find can be found in my MR to add some of them: #1665

PR #1665, opened 2023-05-10
Add some missing combinators to choosing combinator guide

smheidrich:
Solves part of #1664 

Adds:
- `value`
- `many0_count`
- `many1_count`
- `take_while_m_n`
- `all_consuming`
- `cut`
- `into`
- `rest_len`

Doesn't add:
- `fill` because I don't know a good way to write an example in the format of that table for this one...
- `satisfy` because it doesn't work with the `u8` predicates like `is_alphabetic` etc. which makes it awkward to write an example
- `iterator` (not really a combinator as it doesn't return a parser)
- `recognize_float` and `recognize_float_parts` (don't see the point of those)

Someone who understands more about them could pick these up.


Geal (approved):

Geal:
thanks!

PR #1670, opened 2023-06-14
Update nom_recipes /* C-style comments */

red-swan:
The /* C-style comments */ section wouldn't work on my machine. Compiler error was very dense. I added a `tuple` to the three pieces of the parser and it works as signatured.

### DIFF
--- a/doc/nom_recipes.md
+++ b/doc/nom_recipes.md
@@ -94,11 +94,11 @@ use nom::{
 pub fn pinline_comment<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a str, (), E> {
   value(
     (), // Output is thrown away.
-    (
+    tuple((
       tag("(*"),
       take_until("*)"),
       tag("*)")
-    )
+    ))
   )(i)
 }
 ```


Xiretza:
You're trying to use the examples from `main` for a `7.x` nom. `main` has [deprecated `tuple()`](https://github.com/rust-bakery/nom/pull/1425) and updated the examples accordingly, you should be looking at the 7.x branch instead (or simply the [docs.rs page for your version of nom](https://docs.rs/nom/7.1.3/nom/recipes/index.html#-c-style-comments-)).

epage:
> You're trying to use the examples from main for a 7.x nom. 

I kept running into this with clap (didn't help how long v3 was unreleased) which is one of a couple of reasons which led me to integrating examples directly into docs.rs so they are versioned with the API reference and cross-linkable ([example 1](https://docs.rs/clap/latest/clap/_derive/_cookbook/index.html), [example 2](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html)).

red-swan:
Thanks @Xiretza for the reply. Why does `main` diverge from most up-to-date crate? Is there a reason you're not pushing breaking changes to another branch (e.g. `v8.0.0-release-candidate`)?

Xiretza:
That's how @Geal does it, and in my experience it's a fairly common workflow. But of course the variety of git workflows is endless.

Geal:
it is very common to have the main branch being the most up to date version. I understand that sometimes this can cause confusion, though. In general, the documentation is the canonical source for a specific version, that's why I made sure to have the examples from the recipes available in the documentation, matching the relevant version: https://docs.rs/nom/7.1.3/nom/recipes/index.html#-c-style-comments-

PR #1673, opened 2023-06-29
fix: parse "infinity" literal fully in `float`

boxdot:
`float` first parsed the "inf" literal and then "infinity", therefore even though the latter was parsed correctly, the suffix "inity" was returned as remaining input, which is not correct.

### DIFF
--- a/src/number/complete.rs
+++ b/src/number/complete.rs
@@ -1283,11 +1283,11 @@ where
         .map_err(|_| crate::Err::Error(E::from_error_kind(i, ErrorKind::Float)))
     },
     |i: T| {
-      crate::bytes::complete::tag_no_case::<_, _, E>("inf")(i.clone())
+      crate::bytes::complete::tag_no_case::<_, _, E>("infinity")(i.clone())
         .map_err(|_| crate::Err::Error(E::from_error_kind(i, ErrorKind::Float)))
     },
     |i: T| {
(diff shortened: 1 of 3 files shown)


epage:
Not the maintainer but I'm assuming the intent here is to parse the keywords `inf`, similar to `nan`.  `infinity` isn't generally a keyword.  The reason `inity` is being left is because nom-style parsers leave all unrecognized content for the next parser to identify or error on.

For a language when an invalid keyword is used often enough, it can make sense to special case the error situation into the grammar (match on `infinity` but error).

boxdot:
> infinity isn't generally a keyword

I forgot to mention that nom 6 parsed `infinity` as float infinity without remaining input. The current code also contains a case and tests for parsing infinity completely. So, I assume it is a regression in nom 7.

My motivation is actually this upgrade: https://github.com/Garvys/rustfst/pull/240

epage:
Huh, didn't realize that `FromStr` accepts `infinity`.  `FromStr` and `lexical_core` were the nom v6 implementations of `float`.  In nom v7, an attempt was made to switch to `minimal_lexical` (baf5e01) but that had problems (f8633f9).  Whats confusing is I'm not seeing how the `minimal_lexical` version was meant to handle `nan`, `inf`, and `infinity`.

Geal:
right, the order was wrong here, nice catch

Geal:
thank you!

PR #1675, opened 2023-07-14
Add `--generate-link-to-definition` option when building on docs.rs

GuillaumeGomez:
This option generates links in source code pages, allowing to jump to definition and to jump back to doc. It makes browsing the source code pages much nicer. You can see it in action [here](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_middle/lib.rs.html#90).


Geal (approved):

Geal:
thanks!

PR #1683, opened 2023-08-13
[nit] Move cnt definition to the branch it's used in.

ttsugriy:
It's not a big deal and compiler should be able to do this automatically, but this also reduces the scope of the definition.


Geal (approved):

Geal:
thanks!

PR #1684, opened 2023-08-14
[nit] Replace hard-coded bit count with constant.

ttsugriy:
This makes it clear what `8` stands for.


Geal:
I'm not sure it makes things easier to read. If you work with bits, you know why 8 is there

PR #1695, opened 2023-09-29
fix: Recipes had drifted from the implementation and weren't compiling

krtab:
This

1. Adds the recipes to the doctests so that they don't drift anymore
2. Fixes them


doc/nom_recipes.md
```diff
@@ -99,7 +101,7 @@ pub fn pinline_comment<'a, E: ParseError<&'a str>>(i: &'a str) -> IResult<&'a st
       take_until("*)"),
       tag("*)")
     )
-  )(i)
+  ).parse(i)
```
Fyko, line 104: Why use `parse` on all these instead of parser functions?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3479d0fd619dd9886839429c6f9cc9a1

krtab: I think it is required since #1631 has been merged.

Geal: nom is moving towards using traits instead of function as the main parser interface. You can still use functions though if you prefer

src/lib.rs
```diff
@@ -466,3 +466,7 @@ pub mod number;
 #[cfg(feature = "docsrs")]
 #[cfg_attr(feature = "docsrs", cfg_attr(feature = "docsrs", doc = include_str!("../doc/nom_recipes.md")))]
 pub mod recipes {}
+
+#[cfg(doctest)]
```
Geal, line 470: why not add doctest to the definition above?

Geal (approved):

Geal:
it took a bit of fiddling, but now all tests are passing in CI. Thankyou for your help!

krtab:
Perfect, I hadn't added the `cfg` because I wasn't sure how it would interract with the feature gate, great you figured it out.

PR #1702, opened 2023-10-21
Fix clippy warnings

Geal:


Geal:
and done! thanks @thatmarkenglishguy !

PR #1704, opened 2023-10-28
Experimental state management

Geal:


epage:
I'm curious, what problem is this looking to solve?

PR #1709, opened 2023-11-17
Remove dependency on minimal-lexical

nabilwadih:
This dependency is not actually used anywhere in the project, however it was being pulled in as a dep in the Cargo.toml. This change is simply to remove it from the Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,13 +30,9 @@ include = [
 
 [features]
 alloc = []
-std = ["alloc", "memchr/std", "minimal-lexical/std"]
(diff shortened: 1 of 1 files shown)


marshallpierce:
Beyond just trimming unused dependencies being a nice thing for simplicity, another motivation here is to avoid having to do security review on `minimal-lexical` and its liberal use of unsafe just to be able to use `nom` in an environment that does reviews on all third party code.

Geal (approved):

Geal:
thanks!

PR #1714, opened 2023-12-04
Replace uses of `unsafe` with safe alternatives

nabilwadih:
Some of the usages of `unsafe` in the codebase can now be replaced with safe functions from the rust standard library. This PR updates those to remove the unsafe blocks

Verified that all of the tests are still passing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -379,7 +379,6 @@
 #![cfg_attr(feature = "docsrs", feature(doc_cfg))]
 #![allow(clippy::doc_markdown)]
 #![deny(missing_docs)]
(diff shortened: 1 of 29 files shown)


Geal:
these unsafe calls are there to avoid paying anohter bounds check (and for `&str` a `is_char_boundary` check) while code outside is already supposed to guarantee it

src/lib.rs
```diff
@@ -435,6 +435,7 @@ pub use self::bits::*;
 pub use self::internal::*;
 pub use self::traits::*;
 
+
```
Geal, line 438:
```suggestion
```

Geal (approved):
running the benchmarks again, this has no noticeable impact, so this will be ok

PR #1722, opened 2024-01-22
feat: Add separated_list_m_n

jtojnar:
This is useful to naïvely represent some grammar productions
(e.g. `IPv6address` from [RFC 3986]):

```abnf
IPv6address /= [ *2( h16 ":" ) h16 ] "::" 2( h16 ":" ) ls32
```

turns to

```diff
 tuple((
-    opt(tuple((many_m_n(0, 2, tuple((h16, char(':')))), h16))),
+    separated_list_m_n(0, 3, char(':'), h16),
     tag("::"),
     many_m_n(2, 2, tuple((h16, char(':')))),
     ls32,
 ))
```

[RFC 3986]: https://www.rfc-editor.org/rfc/rfc3986#appendix-A


cenodis:
Now that https://github.com/rust-bakery/nom/pull/1608 is merged I think it would be better to implement the proper range syntax instead of adding yet another variant for a specific format. Geal has already mentioned `seperated_list` as a candidate in https://github.com/rust-bakery/nom/issues/1613, so this variant would be made redundant soon anyway.

Issue #1691, opened 2023-08-27
separated_list1 by multipspace0 sep failed


yejunjin:
Hello, and thank you for submitting an issue to nom!

First, please note that, for family reasons, I have limited time to work on
nom, so following the advice here will make sure I will quickly understand
your problem and answer as soon as possible.
Second, if I don't get to work on your issue quickly, that does not mean I
don't consider it important or useful. Major version releases happen once
a year, and a lot of fixes are done for the occasion, once I have had time
to think of the right solution. So I will get back to you :)

## Prerequisites

Here are a few things you should provide to help me understand the issue:

- Rust version : rustc 1.71.1 (eb26296b5 2023-08-03)
- nom version :7.1.3
- nom compilation features used: default

## Test case

Please provide a short, complete (with crate import, etc) test case for
the issue, showing clearly the expected and obtained results.

Example test case:

```
    #[test]
    fn test_separated_list1() {
        fn parser(input: &str) -> IResult<&str, Vec<&str>> {
            separated_list1(multispace0, tag("abc"))(input)
        }
        assert_eq!(parser("abcabc"), Ok(("", vec!["abc", "abc"])));
       // thread 'parser::tests::test_cypher' panicked at 'assertion failed: `(left == right)`
       // left: `Err(Error(Error { input: "abc", code: SeparatedList }))`,
       // right: `Ok(("", ["abc", "abc"]))`
    }
```

epage:
While its not in the docs, in the [code is this comment](https://github.com/rust-bakery/nom/blob/7.x/src/multi/mod.rs#L330)

> infinite loop check: the parser must always consume

Meaning, the separator parser.  I assume this is protecting against writing code where both the separator and the core type are allowed to be empty, you'd get an infinite loop, so nom seems to say the core type *may* be empty but the separator must not be.

SaltyKitkat:
Maybe related with #1573

gdennie:
The `sep` parser in `seperated_list` should not be optional because then its target is not a `separated list`, really. The documentation should mention this constraint but doesn't. I've added a pull request for that. 

Instead of modifying `separated_list`, perhaps we should add another combinator instead, `optionally_separated_list`.

Taywee:
[...]

PR #1723, opened 2024-01-27
stipulate `sep` parser must consume

gdennie:
separated_list1 by multipspace0 sep failed #1691 

the `sep` parser in the `separated_list` parser combinators cannot be "any" parser but must be a consuming parser, contrary to documentation. Ideally, we should have a trait to assert this constraint as well but better clarity in the documentation helps.

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -337,7 +337,7 @@ where
 /// [`cut`][crate::combinator::cut].
 ///
 /// # Arguments
-/// * `sep` Parses the separator between list elements.
+/// * `sep` Parses the separator between list elements. Must be consuming. 
 /// * `f` Parses the elements of the list.
 ///
 /// ```rust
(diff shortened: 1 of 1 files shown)


src/multi/mod.rs
```diff
@@ -337,7 +337,7 @@ where
 /// [`cut`][crate::combinator::cut].
 ///
 /// # Arguments
-/// * `sep` Parses the separator between list elements.
+/// * `sep` Parses the separator between list elements. Must be consuming.
```
epage, line 340: `many0` has a similar note that is separate from the Arguments list

> Note: if the parser passed in accepts empty inputs (like alpha0 or digit0), many0 will return an error, to prevent going into an infinite loop

I'm assuming it'd be good to be consistent.

gdennie: Whether a parser consumes some input is clearly an important characteristic. As such, it would be good if that characteristic could be put into the type system.

Geal (approved):

Geal:
thank you, it's a good idea to be explicit about it

Issue #1717, opened 2023-12-13
Add u48 number parser


hackaugusto:
Numbers with 6bytes are used in some formats, since `nom` supports some non-standard sizes, e.g. [3byte with u24](https://docs.rs/nom/latest/nom/number/complete/fn.i24.html). It would be nice to have it for u48 too.

I have no use case for the other sizes, but maybe it makes sense to include other value ranges for completeness?

Geal:
sorry, I don't think it is a good idea to add more number parsers at this point. Adding parsers like u24 was a mistake, because they are rarely used, and people mainly use the 8, 16, 32 and 64 bit parsers. So in that case I'd recommend you make your own in your project, since nom is designed to let you write your basic parsers and build upon them

PR #1729, opened 2024-02-11
Add 6 bytes integers

JonathanPlasse:
- Implement #1717


Geal:
Hi! thank you for looking into this, but as per https://github.com/rust-bakery/nom/issues/1717#issuecomment-2068131625 I don't think it is a good idea to add another number parser at this point

JonathanPlasse:
Hi! No problem, have a good day.

PR #1741, opened 2024-04-22
fix cargo-codspeed in with workspaces in CI

art049:
This fixes the CodSpeed workflow. We released [a new version yesterday](https://github.com/CodSpeedHQ/codspeed-rust/releases/tag/v2.5.0) and this created a regression. This will be fixed very soon, but meanwhile, these changes fix the behavior.

Sorry for the inconvenience. 

Edit: This has been fixed in [v2.5.1](https://github.com/CodSpeedHQ/codspeed-rust/releases/tag/v2.5.1) so merging this is optional and the workflow should be passing now.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -23,12 +23,10 @@ jobs:
           bins: cargo-codspeed
 
       - name: Build the benchmark target(s)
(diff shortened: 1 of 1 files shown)


Geal (approved):

Geal:
great, thank you for your help

PR #1743, opened 2024-04-24
Change context to pass output mode information to the child parser.

kstrohbeck:
Currently, the `nom::error::context` combinator calls `parse` on its argument, which forces it to run in streaming mode (as well as making it emit output and errors.) By directly implementing `Parser`, we can pass the `OutputMode` information down to the underlying parser.


Geal (approved):
great addition, thanks!

PR #1747, opened 2024-05-05
remove the InputLength trait

Geal:


src/character/complete.rs
```diff
@@ -208,7 +208,7 @@ where
 /// ```
 pub fn line_ending<T, E: ParseError<T>>(input: T) -> IResult<T, T, E>
 where
-  T: Input + InputLength,
+  T: Input + Input,
```
epage, line 211: redundant bound

Geal: nice catch, thanks

PR #1760, opened 2024-05-09
Bump MSRV to 1.57.0 to more correctly match the codebase.

martinfrances107:
As the code has been updated new features that break the currently stated MSRV have been added.

I just want  the codebase to reflect reality.

This issue can be seen by running cargo clippy 

```
warning: current MSRV (Minimum Supported Rust Version) is `1.56.0` but this item is stable since `1.57.0`
   --> src/traits.rs:649:10
    |
649 |     self.as_slice()
    |          ^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
    = note: `#[warn(clippy::incompatible_msrv)]` on by default
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,8 @@ edition = "2021"
 autoexamples = false
 
 # also update in README.md (badge and "Rust version requirements" section)
-rust-version = "1.56"
+# 1.57.0 - Now using '.as_slice()'
+rust-version = "1.57.0"
 
 include = [
   "CHANGELOG.md",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,7 @@ edition = "2021"
 autoexamples = false
 
 # also update in README.md (badge and "Rust version requirements" section)
-# 1.57.0 - Now using '.as_slice()'
-rust-version = "1.57.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -14,7 +14,8 @@ edition = "2021"
 autoexamples = false
 
 # also update in README.md (badge and "Rust version requirements" section)
-rust-version = "1.56"
+# 1.57.0 - Now using '.as_slice()'
+rust-version = "1.57.0"
```
Geal, line 18:
```suggestion
rust-version = "1.65.0"
```

The new MSRV is actually 1.65, because nom 8 will use GATs

Geal (approved):

martinfrances107:
looking at the 14% drop in preformance ( A change  to 1 µs	to 1.2 µs )  well that looks completely to be expect from  from repeatibility noise.

The +10% imrpovement (  from 287.2 ns  to 259.4 ns ) is a welcome minor improvement 

In short,  I think being blocked on this is misguided.

Geal:
chill a bit. Benchmarks can be flaky, that's not an issue, and this change definitely cannot affect performance

PR #1774, opened 2024-08-09
feat: Update README for newer Netflow library.

mikemiles-dev:


Geal (approved):
that looks nice! Any interesting software using it?

PR #1775, opened 2024-08-12
Implement new parser `left_assoc`.

rodrigorc:
This PR adds a new combinator `left_assoc` to parse left-associative operators.

Parsing operator with different precedence in `nom` has usually been a bit of a challenge. There is this other PR#1362, but it seems a bit stalled.

The classical alternative in LALR grammars is to use a recursive rule such as:
```
A := A op B | B
```
Alas, that can't be parsed easily with `nom` because it will recurse infinitely.

The solution is to parse it with this other equivalent rule:
```
A := B (op B)*
```
while calling a callback to give the user the chance to build the AST, or whatever.

This could be implemented as a call to `B` and then `fold_many0` with the `pair(op, B)`. Instead, I chose copying the `separated_list1` code (that `OM` thing is tricky).

As an alternative public API, the type of the left expression and that of the right expression could be different. Then we would need two separated parsers, `left` and `right` returning types `L` and `R`; and the `builder` would be of type `FnMut(L, OP, R) -> L`.  Maybe another combinator, `left_assoc_2()` could be added for that?

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -1868,3 +1868,138 @@ where
     Ok((input, acc))
   }
 }
+
+/// Applies a parser multiple times separated by another parser.
+///
+/// It is similar to [`separated_list1`][crate::multi::separated_list1] but instead of collecting
+/// into a vector, you have a callback to build the output.
+///
+/// In a LALR grammar a left recursive operator is usually built with a rule syntax such as:
+///  * A := A op B | B
+///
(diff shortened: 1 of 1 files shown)


Geal:
Hi! This looks great! FYI I am moving this to a [separate crate](https://github.com/rust-bakery/nom/pull/1792) that is more focused on language parsing, while keeping the core nom library more generalist. This will allow that crate to have a different lifecycle from the core library

PR #1784, opened 2024-11-13
separated_list doc and test update

SaltyKitkat:
- related issue: https://github.com/rust-bakery/nom/issues/1691#issuecomment-2094891869
- related pr: https://github.com/rust-bakery/nom/pull/1756

### DIFF
--- a/src/multi/mod.rs
+++ b/src/multi/mod.rs
@@ -335,7 +335,7 @@ where
 /// [`cut`][crate::combinator::cut].
 ///
 /// # Arguments
-/// * `sep` Parses the separator between list elements. Must be consuming.
+/// * `sep` Parses the separator between list elements.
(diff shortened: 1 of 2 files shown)


Geal (approved):
this looks good, thanks

gdennie:
I haven't studied the code so perhaps it resolves other concerns or bugs but 11.95% seems like a big hit for functionality that seems to be available elsewise. Nonetheless, great. 😉

In a fantasy world we could have a marking trait, `NonConsuming`, that attaches to parsers derived from non-consuming subparsers that would be given the trait. That may mean an extended algebra of trait composition. Of course, for clarity we require restating the trait but it would be compiler checked. Then the trait would then make this constraint on the combinators compiler visible.

Geal:
This is a flaky benchmark so do not read too much into it

Issue #1785, opened 2024-11-13
Add nom-supreme as a (feature-gated) first-class citizen


Ben-PH:
I would like to feature-gate a re-export of `nom-supreme`.

- I consider `ErrorTree` as "the basic very-verbose" error in a minimilist/informative/very-verbose. in-crate `Error` and `VerboseError` make up the other two "basics"
- It signals to users that `nom` isn't _meant_ to be a fully fleshed out utility. Its intent is to make the process of assembling different utilities together.
- It creates a better UX, in that users have the option of a heavy-weight, feature-rich experience.

I searched `supreme` in the issue searcher (not filtering by pr or issue), and nothing came up in the titles. Either this is a new thing (great), but if not, the discussion that follows will document the reasoning behind what's done for future contributors.

PR #1786, opened 2024-11-13
re-export nom-supreme behind feature flag

Ben-PH:
Related issue: #1785 

This PR is opinionated in the sense that the feature-gated re-export of `nom-supreme` (with the namespace `nom::supreme`) is a default feature.

I'm happy for this to be merged as-is, with a follow-up issue/PR pair for documentation (readme, book, examples, etc), or for this to be made into a draft for that followup to be rolled up into this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,10 +29,15 @@ include = [
 ]
 
 [features]
-alloc = []
+default = ["std", "supreme"]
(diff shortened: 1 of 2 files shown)


Geal:
closing this for now, I don't really want to reexport too much from nom. There may be more opportunities for it in nom-language

PR #1787, opened 2024-11-13
Move nom into nom_core workspace and re-export

Ben-PH:
This allows for the re-export of not only nom-core, thus making this a no-change to the api, it also allows for reexport of other community extensions, such as nom-supreme

- #1785 

I hope to follow up on this PR by

- re-exporting nom-supreme behind a feature flag
- re-exporting nom-locate behind a feature flag
- moving some of the less "core" elements of nom into a re-exported workspace, putting them behind (default) feature flags.

The goal is to be able to integrate the nom ecosystem so that, instead of having to run

```
cargo add nom
cargo add nom-supreme
cargo add nom-locate
```

you can:

```
cargo add nom -F supreme locate
```

#### extra bit of fun:

to keep with the theme of this crate, we could also start adding "menu items". "extra toppings" (e.g. more verbose debug tracing). "amuse bouche" (e.g. exercise examples: have test cases and unimplemented parsers and implement them locally).


Geal:
closing this as I went in a different direction with nom-language. Thanks for the suggestion though!

PR #1795, opened 2025-01-11
Changed Copy constraint to Clone

fmsbeekmans:
This PR changes the constraint for `Copy` to `Clone` to enable using it with recursive data structures as their output.

I found myself needing to have both `Clone` to work with the `Input` trait and `Copy` to work with the `nom-language` module. This made it impossible to use a recursive struct as the output `Q` for `precedence` as far as my limited Rust understanding goes.

### DIFF
--- a/nom-language/src/precedence/mod.rs
+++ b/nom-language/src/precedence/mod.rs
@@ -13,7 +13,7 @@ pub struct Unary<V, Q: Ord + Clone> {
 }
 
 /// A binary operator.
-pub struct Binary<V, Q: Ord + Copy + Clone> {
+pub struct Binary<V, Q: Ord + Clone> {
     value: V,
     precedence: Q,
     assoc: Assoc,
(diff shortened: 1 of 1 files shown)


marcdejonge:
The precedence parameter is copied a lot during parsing, so to keep performance high this crate requires it to be something that can be copied (basically it should fit on the stack). Having something that is not copyable, but cloneable, potentially creates a lot of overhead and will slow down your parser a lot. So I think the `Copy` trait here is on purpose.

The whole goal of the precedence is to determine in what order things should be parsed, so I also don't understand why you would use anything other than simple numbers or an enum. Could you provide an example why you would use anything else?

ps. In the current PR you also changed the whitespace of this file. This project uses 2 spaces of indenting. Changing it to 4 spaces for these few files will never be accepted. It also makes the review a lot harder. If you want to continue discussing this change, you should fix this.

fmsbeekmans:
Hi, thanks for the response. This was caused by a misunderstanding on my side. I thought the precedence parameter needed to be the same type as the output type.

PR #1807, opened 2025-01-29
Fix documented examples to not use the deprecated is_alphabetic function

marcdejonge:
It's a trivial documentation change, but the docs should give the good example


Geal (approved):

Geal:
thanks!

PR #1810, opened 2025-01-30
Change the standard Input impl to support any slice

marcdejonge:
The main change is that the Item is a reference instead of a copy.

It also adds an `IntoInput` type that makes it easier to deal with inputs for the tag and take functions.

Possible solution to solve https://github.com/rust-bakery/nom/issues/1809 and https://github.com/rust-bakery/nom/issues/1799


marcdejonge:
Probably also something for 9.0 if you even want to do this

PR #1811, opened 2025-01-30
Fix for issue 1808

marcdejonge:
It turned out that the `split_at_position_complete` for str returned the beginning of the string with length 0, instead of the end. This broke the offset calculations.

Fixes https://github.com/rust-bakery/nom/issues/1808

### DIFF
--- a/tests/issues.rs
+++ b/tests/issues.rs
@@ -269,3 +269,15 @@ fn issue_1586_parser_iterator_impl() {
 
   assert_eq!(parse_input("123\n456").collect::<Vec<_>>(), vec![123, 456]);
 }
+
+#[test]
+fn issue_1808_complete_string_parser_returns_wrong_slice() {
+  use nom::character::complete::multispace0;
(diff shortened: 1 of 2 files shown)


tests/issues.rs
```diff
@@ -269,3 +269,15 @@ fn issue_1586_parser_iterator_impl() {
 
   assert_eq!(parse_input("123\n456").collect::<Vec<_>>(), vec![123, 456]);
 }
+
+#[test]
+fn issue_1809_complete_string_parser_returns_wrong_slice() {
```
ThatOneShortGuy: wrong issue number, should be 1808

marcdejonge: Stupid mistake, also in the commit description. I've fixed it, thanks!

mcheshkov:
@Geal could you please make a crates.io release with this fix?

relrod:
@Geal Hi, also wondering about a release with this fix.

PR #1815, opened 2025-02-04
doc: Adding lp_parser_rs to list of known projects using nom

dandxy89:


Geal (approved):

Geal:
interesting! I did not know about that use case

dandxy89:
It's a fairly niche use-case in all fairness to you.

Extraction of data from them has been historically tedious and challenging - particularly when wanting to compare millions of coefficients and constraints between model runs. 

Thanks 'nom'!

PR #1816, opened 2025-02-05
Add LICENSE file to nom-language

miguelfrde:
Hi, I noticed that the `nom-language` package doesn't contain a `LICENSE` file. Including it would make it easier to use in other codebases.

This change creates a symlink to the root LICENSE file.

This ensures that the `LICENSE` text is contained in the `cargo package` output. With this change after `cargo package` the `nom-language` crate:

```
$ cd target/package/nom-language-0.1.0
$ ls
Cargo.lock      Cargo.toml.orig src
Cargo.toml      LICENSE         target
```

This matches what multiple other repositories do (for example, [actix-web]( https://github.com/actix/actix-web/tree/master/actix-web), [tracing-core](https://github.com/tokio-rs/tracing/tree/master/tracing-core)).


Geal (approved):

Geal:
good idea, thanks

miguelfrde:
Would it be possible to place this file in the release of 0.1.0 or create a release for 0.1.1 that contains it? Thanks in advance!

PR #1818, opened 2025-02-06
Update GLSL URL.

hadronized:
I changed nickname and moved away from GitHub.


Geal:
alright! I'm glad to see this is till going strong :)

PR #1825, opened 2025-02-12
Create Parser implementation for rest

gadunga:
This PR changes the `rest` combinator so that it now returns an `impl Parser`.  I've updated the old tests for `rest` as well as any test/documentation that made  use of the combinator.


shahn:
This would be a breaking change, either it might be for nom v9 or a new function should be added

Issue #1809, opened 2025-01-29
Merge InputLength into Input can introduce extra trait requirement


tisonkun:
See https://github.com/andylokandy/nom-rule/pull/8

Previously, we have:

https://github.com/rust-bakery/nom/blob/869f8972a4383b13cf89574fda28cb7dbfd56517/src/multi/mod.rs#L52-L58

and

```rust
pub trait InputLength {
  fn input_len(&self) -> usize;
}

impl<'a, T> InputLength for &'a [T] {
  #[inline]
  fn input_len(&self) -> usize {
    self.len()
  }
}
```

So `InputLength` is implemented for `&[proc_macro2::TokenTree]`.

But now we have:

https://github.com/rust-bakery/nom/blob/2cec1b3e4c9ccac62c902d60c00de6d1549ccbe1/src/multi/mod.rs#L61-L67

So `&[proc_macro2::TokenTree]` can no longer directly adapted in `many0`.

Perhaps we can wrap a newtype to work it around. But I'd first report this use case and see if there is better/idiomatic way.

andylokandy:
Thank you @tisonkun for investigating the problem. Seems we are not able to add the impls by ourself except we add a new type which is not so ergonomic 🤔.

marcdejonge:
I think there should be an implementation for Input for all slices of objects. The problem is the current implementations are in the way, because they use a copied version of the `Input::Item`. But if the `Input::Item` would return a reference to the thing in the array, nothing would really break and you would have a nice generic implementation.

The main downside is that it's another breaking API change that could have an impact on code that people are using.

marcdejonge:
https://github.com/rust-bakery/nom/pull/1810/files#diff-172747fc43e38d33de743b2d5f564ad6440570670af7aa912235399448a55c7cR192

Something like this actually passes all the tests and should work for you. The main issue is that is changes the external API and it will brake stuff for some people.

asibahi:
I tried to change the default implementation of the `&[u8]` implementation to accept &[T: Clone]` instead. The fork is here: https://github.com/asibahi/nom/tree/impl-clone

I *think* using `clone()` and `Cloned` on `Copy` types just copies them. `u8` at least should do that.

All the tests pass without changing any of them. It would also make it easier for using parsers on arbitrary Token trees.

PR #1831, opened 2025-02-25
impl Input for &[T] where T : Clone

asibahi:
Hello.

This is not related to a specific issue, tho I thought to do this as an alternate implementation of #1810 , and I mentioned it in #1809 .

This generalizes `Input` implementation over any type that implements `Clone`. Which includes `u8` as well as custom lexed `Token` types common in parsers.

My understanding is that any type that implements `Copy`, when `clone()` is called on it, just .. copies. No additional performance overhead. I *would* make this as a separate implementation from `&[u8]` but you'd run into specialization.

All the tests pass with no changes to them whatsoever.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,10 +1,9 @@
 //! Traits input types have to implement to work with nom combinators
-use core::iter::Enumerate;
+use core::iter::{Cloned, Enumerate};
 use core::str::CharIndices;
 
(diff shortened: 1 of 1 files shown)


autumnull:
This would be really useful, I am currently writing a lexer and parser with nom and really don't want to write the boilerplate for this :sob:

PR #1832, opened 2025-03-03
A few doc improvments to `/docs` and others

### LINKED ISSUE #1801: Docs for Parser::parse and Parser::parse_complete do not note any difference
```
A parser takes in input type, and returns a Result containing either the remaining input and the output value, or an error
```
is the docstring for both. Is this the high-level switch between streaming and complete parsers, or should this be done some other way? Should I use parse_complete everywhere if I want a complete parser, or is it enough to use it once? I am not really sure how that fits together.

### LINKED ISSUE #1823: `error_management.md` mentions non-existent `VerboseError`
Hi, I'm trying to upgrade to nom 8 and was trying to figure out how to replace some code that uses `VerboseError`. While looking around, I noticed that the `doc/error_management.md` document still discusses `VerboseError` which doesn't exist anymore, so it looks like this documentation should be updated.

asibahi:
Issues: #1823 , #1801 in addition to some self-prompted fixes.

fixes #1823 
fixes #1801


doc/choosing_a_combinator.md
```diff
@@ -38,7 +38,7 @@ Those are used to recognize the lowest level elements of your grammar, like, "he
...
 | [pair](https://docs.rs/nom/latest/nom/sequence/fn.pair.html) | `pair(tag("ab"), tag("XY"))` | `"abXYZ"` | `Ok(("Z", ("ab", "XY")))` |Gets an object from the first parser, then gets another object from the second parser.|
 | [separated_pair](https://docs.rs/nom/latest/nom/sequence/fn.separated_pair.html) | `separated_pair(tag("hello"), char(','), tag("world"))` | `"hello,world!"` | `Ok(("!", ("hello", "world")))` |Gets an object from the first parser, then matches an object from the sep_parser and discards it, then gets another object from the second parser.|
-| [tuple](https://docs.rs/nom/latest/nom/sequence/fn.tuple.html) | `tuple((tag("ab"), tag("XY"), take(1)))` | `"abXYZ!"` | `Ok(("!", ("ab", "XY", "Z")))` | Chains parsers and assemble the sub results in a tuple. You can use as many child parsers as you can put elements in a tuple|
+| tuples | `((tag("ab"), tag("XY"), take(1)))` | `"abXYZ!"` | `Ok(("!", ("ab", "XY", "Z")))` | `Parser` is implemented for tuples: it chains parsers and assemble the sub results in a tuple. You can use as many child parsers as you can put elements in a tuple|
```
Geal, line 41: why remove the doc link here?

asibahi: the thought was that the `tuple` combinator (which is linked to) is deprecated and the examples would just use tuples immediately

doc/error_management.md
```diff
@@ -115,9 +115,9 @@ println!(
 );
 ```
 
-### getting more information: nom::error::VerboseError
+### getting more information: nom-language::error::VerboseError
 
-The  `VerboseError<I>` type accumulates more information about the chain of
+The  `VerboseError<I>` type (available from the companion `nom-language` crate) accumulates more information about the chain of
```
ianrrees, line 120: Thanks for making this PR, I've come here after trying to use the VerboseError from nom :)

Just a minor suggestion: that the struct comes from a different crate is an important piece of information, I'd not put it in a parenthetical.  Maybe something like this instead:

The  companion `nom-language` crate's `VerboseError<I>` type accumulates

PR #1842, opened 2025-03-21
added value implementation that uses new architecture instead of dropping

asibahi:
Hello,

No related issue. Just something.I noticed while parosing the code.

This updates the `value` combinator to use the `Check` idiom instead of running its inner parser and dropping the result.


asibahi:
Noticed this is a duplicate of #1829. apologies for the noise.

Issue #1844, opened 2025-03-24
nom 8.0.0 take/map combinators are now FnMut but dbg_dmp only accepts Fn


sebosp:
After upgrading from nom 7.1.1 to 8.0.0 on a crate it started to fail.
The CHANGELOG.md points to `map` combinator taking now FnMut.
The same appears to be the case for `take`, if I  want to use `dbg_dmp` it seems the signature needs to be adjusted as well.

```
error[E0277]: expected a `Fn(&[u8])` closure, found `impl FnMut(&[u8]) -> Result<(&[u8], &[u8]), nom::Err<_>>`
  --> src/parser/mpq_user_data.rs:75:40
   |
75 |         let (input, content) = dbg_dmp(take(user_data_header_size as usize), "content")(input)?;
   |                                ------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an `Fn(&[u8])` closure, found `impl FnMut(&[u8]) -> Result<(&[u8], &[u8]), nom::Err<_>>`
   |                                |                                                                                                                                                                                                                                                        |                                required by a bound introduced by this call
   |
   = help: the trait `Fn(&[u8])` is not implemented for `impl FnMut(&[u8]) -> Result<(&[u8], &[u8]), nom::Err<_>>`
   = note: `impl FnMut(&[u8]) -> Result<(&[u8], &[u8]), nom::Err<_>>` implements `FnMut`, but it must implement `Fn`, which is more general
note: required by a bound in `dbg_dmp`                                                                                                                                                                                                                                                      --> /home/seb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nom-8.0.0/src/error.rs:492:1    
```

PR #1845, opened 2025-03-24
Adjust FnMut param in dbg_dmp since take/map are now FnMut

sebosp:
- Addresses #1844 
- test case added

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -490,11 +490,11 @@ macro_rules! error_node_position(
 #[cfg(feature = "std")]
 #[cfg_attr(feature = "docsrs", doc(cfg(feature = "std")))]
 pub fn dbg_dmp<'a, F, O, E: std::fmt::Debug>(
-  f: F,
+  mut f: F,
   context: &'static str,
-) -> impl Fn(&'a [u8]) -> IResult<&'a [u8], O, E>
+) -> impl FnMut(&'a [u8]) -> IResult<&'a [u8], O, E>
 where
-  F: Fn(&'a [u8]) -> IResult<&'a [u8], O, E>,
+  F: FnMut(&'a [u8]) -> IResult<&'a [u8], O, E>,
 {
   use crate::HexDisplay;
   move |i: &'a [u8]| match f(i) {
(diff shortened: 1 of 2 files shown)


asibahi:
I have been mucking around with this. I think a better implementation would be to implement it as a proper `impl Parser` a la `nom8`. Also made it generic over `HexDisplay` rather than be wedded to `&[u8]`.

```rust
#[cfg(feature = "std")]
#[cfg_attr(feature = "docsrs", doc(cfg(feature = "std")))]
pub fn dbg_dmp<'a, P, I, O, E>(parser: P, context: &'static str) -> DbgDmp<P>
where
  E: std::fmt::Debug,
  P: Parser<I, Output = O, Error = E>,
{
  DbgDmp { parser, context }
}

/// dbg_dmp implementation
#[cfg(feature = "std")]
pub struct DbgDmp<P> {
  parser: P,
  context: &'static str,
}

#[cfg(feature = "std")]
impl<F, I> Parser<I> for DbgDmp<F>
where
  I: HexDisplay + Clone,
  F: Parser<I>,
  <F as Parser<I>>::Error: std::fmt::Debug,
{
  type Output = <F as Parser<I>>::Output;
  type Error = <F as Parser<I>>::Error;

  fn process<OM: OutputMode>(&mut self, input: I) -> PResult<OM, I, Self::Output, Self::Error> {
    match self
      .parser
      .process::<OutputM<OM::Output, Emit, OM::Incomplete>>(input.clone())
    {
      Err(e) => {
        eprintln!("{} Error: {:?} at:\n{}", self.context, e, input.to_hex(8));
        match e {
          Err::Error(e) => Err(Err::Error(OM::Error::bind(|| e))),
          Err::Failure(e) => Err(Err::Failure(e)),
          Err::Incomplete(e) => Err(Err::Incomplete(e)),
        }
      }
      Ok(v) => Ok(v),
    }
  }
}
```

(Edit: a slightly more concise code).

It does look overly complex and I do not particularly like it. I adjusted the doctest to accommodate it and it passes.

```rust
use nom::{Parser, IResult, error::dbg_dmp, bytes::tag};

fn f(i: &[u8]) -> IResult<&[u8], &[u8]> {
  dbg_dmp(tag("abcd"), "tag").parse_complete(i)
}

let a = &b"efghijkl"[..];

// Will print the following message:
// tag Error: Error(Error { input: [101, 102, 103, 104, 105, 106, 107, 108], code: Tag }) at:
// 00000000	65 66 67 68 69 6a 6b 6c 	efghijkl
f(a);
```

PR #1874, opened 2026-01-05
Added a combinator similar to separated_list0, but that allows for a trailing separator. (doc)tests are mostly complete.

SleighPenguin22:
The multi module provides a `separated_list\[01\]()` combinator, but the combinator does not consume a trailing separator, defining my own in my project using a macro is doable, but looks weird when compared to the provided function API.

I'm proposing a combinator whose interface is identical to `separated_list[01](sep, parser)`, but it accepts an optional trailing separator and discards it.


SleighPenguin22:
Now that I think of it, perhaps the `SeparatedList[01]` Structs could instead have a boolean field that specifies if a trailing separator should be consumed at runtime?
the constructor then constructs the corresponding struct with the field set to `true` or `false` depending on the called constructor (i.e. `separated_list0(...)` sets it `false` while `separated_list_trailing0(...)` sets it `true`).

PR #1888, opened 2026-07-31
Update README.md

Vellumic:
This provides the following:
- Updated example code making it more streamlined by using version 8 API
- Updated project examples mostly to the active/used in production repositories
- Updated parsers list
- Some other cosmetic improvements

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,7 @@ error prone plumbing.
 
 ```rust
 use nom::{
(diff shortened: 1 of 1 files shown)


Vellumic:
I also suggest to update the list in #14 and lock its discussion to prevent new projects being add in the comments rather than by a PR

